PR #5, opened 2020-07-22
Fix version number

kxu1026:


version.properties
```diff
@@ -1,4 +1,3 @@
 #Version of the produced binaries. This file is intended to be checked-in.
 #It will be automatically bumped by release automation.
-version=1.0.1
-previousVersion=1.0.0
```
wmoustafa, line 4: How is previous version and version the same?

PR #6, opened 2020-07-24
Change group name to com.linkedin.coral

kxu1026:
Right now, the maven pom's group id is 'coral'. It will be better to use 'com.linkedin.coral'

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,6 +20,7 @@ configurations {
 }
 
 allprojects {
+  group = "com.linkedin.coral"
   repositories {
     mavenCentral()
     maven {


settings.gradle
```diff
@@ -1,5 +1,7 @@
 def modules = ['coral-hive', 'coral-pig', 'coral-presto','coral-spark','coral-schema']
 
+rootProject.name = "com.linkedin.coral"
```
shardulm94, line 3: Don't think this is the right way to set group id. See 
https://github.com/linkedin/transport/blob/6fc9a9274c147e999b8b52ee286973a57bbbd7f2/build.gradle#L25

kxu1026: foud the solution. group id.

shardulm94:
+1 LGTM

Issue #18, opened 2020-11-03
Coral prepends hardcoded catalog name while translating a hive view to Presto


phd3:
Coral hardcodes "hive" as the default catalog name while translating views to presto.

i.e. a view hql `SELECT a FROM test_schema.myview` translates to `SELECT a FROM hive.test_schema.myview` in Presto.

This is problematic for users who don't have "hive" as their default catalog name.

Presto Users reported this regression since coral was integrated for more complex hive view support: https://github.com/prestosql/presto/issues/5785

We need to modify the translation such that the underlying tables are resolved to the correct catalog name.

phd3:
cc @HotSushi

HotSushi:
Based on the discussion here: https://github.com/prestosql/presto/issues/5785?#issuecomment-721313566 , it makes sense to not prepend the catalog name in coral-presto. This is also the behavior for other modules like coral-spark.

PR #19, opened 2020-11-03
Coral-presto: produce db.table instead of catalog.db.table in Presto SQL

HotSushi:
Fixes: https://github.com/linkedin/coral/issues/18

With this change last two entries of full qualified table name are used in coral-presto generated SQL. Instead of generating "SELECT * FROM hive.db.table" coral-presto now generates "SELECT * FROM db.table".

This behavior is similar to coral-spark. 

Change is tested with FuzzyUnionViewTest.java

### DIFF
--- a/coral-presto/src/main/java/com/linkedin/coral/presto/rel2presto/RelToPrestoConverter.java
+++ b/coral-presto/src/main/java/com/linkedin/coral/presto/rel2presto/RelToPrestoConverter.java
@@ -14,6 +14,7 @@ import java.util.Map;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.core.Correlate;
 import org.apache.calcite.rel.core.Project;
+import org.apache.calcite.rel.core.TableScan;
 import org.apache.calcite.rel.core.Uncollect;
 import org.apache.calcite.rel.core.Values;
 import org.apache.calcite.rel.core.Window;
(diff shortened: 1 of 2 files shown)


coral-presto/src/main/java/com/linkedin/coral/presto/rel2presto/RelToPrestoConverter.java
```diff
@@ -168,6 +170,29 @@ public Result visit(Uncollect e) {
     return result(asNode, ImmutableList.of(Clause.FROM), e, null);
   }
 
+  /**
+   * This overridden function makes sure that the basetable names in the output SQL
+   * will be in the form of "dbname.tablename" instead of "catalogname.dbname.tablename"
+   *
+   * Presto can have configurable catalog names. In that case the HiveToRelConverter's default "hive" catalog will
+   * cause failures:  https://github.com/prestosql/presto/issues/5785. If catalogname is not prepended, presto uses
```
findepi, line 178: curious, where does the prepending happen?

HotSushi: Prepending is the default behavior of `RelToSqlConverter::visit(TableScan e)`. We override the behavior here.

funcheetah:
Thanks for fixing this.

PR #21, opened 2020-11-09
Add coral-beam for converting Calcite logical plan into Apache Beam API Java code

khaitranq:
End-to-end implementation and tests for translating from Pig Latin scripts into Apache Beam Java code

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,8 @@
 **Coral** is a library for analyzing, processing, and rewriting views defined in the Hive Metastore, and sharing them
 across multiple execution engines. It performs SQL translations to enable views expressed in HiveQL (and potentially
 other languages) to be accessible in engines such as [Presto](https://prestosql.io/),
-[Apache Spark](https://spark.apache.org/), and [Apache Pig](https://pig.apache.org/).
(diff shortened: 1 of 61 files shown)


wmoustafa:
Could you please remove Pig references and unit tests since Currently Coral supports reading from HiveQL only?

RohitR1:
Any update on Apache Beam integration? it will be very useful to have beam as well.

cc: @wmoustafa @khaitranq

wmoustafa:
Hi Rohit, thanks for reaching out! 

> Any update on Apache Beam integration? it will be very useful to have beam as well.
> 
> cc: @wmoustafa @khaitranq

Thanks @RohitR1 for reaching out! Do you have a use case for this? Before merging this PR, we need to validate that the two sides of the conversion in this PR (from Pig and to Beam) commute with other sides already implemented in Coral (e.g, from Hive QL and to Trino SQL), so one can convert Hive QL (or Spark SQL) to Beam for example. It might also be best to split the PRs to two halves to ensure there is no coupling between the two conversion sides. I think this is the major remaining steps here. We are not planning to pick up this now, but please feel free to make suggestions.

PR #29, opened 2020-12-15
Support more types in converting RelDatatype to Avro type in coral-schema

funcheetah:
Support more types in converting RelDatatype to Avro type


wmoustafa (approved):
Reviewed internally. Thanks for the patch.

PR #31, opened 2020-12-29
Rename Presto to Trino

martint:
PrestoSQL was rebranded as Trino. See: https://trino.io/blog/2020/12/27/announcing-trino.html


wmoustafa (approved):
Thanks @martint!

PR #35, opened 2021-01-08
Ensure field-properties and schema properties being replicated when cloning schema object

autumnust:
There are couple of methods attempting to replicate the schema object, a part of steps is to replicate the `List<Schema.Field>` since there's no public `clone` method for schema object. The replication was buggy since it doesn't carry over the field-level properties (which could be critical in some of cases like logical-type handling).  With this PR, this problem should be resolved. 

Note there are multiple places where the schema is being replicated.

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
@@ -31,6 +31,7 @@ import org.codehaus.jackson.JsonNode;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.linkedin.coral.com.google.common.base.Preconditions;
 import com.linkedin.coral.com.google.common.base.Strings;
 import com.linkedin.coral.schema.avro.exceptions.SchemaNotFoundException;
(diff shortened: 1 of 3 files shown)


autumnust:
@wmoustafa  Can you help take a look ? thanks.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -303,21 +317,13 @@ static Schema joinSchemas(@Nonnull Schema leftSchema, @Nonnull Schema rightSchem
...
-    for (Schema.Field field : leftSchema.getFields()) {
-      combinedSchemaFields
-          .add(new Schema.Field(field.name(), field.schema(), field.doc(), field.defaultValue(), field.order()));
-    }
-
-    for (Schema.Field field : rightSchema.getFields()) {
-      combinedSchemaFields
-          .add(new Schema.Field(field.name(), field.schema(), field.doc(), field.defaultValue(), field.order()));
-    }
+    List<Schema.Field> combinedSchemaFields = new ArrayList<>(cloneFieldList(leftSchema.getFields()));
```
wmoustafa, line 320: Why create a new ArrayList?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -303,21 +317,13 @@ static Schema joinSchemas(@Nonnull Schema leftSchema, @Nonnull Schema rightSchem
...
+    combinedSchemaFields.addAll(cloneFieldList(rightSchema.getFields()));
 
     Schema combinedSchema =
         Schema.createRecord(leftSchema.getName(), leftSchema.getDoc(), leftSchema.getNamespace(), leftSchema.isError());
     combinedSchema.setFields(combinedSchemaFields);
+    // Not replicate schema level props as it may not be well-defined in joined schema.
```
wmoustafa, line 326: We may change this to: "While we replicate schema-level props in addPartitionColsToSchema(), we avoid doing this here since it may lead to conflict"

That said, what would be the conflict, and does it warrant throwing an exception instead?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
autumnust: oh yeah thanks for catching - was using IDE's auto-fixing feature which added additional wrapper

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
autumnust: @wmoustafa  I actually mean to say the merging of properties from two schema is not a clearly defined behavior, being conflict with each other could be one case. 

Do you want to do something like: merge properties anyway and throw exception if conflict detected?

wmoustafa: Since you have examples of the properties, you may have a better idea if a conflict means skipping, picking one arbitrarily, or throwing an exception. I think throwing an exception may be a bit extreme.

autumnust: Discussed offline - decide to merge the properties in schema level from left to right.

wmoustafa:
Looks good to me but there is a build error. Could you please check?

wmoustafa (approved):
Thank you @autumnust for contributing this patch!

PR #37, opened 2021-01-19
Use Ivy to resolve UDF dependencies; use dynamic class loader to execute UDF type inference

wmoustafa:
This patch introduces the following changes:
1) It uses Ivy instead of Groovy to resolve UDF dependencies. When using Coral as a shaded jar, Groovy is not fully shaded. It turns out that Groovy is difficult to fully shade due to the fact that it depends on loading and saving a binary file referencing class names which are required to be shaded. I attempted to extend the shading plugin to rewrite binary files referencing class names, but that was not sufficient as shading still required rewriting some class signatures, or otherwise some auto-generated code will throw exceptions. Instead of fully shading Groovy, using Ivy sounds like a more feasible approach since it is more shading-friendly.
2) It uses a dynamic class loader to infer the UDF type. Previously, the ObjectInspector class was loaded from the current class loader. However, this does not work since the ObjectInspector class is intended to be used to call the `initialize()` method of the UDF, which is loaded using a URL class loader whose ObjectInspector class is not interoperable with the current class loader's ObjectInspector. Therefore, in this patch, we call the `initialize()` method using objects and classes that belong to the URL class loader exclusively, and interoperate with the current class loader by String encoding and decoding of intermediate results.

Testing was done by running coral-tools regression test on all production views.

### DIFF
--- a/coral-hive/build.gradle
+++ b/coral-hive/build.gradle
@@ -17,8 +17,7 @@ dependencies {
     exclude group: 'org.apache.avro', module: 'avro-tools'
   }
   compile deps.'hadoop'.'hadoop-common'
-
-  compile 'org.codehaus.groovy:groovy-all:2.4.4'
+  compile deps.'ivy'
 
   // required for hiveSQL parser code
   compile deps.'antlr-runtime'
--- a/coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/CoralTestUDF.java
+++ b/coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/CoralTestUDF.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2018-2020 LinkedIn Corporation. All rights reserved.
+ * Copyright 2018-2021 LinkedIn Corporation. All rights reserved.
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
--- a/coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/CoralTestUDF.java
+++ b/coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/CoralTestUDF.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2018-2021 LinkedIn Corporation. All rights reserved.
+ * Copyright 2018-2020 LinkedIn Corporation. All rights reserved.
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/ArtifactsResolver.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/ArtifactsResolver.java
@@ -5,263 +5,202 @@
  */
 package com.linkedin.coral.hive.hive2rel.functions;
 
-import java.io.BufferedOutputStream;
 import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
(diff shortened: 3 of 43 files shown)


funcheetah:
Thanks a lot for this patch!

Is it tested with engine integration (e.g. Spark)?

Please resolve the merge conflicts. Thanks.

coral-hive/build.gradle
```diff
@@ -17,6 +17,7 @@ dependencies {
     exclude group: 'org.apache.avro', module: 'avro-tools'
   }
   compile deps.'hadoop'.'hadoop-common'
+  compile 'org.apache.ivy:ivy:2.4.0'
```
cwsteinbach, line 20: Please add this to gradle/dependencies.gradle

coral-hive/build.gradle
```diff
@@ -17,6 +17,7 @@ dependencies {
...
 
   compile 'org.codehaus.groovy:groovy-all:2.4.4'
```
cwsteinbach, line 22: Remove this, right?

cwsteinbach (changes requested):

cwsteinbach (changes requested):
> Testing was done by running coral-tools regression test on all production views.

Please add the tests to this patch.

coral-hive/build.gradle
wmoustafa: Done. I was planning to do this along with the other comment, and rebasing as follow-ups, but somehow the comments came too soon :)

coral-hive/build.gradle
wmoustafa: Yes.

wmoustafa:
@funcheetah, yes I tested in Spark after shading. It now works as expected.
@cwsteinbach, this change is supposed to be a replacement to an existing functionality. I agree there should have been test along with the feature when it was first introduced, but for now I think it is okay to rely on integration tests, which are also much more comprehensive (testing thousands of views for regression). The integration test tool is not open-source though, and its testing logs are internal. I can share offline. For the unit tests, the reason the testing may have not been added yet to this feature is that it requires communicating with an Artifact Management Repository, and adding a test means creating an instance of that system locally as part of the test. @HotSushi is working on this feature for testing Spark integration, and we will leverage it in Coral once it is ready.

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/CoralTestUDF.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2018-2020 LinkedIn Corporation. All rights reserved.
+ * Copyright 2018-2021 LinkedIn Corporation. All rights reserved.
```
cwsteinbach, line 2: The spotlessApply task only checks/updates files that have changed since origin/master. Files like this where only the copyright date was modified should be backed out.

cwsteinbach (changes requested):

cwsteinbach:
> @HotSushi is working on this feature for testing Spark integration, and we will leverage it in Coral once it is ready.

Please create a GH issue to ensure we don't forget about this. Thanks!

wmoustafa:
> Please create a GH issue to ensure we don't forget about this. Thanks!
Tracked in #38.

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/CoralTestUDF.java
wmoustafa: Reverted.

funcheetah:
This patch looks good to me. Thanks!

PR #39, opened 2021-01-25
Add .java-version for jenv

losipiuk:
PR add .java-version file which will allow for automatic selection of java 1.8 on environments when multiple java runtimes are managed by jenv.


findepi (approved):
Having this would save me some time figuring out problems with local build, so i guess it may save trouble to someone in the future.

PR #36, opened 2021-01-12
Migrated to new Shipkit plugins

mockitoguy:
This change is fully backwards compatible, please review!

- enables concurrent releases
- avoids unnecessary version bump commits
- enables using higher versions of Gradle ("org.shipkit.java" plugin is deprecated)
- removed read-only token from checked-in files (even though it is read read-only, it should not be exposed to unexpected users/bots)

Tested by:
- ran `./gradlew publishToMavenLocal` with the old code, generated all artifacts and poms, then I ran the new code and compared the artifacts and poms
- executed the release steps locally:
    - `./gradlew githubRelease` -> released to GitHub: https://github.com/mockitoguy/coral/releases/tag/v1.0.19
    - `./gradlew bintrayUpload` -> released to Bintray: https://bintray.com/shipkit/examples/coral/1.0.19

PR #44, opened 2021-02-06
Revert 6432dde990f73821610d6b7e8f8e21d210404750: Migrate to new Shipkit plugin

wmoustafa:
Published artifact signing files are no longer generated as of v1.0.20, and hence Coral repo in Bintray cannot be synced with Maven Central. Version 1.0.20 contains two changes. One of the two changes (https://github.com/linkedin/coral/pull/36) upgrades Shipkit plugin to a new version. At the same time, LinkedIn Bintray account did not seem to have recent changes that would make those files no longer be generated. Current theory is that one of the changes in v1.0.20 caused the regression. This patch reverts one of those changes to check if it is correlated.


shardulm94 (approved):

wmoustafa:
Reverting seems to have successfully resulted in the files being generated again https://bintray.com/linkedin/maven/coral/1.0.19#files/com%2Flinkedin%2Fcoral%2Fcoral-hive%2F1.0.19.

PR #47, opened 2021-02-16
Use enclosing SqlOperator UDF name when translating Transport UDFs in Spark

wmoustafa:
Before this patch, UDF names were hard-coded in the Transport UDF map (`TransportableUDFMap`). This could lead functions registered in the same Spark session by the end users using the same name to collide with those hard-coded names, especially that the hard-coded names are commonly used. This patch uses the view-specific UDF name from the enclosing `SqlOperator` instead. The `DaliSpark.testDaliUdf()` test is updated to capture the new expected behavior.


funcheetah:
Thanks for the patch! Could you please add information regarding test done for this patch? Thanks.

wmoustafa:
@funcheetah, updated the description.

PR #49, opened 2021-02-22
Use assertEquals in HiveToPrestoConverterTest

ebyhr:
assertTrue doesn't report helpful information when failed.


wmoustafa:
Could you please add context in the description?
Also, I think it makes sense to replace `AS t[i]` with a canonical/empty string and then compare the expected and actual to avoid test flakiness in case those `t[i]` indexes change based on some criteria.

wmoustafa:
Also, is there a way to force the `AS` in the input SQL (i.e., view definition) so that the output is not left to Calcite to determine?

findepi:
I hit the same problem that exception message produced by a test is not helpful. It's addressed in https://github.com/linkedin/coral/pull/61

However, the underlying problem is that test cases depend on each other, as alias generation uses a shared counter.
https://github.com/linkedin/coral/pull/61 doesn't solve this, and using `contains` is not sufficient, as the aliases show up not only at the end (where they can be easily omitted), but also within query text.

A longer term solution could be to reset the alias generation counter, or even better make it a per-translation scoped.

PR #51, opened 2021-03-02
[Reference PR] Skip publishing to Bintray from Shipkit; publish to Maven Central instead

wmoustafa:
[Bintray is sunset](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/) starting 3/1/2021. This patch adapts Shipkit plugin integration so that it skips the Bintray publication step. Also, it sets up the `maven-publish` plugin so that it can publish to Maven Central. This patch is based on combining guides from [Shipkit](https://github.com/mockito/shipkit/blob/master/docs/features/publishing-binaries-using-maven-publish-plugin.md), [OSSRH with Gradle](https://central.sonatype.org/pages/gradle.html) and an [example Maven Central + Gradle integration from dex-test-parser](https://github.com/linkedin/dex-test-parser/blob/main/parser/build.gradle).

This patch is tested by running `./gradlew publishToMavenLocal` which publishes artifacts locally. I found the expected artifacts in the local m2 repo, but did not notice he signing files. I will wait after check-in to see if they will be generated in the non-local publication case.

The patch was also tested by running Shipkit's release in the dry run mode: `./gradlew performRelease -PdryRun`. It works all the way up to pushing. I did not provide a Github write token so that it does not accidentally push an undesired change.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,14 @@
 plugins {
   id "com.diffplug.spotless" version "5.8.2"
   id 'com.github.johnrengelman.shadow' version '5.2.0'
-  id "org.shipkit.java" version "2.3.4"
+  // The following Shipkit plugins are required to make Shipkit skip publishing to Bintray.
+  // Bintray is sunset starting 3/1/2021.
+  // For more details on how to integrate Shipkit with other repositories please see
+  // https://github.com/mockito/shipkit/blob/master/docs/features/publishing-binaries-using-maven-publish-plugin.md
+  id "org.shipkit.travis" version "2.3.4"
+  id "org.shipkit.github-pom-contributors" version "2.3.4"
+  id "org.shipkit.release-notes" version "2.3.4"
+  id "org.shipkit.compare-publications" version "2.3.4"
 }
 
 configurations {
@@ -18,9 +25,6 @@ allprojects {
 
   repositories {
     mavenCentral()
-    maven {
-      url 'https://linkedin.bintray.com/maven/'
-    }
   }
 
   spotless {
@@ -39,8 +43,10 @@ allprojects {
     }
   }
 }
+tasks.updateReleaseNotes.publicationRepository = "https://repo1.maven.org/maven2/com/linkedin/coral/"
+tasks.updateReleaseNotesOnGitHub.publicationRepository = "https://repo1.maven.org/maven2/com/linkedin/coral/"
 
-subprojects {
+subprojects { subproject ->
   plugins.withType(JavaPlugin) {
     dependencies {
       testCompile deps.'testing'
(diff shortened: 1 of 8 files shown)


coral-spark/build.gradle
```diff
@@ -20,5 +21,5 @@ dependencies {
 }
 
 artifacts {
-  archives jar, javadocJar, sourcesJar
+  archives jar
```
funcheetah, line 24: Is there any reason to remove javadocJar and sourcesJar? There are needed to publish to maven central

wmoustafa: Those types of jars are referenced in the main `build.gradle`. Including them here was throwing an error. I verified that they existed in the local m2 repo after the local release step.

build.gradle
```diff
@@ -59,4 +65,95 @@ subprojects {
...
+            subproject.name + "-" + tasks.fetchReleaseNotes.previousVersion + "-sources.jar"
+        maven(MavenPublication) { publication ->
+          tasks.performRelease.dependsOn(tasks.publish)
+          from components.java
+          artifact(javadocJar) {
+            classifier = 'javadoc'
+          }
+          artifact(sourcesJar) {
+            classifier = 'sources'
+          }
```
drewhannay, line 87: If your modules are all using the `java` plugin, you should be able to use this block:
```groovy
java {
    withSourcesJar()
    withJavadocJar()
}
```
in each of your modules (or in this file, outside the `publications` block) and then source and javadoc jars will get added to the publication automatically since you're using `from components.java`

build.gradle
```diff
@@ -59,4 +65,95 @@ subprojects {
...
+          pom {
+            groupId = 'com.linkedin.coral'
+            artifactId = subproject.name
+            name = 'Coral'
+            description = 'Coral is a SQL analysis, translation, and rewrite engine'
```
drewhannay, line 92: Up to you, but you might consider having the name and description describe what the actual artifact is. As a consumer, I find that to be a nicer use of name + description here.

coral-hive/build.gradle
```diff
@@ -46,7 +48,7 @@ generateGrammarSource {
   ]}
 
 artifacts {
-  archives jar, javadocJar, sourcesJar
+  archives jar
 }
```
drewhannay, line 52: I think you shouldn't need to add the `jar` here either, since you're using `from components.java` in your `publications` block. The `java` plugin should be adding that for you automatically (same comment for all the other modules)

mockitoguy:
We need to use "io.github.gradle-nexus.publish-plugin" plugin so that the publication to Maven Central is fully automatic, see example here: https://github.com/shipkit/shipkit-demo/blob/master/gradle/release.gradle#L20

Also, let's try to move off the deprecated Shipkit plugins. E.g. let's try to re-merge https://github.com/linkedin/coral/pull/36 I'll put some cycles on it.

mockitoguy:
Here's my suggestion how to move forward: https://github.com/linkedin/coral/pull/52

PR #54, opened 2021-03-11
Adjust array indices for non-literals

shardulm94:
Hive uses 0-based array indices while ANSI SQL/Presto uses 1-based indices. When we convert from Hive QL to Calcite RelNodes we add 1 to all array indices. Hence `arrayField[0]` is converted to `arrayField[1]`. However today this adjustment only considers integer literals, all other indices are ignored and passed through as-is. This causes non-literal indices to not be adjusted, e.g. HiveQL `arrayField[size(arrayField) - 1]` still remains `arrayField[size(arrayField) - 1]` in RelNode leading to silent correctness issues.

This PR translates any non-integer literals by using a SQL function call to add one to the index.
`arrayField[size(arrayField) - 1]` => `arrayField[size(arrayField) - 1 + 1]`

I decided to keep the original implementation for integer literals for simplicity of the converted SQL in the most common cases, so that `arrayField[0]` is converted to `arrayField[1]` instead of `arrayField[0 + 1]`.

cc: @wmoustafa @funcheetah

### DIFF
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveRelConverter.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveRelConverter.java
@@ -158,8 +158,8 @@ public class HiveRelConverter extends RelShuttleImpl {
             RexLiteral newItemRef = rexBuilder.makeExactLiteral(new BigDecimal(val + 1), itemRef.getType());
             return rexBuilder.makeCall(call.op, columnRef, newItemRef);
           } else {
-            RexNode oneBasedIndex = rexBuilder.makeCall(
-                SqlStdOperatorTable.PLUS, itemRef, rexBuilder.makeExactLiteral(BigDecimal.ONE));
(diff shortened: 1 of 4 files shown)


funcheetah (commented):
Thanks for the patch! It looks good to me. Let's do integration test first before checking in.

shardulm94:
Our internal regression tests passed!

findepi:
@funcheetah i see this in a tagged release here https://github.com/linkedin/coral/releases/tag/v1.0.29

do you plan to push that to Maven Central?
the latest i see on https://search.maven.org/artifact/com.linkedin.coral/coral-presto is 1.0.25

wmoustafa:
@findepi We were working on migrating our release automation system after [decommissioning Bintray](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/). Version [1.0.33 should be published](https://repo1.maven.org/maven2/com/linkedin/coral/coral-presto/) now.

PR #23, opened 2020-11-18
Add translate function mapping in Presto

ebyhr:
https://prestosql.io/docs/current/functions/string.html#translate

PR #55, opened 2021-03-14
Make CARDINALITY return INT in Trino

shardulm94:
We convert the `size` function in Hive to `cardinality` in Trino (FKA PrestoSQL). However, `size` returns an `int` while `cardinality` returns `bigint`. This causes type validation issues in Trino when reading the view to fail with the error `<view_name> is stale; it must be re-created`.

When unparsing the `cardinality` function in Trino, we should add `CAST` to an integer to retain the semantics of Hive's `size` and Calcite's `cardinality`

### DIFF
--- a/coral-presto/src/main/java/com/linkedin/coral/presto/rel2presto/functions/PrestoCardinalityFunction.java
+++ b/coral-presto/src/main/java/com/linkedin/coral/presto/rel2presto/functions/PrestoCardinalityFunction.java
@@ -1,0 +1,40 @@
+/**
+ * Copyright 2018-2021 LinkedIn Corporation. All rights reserved.
+ * Licensed under the BSD-2 Clause license.
+ * See LICENSE in the project root for license information.
+ */
+package com.linkedin.coral.presto.rel2presto.functions;
+
+import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlFunction;
+import org.apache.calcite.sql.SqlFunctionCategory;
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.type.OperandTypes;
+import org.apache.calcite.sql.type.ReturnTypes;
+
+
+/**
+ * Represents the CARDINALITY function in Presto
+ *
+ * Identical the CARDINALITY SQL function, but CARDINALITY in PRESTO returns a BIGINT so we
+ * need to cast it to an INTEGER when unparsing
+ */
+public class PrestoCardinalityFunction extends SqlFunction {
+  public static final PrestoCardinalityFunction INSTANCE = new PrestoCardinalityFunction();
+
+  public PrestoCardinalityFunction() {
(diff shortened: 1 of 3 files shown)


coral-presto/src/main/java/com/linkedin/coral/presto/rel2presto/functions/PrestoCardinalityFunction.java
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  @Override
+  public void unparse(SqlWriter writer, SqlCall call, int leftPrec, int rightPrec) {
+    final SqlWriter.Frame frame = writer.startFunCall(SqlStdOperatorTable.CAST.getName());
+    super.unparse(writer, call, leftPrec, rightPrec);
+    writer.sep("AS");
+    writer.sep("INTEGER");
+    writer.endFunCall(frame);
+  }
```
wmoustafa, line 39: Does it make sense to add the `CAST` in `RelToPrestoConverter`?

shardulm94: That was my first option as well, however the issue is that you have `x` with is a cardinality call, the result will be `CAST(x AS INT)` and now the converter will try to visit the `CAST` and the run into `x` again and try to add a cast again going into infinite recursion. Unfortunately when the call is delegated to our visitor, we don't have any context about the parent to be able to break the recursion.

wmoustafa: Another option is to add it to the `CalcitePrestoUDFMap`, and apply an operand transformer? For example, the `Truncate` function goes through a transformation:
```
    createUDFMapEntry(UDF_MAP, SqlStdOperatorTable.TRUNCATE, 2, "TRUNCATE",
        "[{\"op\":\"*\",\"operands\":[{\"input\":1},{\"op\":\"^\",\"operands\":[{\"value\":10},{\"input\":2}]}]}]",
        "{\"op\":\"/\",\"operands\":[{\"input\":0},{\"op\":\"^\",\"operands\":[{\"value\":10},{\"input\":2}]}]}");
```
Operand Transformer is just a convenience method to return a new `RexNode`. If the string based operand transformation of the operand transformer is not expressive enough for this case, it may be fair to use a custom `RexNode`.

shardulm94: I had also tried this by creating a number operator called CAST as a result transformer. It has the same problem of infinite recursion.

wmoustafa: Discussed another approach offline: infinite recursion is due to the collision between the Calcite name of the operator and the Presto name (`Truncate` in this case).  We have another symptom of this problem in https://github.com/linkedin/coral/pull/23. I think Coral needs its unambiguous operator names to avoid cases like this and https://github.com/linkedin/coral/pull/23. The same suggestion in [this comment](https://github.com/linkedin/coral/pull/23#issuecomment-774927641) may help here.

ljfgem:
Closing this PR given #185 has been merged.

PR #60, opened 2021-03-24
Test LATERAL VIEW EXPLODE on string array

findepi:


findepi:
cc @wmoustafa

wmoustafa (approved):
Thanks for adding the test @findepi!

PR #62, opened 2021-03-25
Support for LATERAL VIEW json_tuple(..) jt AS a,b

findepi:
~Based on https://github.com/linkedin/coral/pull/59, https://github.com/linkedin/coral/pull/60, https://github.com/linkedin/coral/pull/61~

Fixes https://github.com/linkedin/coral/issues/53

PR #63, opened 2021-03-30
Throw error when LATERAL VIEW uses unsupported table operator

findepi:
Previously, an `EXPLODE(...)` was assumed.


findepi:
Currently based on https://github.com/linkedin/coral/pull/62

wmoustafa (approved):
Thanks for this patch @findepi.

PR #64, opened 2021-03-30
Assert strict equality in HiveToPrestoConverterTest

findepi:
Previously the VIEW conversion test asserted on the prefix, stripping
out meaningless final alias. However, in more complex cases, the alias
is used in other parts of the converted query text. Asserting on full
text means the expected results are self-contained now.


findepi:
Currently based on https://github.com/linkedin/coral/pull/62

cc @ebyhr as it relates to https://github.com/linkedin/coral/pull/49

wmoustafa (approved):
Nice improvement @findepi. Thanks!

PR #70, opened 2021-04-12
Support 'SELECT NULL null_field' and union schemas with NULL type field in coral-schema

ljfgem:
Views like `premium_mp.inapp_payment_event` are created using SQL like 
```
SELECT NULL null_field, ..., FROM table1
UNION ALL
SELECT ... FROM table2
```
There are 2 issues when coral-schema process views like it.
1. While visiting `SELECT NULL null_field`, `SchemaUtilities.appendField` is called to append `NULL` to the type of `null_field`, but there is no check to determine if the provided `null_field`'s type is NULL or not. This PR adds the checking before union NULL with the provided field type to avoid union 2 NULL values.
2. After checking and leaving the type of `null_field` as NULL, there is no support for NULL type in `SchemaUtilities.isUnionRecordSchemaCompatible` and `SchemaUtilities.setupNestedNamespaceForRecord`. This PR also adds the NULL in the cases.

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
@@ -171,7 +171,7 @@ class SchemaUtilities {
     Schema fieldSchema = RelDataTypeToAvroType.relDataTypeToAvroTypeNonNullable(fieldRelDataType, fieldName);
 
     // TODO: handle default value properly
-    if (isNullable) {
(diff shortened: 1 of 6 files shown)


wmoustafa (commented):
Could you add more context to the description? For example, a minimal view or data type that causes this. Also, for the PR title, it is better to state the change as opposed to what bug this PR is fixing. When stating the change, we would state what we did to fix the bug instead of stating the bug itself.

funcheetah:
Please add a unit test for this change.

PR #75, opened 2021-04-23
Manually add GenericProject and IsTestMemberId to static registry

rzhang10:
Manually add GenericProject and IsTestMemberId to static registry to bypass dynamic inference of the UDF's return type.

Tested via gradle `translate` task.


wmoustafa:
Do the views that reference those UDFs get translated successfully now?

funcheetah (approved):

rzhang10:
> Do the views that reference those UDFs get translated successfully now?

Yes, they got translated successfully now.

funcheetah (approved):

PR #79, opened 2021-04-30
Alternate implementation to #74

wmoustafa:
Going through this, I see there could be further simplifications to `SchemaUtilities`. I may do them in a separate PR.

The only difference from https://github.com/linkedin/coral/pull/74 is that if the `leftSchema` and `rightSchema` do not match, the exception is thrown without the field names since the check has moved to `mergeUnionRecordSchema`. I think this is more reasonable, because field names could only be obtained when in a `Record`. Since the merging is recursive, I would say that all recursion levels should be treated equally, and hence reporting the error at the `Schema` level is appropriate.

Unit tests pass.


coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -436,46 +445,62 @@ private static boolean isUnionSchemaCompatible(@Nonnull Schema leftSchema, @Nonn
...
-        boolean isSameType = leftSchema.getType() == rightSchema.getType();
-        boolean isSameNamespace = Objects.equals(leftSchema.getNamespace(), rightSchema.getNamespace());
-
-        return isSameType && (!strictMode || isSameNamespace);
+        compatible &= isSameNamespace(leftSchema, rightSchema, strictMode);
+        outputSchema = leftSchema;
+        break;
       case ENUM:
-        boolean isSameEnumType = (leftSchema.getType() == rightSchema.getType());
         boolean isSameSymbolSize = (leftSchema.getEnumSymbols().size() == rightSchema.getEnumSymbols().size());
```
ljfgem, line 455: There is an issue if we don't check the type of `rightSchema` before calling `rightSchema.getEnumSymbols()`, which also happens in `RECORD`, `MAP`, `ARRAY`, maybe we need to throw the exception first if `compatible == false` before `switch`.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -436,46 +445,62 @@ private static boolean isUnionSchemaCompatible(@Nonnull Schema leftSchema, @Nonn
...
-            AvroSerdeUtils.isNullableType(leftSchema) && AvroSerdeUtils.isNullableType(rightSchema);
-
-        Schema leftOtherType = AvroSerdeUtils.getOtherTypeFromNullableType(leftSchema);
-        Schema rightOtherType = AvroSerdeUtils.getOtherTypeFromNullableType(rightSchema);
-        boolean isOtherTypeUnionCompatible = isUnionSchemaCompatible(leftOtherType, rightOtherType, strictMode);
-
-        return isSameUnionType && isBothNullableType && isOtherTypeUnionCompatible;
-
+        // If we are in this branch, we know that neither leftSchema nor rightSchema is a nullable union, but can still be a union
+        outputSchema = getUnionFieldSchema(leftSchema, rightSchema, strictMode);
```
ljfgem, line 472: At present, this case doesn't happen because non-nullable union type like `[int, boolean]` is not supported in coral-hive.
But if it is supported in the future, I think 
`outputSchema = getUnionFieldSchema(leftSchema, rightSchema, strictMode);`
will cause infinite recursion.

ljfgem (commented):
Tested on the production views over last 6 month, it can work, but I think there are 2 small issues.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -394,40 +384,59 @@ static boolean isUnionRecordSchemaCompatible(@Nonnull Schema leftSchema, @Nonnul
...
+      return makeNullable(rightSchema);
+    }
+    if (rightSchemaType == NULL) {
+      return makeNullable(leftSchema);
+    }
+    if (isNullableType(leftSchema) || isNullableType(rightSchema)) {
+      return makeNullable(getUnionFieldSchema(makeNonNullable(leftSchema), makeNonNullable(rightSchema), strictMode));
+    }
+
+    boolean compatible = leftSchemaType == leftSchemaType;
```
ljfgem, line 435: boolean compatible = leftSchemaType == rightSchemaType;

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
wmoustafa: Good point. Refactored. Please take another look.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
wmoustafa: True. Removed.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -393,90 +383,117 @@ static boolean isUnionRecordSchemaCompatible(@Nonnull Schema leftSchema, @Nonnul
...
-            AvroSerdeUtils.isNullableType(leftSchema) && AvroSerdeUtils.isNullableType(rightSchema);
+    if (leftSchemaType == rightSchemaType) {
+      Schema outputSchema;
+      boolean compatible = true;
 
-        Schema leftOtherType = AvroSerdeUtils.getOtherTypeFromNullableType(leftSchema);
-        Schema rightOtherType = AvroSerdeUtils.getOtherTypeFromNullableType(rightSchema);
-        boolean isOtherTypeUnionCompatible = isUnionSchemaCompatible(leftOtherType, rightOtherType, strictMode);
+      // Each of the following case options adjusts the value of "compatible" with additional conditions
+      // and sets the value of "outputSchema". This method will return "outputSchema" if "compatible" is true.
```
ljfgem, line 439: `outputSchema` and `compatible` can be removed

wmoustafa: Done.

ljfgem (approved):
LGTM, thank you for this patch

funcheetah (approved):

PR #81, opened 2021-05-03
Avoid throwing exception 'Unable to fetch dependencies' if 'report.hasError()' in ArtifactsResolver

ljfgem:
After removing this exception, views which threw the exception `Unable to fetch dependencies: [XXX]` could be translated successfully or threw other known exceptions such as 
`java.lang.RuntimeException: Unable to find org.apache.hadoop.hive.ql.udf.generic.GenericUDF.initialize() on: <UDF>` or `java.lang.RuntimeException: Unable to call org.apache.hadoop.hive.ql.udf.generic.GenericUDF.initialize() on: <UDF>`.

Tested on production views over last 6 month.

### DIFF
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/ArtifactsResolver.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/ArtifactsResolver.java
@@ -124,7 +124,11 @@ public class ArtifactsResolver {
         .setOutputReport(true).setValidate(true).setTransitive(dependencySpec.transitive);
     resolveOptions.setLog(LogOptions.LOG_QUIET);
     try {
-      return _ivyInstance.resolve(md, resolveOptions);
+      ResolveReport report = _ivyInstance.resolve(md, resolveOptions);
+      if (report.hasError()) {
+        LOG.warn("Unable to fetch dependencies: " + report.getAllProblemMessages());
+      }
+      return report;
     } catch (ParseException | IOException e) {
       throw new RuntimeException("Unable to fetch dependencies", e);
     }
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/ArtifactsResolver.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/ArtifactsResolver.java
@@ -124,7 +124,7 @@ public class ArtifactsResolver {
         .setOutputReport(true).setValidate(true).setTransitive(dependencySpec.transitive);
     resolveOptions.setLog(LogOptions.LOG_QUIET);
     try {
-      ResolveReport report = _ivyInstance.resolve(md, resolveOptions);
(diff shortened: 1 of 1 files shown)


coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/ArtifactsResolver.java
```diff
@@ -124,14 +124,8 @@ private ResolveReport getDependencies(DependencySpec dependencySpec) {
         .setOutputReport(true).setValidate(true).setTransitive(dependencySpec.transitive);
     resolveOptions.setLog(LogOptions.LOG_QUIET);
     try {
-      final ResolveReport report = _ivyInstance.resolve(md, resolveOptions);
-      if (report.hasError()) {
-        throw new RuntimeException("Unable to fetch dependencies: " + report.getAllProblemMessages());
```
funcheetah, line 129: Is this the case that other exceptions will be wrongly captured here before?

ljfgem: It might be this case, because after removing it, one affected view could be translated well even if there is still error in `report` like:
```
unresolved dependency: XXX: configuration not found in XXX: 'master'. It was required from YYY compile
```

And for most of the affected production views (8 in 10), after removing this exception and adding UDFs manually to avoid exceptions like `java.lang.RuntimeException: Unable to find/call org.apache.hadoop.hive.ql.udf.generic.GenericUDF.initialize() on: <UDF>`, there is still an exception: `java.lang.ArrayIndexOutOfBoundsException`, which is because Calcite doesn't support source SQL with "SELECT table.nested.field FROM (..)" logic. For more details, see: [CALCITE-1518](https://issues.apache.org/jira/browse/CALCITE-1518). So they still can't be translated well.

funcheetah: I see. Basically `report.hasError()` can have false positive resolving exceptions. A followup question will be: if we remove it, will there be false negative (it is resolving exception but not get captured)?

ljfgem: I tested this PR on **all** the production views.
Among the 32 affected views which failed because of `Unable to fetch dependencies: [XXX]`, 4 of them could be translated well with this PR, and 28 of them would throw known exceptions (most are `java.lang.ArrayIndexOutOfBoundsException`). Therefore, I think at least for all the existing views, there is no false negative case.

ljfgem:
> when you say that with this change some views now throw **known** exceptions, does that mean these are exceptions we already analyzed and plan to work on, or does it mean they are valid exceptions and the views has issues that need to be addressed?

We already analyzed these exceptions and plan to work on them, which are thrown because of other reasons.

shenodaguirguis (approved):

PR #82, opened 2021-05-04
Decouple Coral SchemaUtilities with Avro Version

kxu1026:
By applying AvroCompatibilityHelper, we can decouple the appendField logic in SchemaUtilities with a particular avro version.


coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -211,12 +213,11 @@ static void appendField(@Nonnull String fieldName, @Nonnull Schema.Field field,
     Preconditions.checkNotNull(field);
     Preconditions.checkNotNull(fieldAssembler);
 
-    JsonNode defaultValue = field.defaultValue();
-
     SchemaBuilder.GenericDefault genericDefault = fieldAssembler.name(fieldName).type(field.schema());
-    if (defaultValue != null) {
+    try {
+      Object defaultValue = AvroCompatibilityHelper.getGenericDefaultValue(field);
```
ljfgem, line 218: Is it because `field.defaultValue()` is no longer supported in new avro versions?
If so, there are some other `field.defaultValue()` in this file, should we also change them?

kxu1026: good point.

wmoustafa:
Can we postpone this change until it is really needed? I do not think we should add another dependency when we know Hive pulls in 1.7 already.

kxu1026:
> Can we postpone this change until it is really needed? I do not think we should add another dependency when we know Hive pulls in 1.7 already.

sure. I just want to propose a way to address our avro 1.9 problem by AvroCompatibilityHelper

kxu1026:
drop and defer to future discussion.

PR #86, opened 2021-05-11
Let coral hive2rel treat a union typeinfo as a struct typeinfo, inste…

rzhang10:
…ad of throwing exception

Also added `extract_union` function to static registry.

Tested on hive view which contains `extract_union` and coral can translate it correctly to spark sql.

### DIFF
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
@@ -12,7 +12,6 @@ import java.util.function.Predicate;
 
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
-import org.apache.calcite.rel.type.RelDataTypeImpl;
 import org.apache.calcite.sql.SqlCallBinding;
 import org.apache.calcite.sql.SqlIdentifier;
 import org.apache.calcite.sql.SqlOperandCountRange;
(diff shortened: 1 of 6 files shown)


wmoustafa:
Could you please add unit tests?

shenodaguirguis:
+1 to @wmoustafa's request to add tests. LGTM o.w.

shenodaguirguis (approved):
thanks for adding the unit tests.

funcheetah:
@rzhang10 @shenodaguirguis  is this PR integration tested for all views since it is not a trivial change?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
```diff
@@ -316,6 +316,7 @@ public RelDataType inferReturnType(SqlOperatorBinding opBinding) {
 
     createAddUserDefinedFunction("array_contains", ReturnTypes.BOOLEAN, family(SqlTypeFamily.ARRAY, SqlTypeFamily.ANY));
     createAddUserDefinedFunction("sort_array", ARG0, ARRAY);
+    createAddUserDefinedFunction("extract_union", ARG0, or(ANY, family(SqlTypeFamily.ANY, SqlTypeFamily.INTEGER)));
```
wmoustafa, line 319: For the 2-argument case, the output type will not be `ARG0`. Also, for the 1-argument case, the type would always be a struct, but it seems in the test, we define the table as containing `uniontype` in the tests, and hence the output could be a `uniontype` if we use `ARG0`?

PR #88, opened 2021-05-18
Add lowercase `regexp` and `rlike` to StaticHiveFunctionRegistry

ljfgem:
Tested on the production views which failed because of the exception:
```
com.linkedin.coral.hive.hive2rel.functions.UnknownSqlFunctionException: Unknown function name: regexp
```
They could be translated well with this patch.

### DIFF
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
@@ -73,7 +73,9 @@ public class StaticHiveFunctionRegistry implements HiveFunctionRegistry {
 
     // operators
     addFunctionEntry("RLIKE", HiveRLikeOperator.RLIKE);
+    addFunctionEntry("rlike", HiveRLikeOperator.RLIKE);
     addFunctionEntry("REGEXP", HiveRLikeOperator.REGEXP);
+    addFunctionEntry("regexp", HiveRLikeOperator.REGEXP);
     addFunctionEntry("!=", NOT_EQUALS);
     addFunctionEntry("==", EQUALS);
 
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
@@ -72,6 +72,9 @@ public class StaticHiveFunctionRegistry implements HiveFunctionRegistry {
     //addFunctionEntry("in", SqlStdOperatorTable.IN);
 
     // operators
+    // We need to add both lower and upper cases for operators because DaliOperatorTable.lookupOperatorOverloads
+    // resolves the operator functions case-sensitively. If we just add the lowercase operator,
+    // DaliOperatorTable.lookupOperatorOverloads can't resolve the uppercase operator, and there will be null pointer exception.
     addFunctionEntry("RLIKE", HiveRLikeOperator.RLIKE);
     addFunctionEntry("rlike", HiveRLikeOperator.RLIKE);
     addFunctionEntry("REGEXP", HiveRLikeOperator.REGEXP);


coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
```diff
@@ -73,7 +73,9 @@
 
     // operators
     addFunctionEntry("RLIKE", HiveRLikeOperator.RLIKE);
+    addFunctionEntry("rlike", HiveRLikeOperator.RLIKE);
```
shenodaguirguis, line 76: Wouldn't it be better if we modify `addFunctionality` (and `lookup`) to  lower-case all entries? This seems to match  what`CalcitePrestoUDFMap.addDaliUDFs()` does with function names

ljfgem: Thank you for review! In `lookup` function, we need to support `isCaseSensitive`, so I don't think we should use lower-case all entries.
And `CalcitePrestoUDFMap.addDaliUDFs()` does that only for Dali UDFs starting with `com.linkedin`, these 2 functions (`regexp` and `rlike`) are not this case.

shenodaguirguis: I understand. I am saying we choose to handle Dali UDFs with lower-casing, so we can be consistent. 
Since `lookup()` already accepts `isCaseSensitive` flag, why can't we leverage that, instead of adding 2 new functions just for the case?
If we handle/fix this with a generic approach, we will cover all functions that might get into this, and also with all different casing (e.g., "Regex", "RegEx", etc).

funcheetah (approved):

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
ljfgem: We just need to add lower-case one to work for all the different casing when `isCaseSensitive == false`.
However, operator function is kind of special, I added a comment to the code:

We need to add both lower and upper cases for operators because `DaliOperatorTable.lookupOperatorOverloads`
resolves the operator functions case-sensitively. If we just add the lowercase operator,
`DaliOperatorTable.lookupOperatorOverloads` can't resolve the uppercase operator, and there will be null pointer exception.

shenodaguirguis: can we change `DaliOperatorTable.lookupOperatorOverloads` resolve operator functions with `isCaseSensitive == false`?

ljfgem: I am not sure if it is safe to do that, it was written 3 years ago and there might be unexpected exceptions if we change it. @funcheetah What is your opinion?

wmoustafa: We should see the correct spec regarding whether those operators are case sensitive or not.

ljfgem: If set `DaliOperatorTable.lookupOperatorOverloads` with `isCaseSensitive = false`, there is a new exception which didn't happen before:
`org.apache.calcite.runtime.CalciteContextException: At line 0, column 0: No match found for function signature <One LinkedIn UDF>`. So I think we should keep it case sensitive.

shenodaguirguis: this new exception suggests there is another place that needs to be made case insensitive.
to @wmoustafa 's point, we should check the specs, which I doubt it would be case sensitive. If things are SQL-like, we should not be case sensitive.

ljfgem: I think we should be case sensitive for UDFs like `com.linkedin.dali.udf.date.hive.EpochToDateFormat`, and be case insensitive for operators like `rlike` & `regexp`.
There is a comment in `StaticHiveFunctionRegistry`:
```
// NOTE: all built-in keyword-based function names should be lowercase for case-insensitive comparison.
// All Dali UDFs should have case sensitive function class names when we do comparison to look up.
```
If we want to follow the rule above, and want to just add the lowercase `regexp` and `rlike` in `StaticHiveFunctionRegistry`, we have 2 options:
1. In `DaliOperatorTable.lookupOperatorOverloads`, if the function is `regexp` or `rlike`, we set `isCaseSensitive=false`
2. We can avoid all the uppercase `REGEXP` or `RLIKE` in `DaliOperatorTable.lookupOperatorOverloads` by modifying their definition class `HiveRLikeOperator`, setting the name to be lowercase:
```
public static final HiveRLikeOperator RLIKE = new HiveRLikeOperator("rlike", false);
public static final HiveRLikeOperator REGEXP = new HiveRLikeOperator("regexp", false);
```

If we want to make all case-insensitive (including UDFs), we just need to modify `StaticHiveFunctionRegistry.addFunctionEntry` to add lowercase keys, and remove all the `isCaseSensitive` parameters.

I think both case-sensitive & case-insensitive for UDFs could work, but I prefer to keep case-sensitive for UDFs.
@funcheetah @wmoustafa @shenoda What's your opinion?

funcheetah: Thanks for the investigation. I prefer to keep case-sensitive for UDFs for now unless we have a good reason to change the semantics. Regarding `regexp` and `rlike`, registering twice and adding proper documentation in `StaticHiveFunctionRegistry ` sounds cleaner to me as `StaticHiveFunctionRegistry` is the central place we keep track of function registry. Special handling in both approach 1 and 2 needs more attention to maintain.

PR #89, opened 2021-05-19
Add timestamp type support in RelDataTypeToAvroType for Coral-Schema

ljfgem:
This patch added the `TIMESTAMP` type support in `RelDataTypeToAvroType` to avoid the exception `java.lang.UnsupportedOperationException: TIMESTAMP is not supported.`

For `TIMESTAMP` RelDataType, we convert it to the following avro schema:
```
{
      "type" : "long",
      "logicalType" : "timestamp"
}
```
, which aligns with our internal handling of avro schema in Spark.

Tests:
1. unit test
2. test on the affected production views, which could be translated well
3. Integration test, no regression after verification

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
@@ -102,6 +102,10 @@ class RelDataTypeToAvroType {
         return Schema.create(Schema.Type.NULL);
       case ANY:
         return Schema.create(Schema.Type.BYTES);
+      case TIMESTAMP:
+        Schema schema = Schema.create(Schema.Type.LONG);
+        schema.addProp("logicalType", "timestamp");
+        return schema;
       default:
         throw new UnsupportedOperationException(relDataType.getSqlTypeName() + " is not supported.");
     }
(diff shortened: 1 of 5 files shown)


coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
```diff
@@ -102,6 +102,10 @@ private static Schema basicSqlTypeToAvroType(BasicSqlType relDataType) {
         return Schema.create(Schema.Type.NULL);
       case ANY:
         return Schema.create(Schema.Type.BYTES);
+      case TIMESTAMP:
+        Schema schema = Schema.create(Schema.Type.LONG);
+        schema.addProp("logicalType", "timestamp");
```
funcheetah, line 107: Is the `milliseconds` information lost during the conversion since base table schema is `"logicalType" : "timestamp-millis"`?

ljfgem: Yes, because in our internal Spark 2.X, if `logicalType == "timestamp"`, Spark will convert the avro type to timestamp. So the logicalType should be `"timestamp"` to be recognized by our internal Spark.

funcheetah (approved):

PR #90, opened 2021-05-20
Make `StaticHiveFunctionRegistry.lookup` case-insensitive for all functions

ljfgem:
Previously, `StaticHiveFunctionRegistry.lookup` treats built-in functions case-insensitively and Dali UDFs case-sensitively, which might bring some issues. For example, `RLIKE` and `rlike` need to be registered twice.

This patch makes `StaticHiveFunctionRegistry.lookup` case-insensitive for all functions including Dali UDFs, because there is very low chance of having two different UDF classes that have the same package and class name but only differ in casing.

Tests:
1. `./gradlew clean build`
2. views affected by exception: `com.linkedin.coral.hive.hive2rel.functions.UnknownSqlFunctionException: Unknown function name: regexp`, which can be translated with this patch


shenodaguirguis (approved):
thanks for this patch!
nit: the PR description says "Previously, StaticHiveFunctionRegistry.lookup treats built-in functions case-insensitively" which is not true, it was treated case-sensitive (contrary to the specs)

wmoustafa (approved):
Great work! Thanks @ljfgem, @shenodaguirguis, and @funcheetah for all the offline discussions.

PR #91, opened 2021-05-24
escape newline for dali row schema string

hanghangliu:
This change is to fix the issue when schema string returned from dali row schema, newline character exists(most seen in docs fields). The unescaped newline("\n") need to be escaped(replace by "\\\\n") in order to be parsed by Schema.Parser().


ljfgem (approved):
Integration test done, no regression after verification.

PR #93, opened 2021-05-27
Support Hive->Trino for LATERAL VIEW EXPLODE over ARRAY of STRUCTS

losipiuk:
This PR adds a wrapper Record type with single column when unnesting an `ARRAY` of `ROW`s. This is not in-line with what Hive's `LATERAL VIEW EXPLODE` does, exposing whole `ROW` (struct) as a single column. Therefore, we add an extra artificial wrapping `ROW` with single field simulates Hive semantics in Trino.

Example transformation:

Given table with an array of structs column:
`CREATE TABLE example_table(id INTEGER, arr array<struct<sa: int, sb: string>>)`

We rewrite view defined as:
`SELECT id, arr_exp FROM example_table LATERAL VIEW EXPLODE(arr) t AS arr_exp`

To:

```
SELECT "$cor0".id AS id, t1.arr_exp AS arr_exp
FROM example_table AS "$cor0"
CROSS JOIN LATERAL (SELECT arr_exp
FROM UNNEST(TRANSFORM("$cor0".arr, x -> ROW(x))) AS t0 (arr_exp)) AS t1
```
The crucial part in above transformation is call to TRANSFORM with lambda which adds extra layer of `ROW` wrapping.

### DIFF
--- a/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/RelToTrinoConverter.java
+++ b/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/RelToTrinoConverter.java
@@ -12,34 +12,26 @@ import java.util.Locale;
 import java.util.Map;
 
 import org.apache.calcite.rel.RelNode;
-import org.apache.calcite.rel.core.Correlate;
-import org.apache.calcite.rel.core.Project;
-import org.apache.calcite.rel.core.TableScan;
-import org.apache.calcite.rel.core.Uncollect;
-import org.apache.calcite.rel.core.Values;
-import org.apache.calcite.rel.core.Window;
+import org.apache.calcite.rel.core.*;
 import org.apache.calcite.rel.rel2sql.RelToSqlConverter;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.rel.type.RelDataTypeFieldImpl;
+import org.apache.calcite.rel.type.RelRecordType;
 import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexNode;
-import org.apache.calcite.sql.JoinConditionType;
-import org.apache.calcite.sql.JoinType;
-import org.apache.calcite.sql.SqlCall;
-import org.apache.calcite.sql.SqlIdentifier;
-import org.apache.calcite.sql.SqlJoin;
-import org.apache.calcite.sql.SqlKind;
-import org.apache.calcite.sql.SqlLiteral;
-import org.apache.calcite.sql.SqlNode;
-import org.apache.calcite.sql.SqlNodeList;
+import org.apache.calcite.sql.*;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.sql.validate.SqlValidatorUtil;
 import org.apache.calcite.util.Util;
 
 import com.linkedin.coral.com.google.common.collect.ImmutableList;
+import com.linkedin.coral.hive.hive2rel.rel.HiveUncollect;
+import com.linkedin.coral.trino.rel2trino.functions.TrinoArrayTransformFunction;
 
-import static com.linkedin.coral.trino.rel2trino.Calcite2TrinoUDFConverter.*;
+import static com.linkedin.coral.trino.rel2trino.Calcite2TrinoUDFConverter.convertRel;
 
 
 public class RelToTrinoConverter extends RelToSqlConverter {
@@ -130,7 +122,44 @@ public class RelToTrinoConverter extends RelToSqlConverter {
     // Build <unnestColumns>
     final List<SqlNode> unnestOperands = new ArrayList<>();
     for (RexNode unnestCol : ((Project) e.getInput()).getChildExps()) {
-      unnestOperands.add(x.qualifiedContext().toSql(null, unnestCol));
+      if (e instanceof HiveUncollect && unnestCol.getType().getSqlTypeName().equals(SqlTypeName.ARRAY)
+          && unnestCol.getType().getComponentType().getSqlTypeName().equals(SqlTypeName.ROW)) {
+
+        // wrapper Record type with single column.
+        // It is needed as Trino follows SQL standard when unnesting
+        // ARRAY of ROWs, exposing each field in a ROW as separate column. This is not in-line with what
+        // Hive's LATERAL VIEW EXPLODE does, exposing whole ROW (struct) as a single column.
+        // Adding extra artificial wrapping ROW with single field simulates Hive semantics in Trino.
+        //
+        // Example transformation:
+        //
+        // Given table with an array of structs column:
+        //   CREATE TABLE example_table(id INTEGER, arr array<struct<sa: int, sb: string>>)
+        // We rewrite view defined as:
+        //  SELECT id, arr_exp FROM example_table LATERAL VIEW EXPLODE(arr) t AS arr_exp
+        // To:
+        //  SELECT "$cor0".id AS id, t1.arr_exp AS arr_exp
+        //    FROM example_table AS "$cor0"
+        //    CROSS JOIN LATERAL (SELECT arr_exp
+        //    FROM UNNEST(TRANSFORM("$cor0".arr, x -> ROW(x))) AS t0 (arr_exp)) AS t1
+        //
+        // The crucial part in abave transformation is call to TRANSFORM with lambda which adds extra layer of
+        // ROW wrapping.
+
+        RelRecordType transformDataType = new RelRecordType(
+            ImmutableList.of(new RelDataTypeFieldImpl("wrapper_field", 0, unnestCol.getType().getComponentType())));
+
+        // wrap unnested field to type defined above using transform(field, x -> ROW(x))
+        TrinoArrayTransformFunction tranformFunction = new TrinoArrayTransformFunction(transformDataType);
+        SqlNode fieldRef = x.qualifiedContext().toSql(null, unnestCol);
+        String fieldRefString = fieldRef.toSqlString(TrinoSqlDialect.INSTANCE).getSql();
+        SqlCharStringLiteral transformArgsLiteral =
+            SqlLiteral.createCharString(String.format("%s, x -> ROW(x)", fieldRefString), POS);
+
+        unnestOperands.add(tranformFunction.createCall(POS, transformArgsLiteral));
+      } else {
+        unnestOperands.add(x.qualifiedContext().toSql(null, unnestCol));
+      }
     }
 
     // Build UNNEST(<unnestColumns>)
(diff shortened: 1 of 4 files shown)


losipiuk:
@findepi PTAL

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/RelToTrinoConverter.java
```diff
@@ -130,7 +122,25 @@ public Result visit(Uncollect e) {
     // Build <unnestColumns>
     final List<SqlNode> unnestOperands = new ArrayList<>();
     for (RexNode unnestCol : ((Project) e.getInput()).getChildExps()) {
-      unnestOperands.add(x.qualifiedContext().toSql(null, unnestCol));
+      if (e instanceof HiveUncollect && unnestCol.getType().getSqlTypeName().equals(SqlTypeName.ARRAY)
+          && unnestCol.getType().getComponentType().getSqlTypeName().equals(SqlTypeName.ROW)) {
+
+        // wrapper Record type with single column
```
findepi, line 128: maybe add also a comment why this is needed

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/RelToTrinoConverter.java
```diff
@@ -130,7 +122,25 @@ public Result visit(Uncollect e) {
...
+        List<RelDataTypeField> fields = new ArrayList();
```
findepi, line 129: perhaps turn this into a `ImmutableList.of()` singleton list

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
```diff
@@ -106,20 +106,28 @@ public void testViews(String database, String view, String expectedSql) {
...
+
         { "test", "current_date_and_timestamp_view", "SELECT CURRENT_TIMESTAMP, TRIM(CAST(CURRENT_TIMESTAMP AS VARCHAR(65535))) AS \"ct\", CURRENT_DATE, CURRENT_DATE AS \"cd\", \"a\"\nFROM \"test\".\"tablea\"" },
 
-        { "test", "lateral_view_json_tuple_view", "SELECT \"$cor4\".\"a\" AS \"a\", \"t0\".\"d\" AS \"d\", \"t0\".\"e\" AS \"e\", \"t0\".\"f\" AS \"f\"\n"
-            + "FROM \"test\".\"tablea\" AS \"$cor4\"\nCROSS JOIN LATERAL (SELECT "
-            + "\"if\"(\"REGEXP_LIKE\"('trino', '^[^\\\"]*$'), CAST(\"json_extract\"(\"$cor4\".\"b\".\"b1\", '$[\"' || 'trino' || '\"]') AS VARCHAR(65535)), NULL) AS \"d\", "
-            + "\"if\"(\"REGEXP_LIKE\"('always', '^[^\\\"]*$'), CAST(\"json_extract\"(\"$cor4\".\"b\".\"b1\", '$[\"' || 'always' || '\"]') AS VARCHAR(65535)), NULL) AS \"e\", "
-            + "\"if\"(\"REGEXP_LIKE\"('rocks', '^[^\\\"]*$'), CAST(\"json_extract\"(\"$cor4\".\"b\".\"b1\", '$[\"' || 'rocks' || '\"]') AS VARCHAR(65535)), NULL) AS \"f\"\n"
+        { "test", "lateral_view_json_tuple_view", "SELECT \"$cor8\".\"a\" AS \"a\", \"t0\".\"d\" AS \"d\", \"t0\".\"e\" AS \"e\", \"t0\".\"f\" AS \"f\"\n"
```
findepi, line 119: we should reset `$cor` counter between tests

(no change requsted here)

findepi (approved):
awesome!

will you please add correctness regression tests in Trino as well?

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
losipiuk: Added a commit

losipiuk:
> awesome!
> 
> will you please add correctness regression tests in Trino as well?

https://github.com/trinodb/trino/pull/8120; passes locally with coral changes.

losipiuk:
@wmoustafa PTAL :)

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
findepi: @losipiuk i suggest separate PR, since we squash-merge here.

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
```diff
@@ -102,24 +102,24 @@ public void testViews(String database, String view, String expectedSql) {
             + "FROM \"test\".\"table_with_string_array\" AS \"$cor0\"\n"
             + "CROSS JOIN LATERAL (SELECT \"c\"\nFROM UNNEST(\"$cor0\".\"b\") AS \"t0\" (\"c\")) AS \"t1\"" },
 
-        { "test", "view_with_outer_explode_string_array", "SELECT \"$cor1\".\"a\" AS \"a\", \"t1\".\"c\" AS \"c\"\n"
-            + "FROM \"test\".\"table_with_string_array\" AS \"$cor1\"\n"
-            + "CROSS JOIN LATERAL (SELECT \"c\"\nFROM UNNEST(\"if\"(\"$cor1\".\"b\" IS NOT NULL AND CARDINALITY(\"$cor1\".\"b\") > 0, \"$cor1\".\"b\", ARRAY[NULL])) AS \"t0\" (\"c\")) AS \"t1\"" },
+        { "test", "view_with_outer_explode_string_array", "SELECT \"$cor0\".\"a\" AS \"a\", \"t1\".\"c\" AS \"c\"\n"
```
findepi, line 105: thank you!

findepi (approved):

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/RelToTrinoConverter.java
```diff
@@ -130,7 +122,28 @@ public Result visit(Uncollect e) {
     // Build <unnestColumns>
     final List<SqlNode> unnestOperands = new ArrayList<>();
     for (RexNode unnestCol : ((Project) e.getInput()).getChildExps()) {
-      unnestOperands.add(x.qualifiedContext().toSql(null, unnestCol));
+      if (e instanceof HiveUncollect && unnestCol.getType().getSqlTypeName().equals(SqlTypeName.ARRAY)
+          && unnestCol.getType().getComponentType().getSqlTypeName().equals(SqlTypeName.ROW)) {
```
wmoustafa, line 126: Could you please add an example of the intended transformation as a comment before the `if` statement? Might as well update the RB description.

losipiuk: Added both

wmoustafa (approved):

losipiuk:
Thanks @wmoustafa :)

ljfgem:
After adding the `transform`, some production views at LinkedIn could not be queried by trino. And we received a ticket complaining it.
i.e.
```
CREATE TABLE test.table_with_struct_array(a int, b array<struct<sa: int, sb: string>>)
CREATE VIEW test.view_with_explode_struct_array AS SELECT a, c.sa FROM test.table_with_struct_array LATERAL VIEW EXPLODE(b) t AS c
```
The translated Trino SQL of view `test.view_with_explode_struct_array` with this PR would be:
```
SELECT "$cor0"."a" AS "a", "t0"."c"."sa" AS "sa"
FROM "test"."table_with_struct_array" AS "$cor0"
CROSS JOIN UNNEST(TRANSFORM("$cor0"."b", x -> ROW(x))) AS "t0" ("c")
```
Since we used `TRANSFORM("$cor0"."b", x -> ROW(x))`, the schema of `t0.c` would be `row(row(sa integer, sb varchar))`, therefore, trino couldn't find `sa` from `t0.c` and will throw the exception `Column 't0.c.sa' cannot be resolved`.

Before this patch, without `TRANSFORM("$cor0"."b", x -> ROW(x))`, the schema of `t0.c` was `row(sa integer, sb varchar)` and `t0.c.sa` could be resolved and the translated trino SQL could run well.

cc: @wmoustafa @funcheetah @losipiuk @findepi

losipiuk:
> Since we used `TRANSFORM("$cor0"."b", x -> ROW(x))`, the schema of `t0.c` would be `row(row(sa integer, sb varchar))`, therefore, trino couldn't find `sa` from `t0.c` and will throw the exception `Column 't0.c.sa' cannot be resolved`.

I do not think this is true. Adding extra ROW wrapping (translating `ARRAY<ROW<...>>` to ``ARRAY<ROW<ROW<...>>`) is on purpose. The reason for that is that SQL standard (which Trino follows) defines that when we do CROSS JOIN UNNEST over `ARRAY<ROW<...` that not only array is untested - but also top level ROW.
This is not what Hive's `LATERAL VIEW EXPLODE` does; this one only unnests array. 

I tested view definition from your example above and it works fine in Trino.

On Hive:
```
hive> create database test;
hive> CREATE TABLE test.table_with_struct_array(a int, b array<struct<sa: int, sb: string>>)
hive> CREATE VIEW test.view_with_explode_struct_array AS SELECT a, c.sa FROM test.table_with_struct_array LATERAL VIEW EXPLODE(b) t AS c
```

On Trino:
```
trino:test> select * from view_with_explode_struct_array;
 a | sa
---+----
(0 rows)

trino:test> prepare x from select * from view_with_explode_struct_array;
PREPARE
trino:test> describe output x;
 Column Name | Catalog | Schema |             Table              |  Type   | Type Size | Aliased
-------------+---------+--------+--------------------------------+---------+-----------+---------
 a           | hive    | test   | view_with_explode_struct_array | integer |         4 | false
 sa          | hive    | test   | view_with_explode_struct_array | integer |         4 | false
(2 rows)


```

findepi:
> After adding the `transform`, some production views at LinkedIn could not be queried by trino.

@ljfgem @wmoustafa what Trino version you observe this on?

What do the following queries return on that version?

```
trino> SELECT * FROM UNNEST(ARRAY[ROW(1, 2)]);
 _col0 | _col1
-------+-------
     1 |     2

trino> SELECT * FROM UNNEST(transform(ARRAY[ROW(1, 2)], x -> ROW(x)));
 _col0
--------
 {1, 2}

trino> SELECT DISTINCT node_version FROM system.runtime.nodes;
 node_version
--------------
 361
```

ljfgem:
Thanks @losipiuk @findepi!
Looks like it is because LinkedIn is using a lower version (352) of trino, which only unnests `ARRAY` for `ARRAY<ROW<...`:
```
trino> SELECT * FROM UNNEST(ARRAY[ROW(1, 2)]);
 _col0  
--------
 {1, 2} 
```
```
trino> SELECT * FROM UNNEST(transform(ARRAY[ROW(1, 2)], x -> ROW(x)));
  _col0   
----------
 {{1, 2}} 
```

cc: @wmoustafa @funcheetah

phd3:
@findepi @ljfgem @wmoustafa I think this is because Linkedin Trino uses a legacy mechanism for unnest-ing, which doesn't remove the struct. But according to [SQL spec](https://github.com/prestodb/presto/issues/8151#issue-232411751), that isn't right and hence correctly changed in Trino OSS. So #144 shouldn't be merged, as it'd re-introduces the issue. We need to find a solution internally.

findepi:
For reference

- https://github.com/trinodb/trino/commit/f3f7aa3f20f56c60268d445321dea2ad85625913 "Unnest array of row to multiple columns" in v 0.207
  - related issue https://github.com/prestodb/presto/issues/8151 (mentioned already by Pratham)
- https://github.com/trinodb/trino/commit/ec2fb2ff55d848cb9d4619b5978c8c1478335393 "Remove deprecated unnest behavior" in v 308

PR #94, opened 2021-05-28
Avoid throwing exception if fails to get columns using deserializer

ljfgem:
If fails to get the deserializer, there is an exception:
```
java.lang.RuntimeException: MetaException(message:java.lang.ClassNotFoundException Class com.linkedin.opal.hadoop.OpalSerde not found)
```
This patch uses `sd.getCols()` to avoid throwing exception.

Tests:
1. tested on all the affected production views, which could be translated well with this patch
2. integration test, no regression


shenodaguirguis (approved):
LGTM

funcheetah (approved):

PR #95, opened 2021-05-28
 Add alias if select field from struct column

ljfgem:
If select a field from a struct column, like 
```
SELECT table.struct_column.field_name FROM table
``` 
we need to add an alias to make it as:
```
SELECT table.struct_column.field_name AS field_name FROM table
```
, which can avoid null pointer exception:
```
java.lang.NullPointerException
        at org.apache.calcite.sql.validate.SqlValidatorUtil.getTypeEntry(SqlValidatorUtil.java:978)
        at org.apache.calcite.prepare.CalciteCatalogReader.getNamedType(CalciteCatalogReader.java:174)
        at org.apache.calcite.sql.validate.SqlValidatorImpl.getValidatedNodeTypeIfKnown(SqlValidatorImpl.java:1609)
        ...
```

Tests:
1. unit test
2. test on affected production views by the above exception, they could be translated well with this patch
3. integration test on all the production views, no regression


coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilder.java
```diff
@@ -529,7 +530,19 @@ private SqlNode visitFunctionInternal(ASTNode node, ParseContext ctx, SqlLiteral
   protected SqlNode visitSelectExpr(ASTNode node, ParseContext ctx) {
     List<SqlNode> sqlNodes = visitChildren(node, ctx);
     if (sqlNodes.size() == 1) {
-      return sqlNodes.get(0);
+      final SqlNode sqlNode = sqlNodes.get(0);
+      // if select a field from a struct column, like
+      // SELECT table.struct_column.field_name FROM table
+      // we need to add an alias to avoid null pointer exception:
```
shenodaguirguis, line 536: can you clarify in the comment how a null pointer is thrown if no alias is added?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilder.java
```diff
@@ -529,7 +530,19 @@ private SqlNode visitFunctionInternal(ASTNode node, ParseContext ctx, SqlLiteral
...
+      // SELECT table.struct_column.field_name AS field_name FROM table
+      if (sqlNode.getKind() == SqlKind.IDENTIFIER) {
+        List<String> names = ((SqlIdentifier) sqlNode).names;
+        final int size = names.size();
+        if (size > 2) {
```
shenodaguirguis, line 541: a) what is the deal with  `table.struct_column.*`, does this need an alias as well? 
b) is there a better/more accurate way to identify struct columns? What if we have `user.table.non_struct_col` or `struct_column.field` ?  (we would need to add these cases to unit tests)

shenodaguirguis (commented):
I am not clear how/where the NPE is thrown, but it seems it would be better to handle the NPE where it occurs for this case (struct.field) rather than tweaking the input SQL (given the fact that it is a valid SQL).

ljfgem:
> I am not clear how/where the NPE is thrown, but it seems it would be better to handle the NPE where it occurs for this case (struct.field) rather than tweaking the input SQL (given the fact that it is a valid SQL).

Thank you @shenodaguirguis, the root cause is in calcite, and I opened a PR for that: https://github.com/linkedin/linkedin-calcite/pull/39

funcheetah:
If the calcite [PR](https://github.com/linkedin/linkedin-calcite/pull/39) address the issue, could we close this one? @ljfgem

PR #99, opened 2021-06-13
Parse map(k1,v1,k2,v2) from hive to trino

JiajunBernoulli:
Hi, everybody! 
  I'm interested in converting SQL from hive to trino. This is my first PR.  Here's a little bit of my work:

For hive, `SELECT MAP(k1, v1, k2, v2)` is

> {k1=v1, k2=v2};

but for trino, sql should be `SELECT MAP(ARRAY[k1,k2], ARRAY[v1,v2])`

Welcome your suggestions.

### DIFF
--- a/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
+++ b/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
@@ -6,6 +6,7 @@
 package com.linkedin.coral.trino.rel2trino;
 
 import java.math.BigDecimal;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 
@@ -34,6 +35,8 @@ import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexShuttle;
 import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlOperator;
+import org.apache.calcite.sql.fun.SqlMapValueConstructor;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.type.SqlTypeFamily;
 
 import com.linkedin.coral.com.google.common.collect.ImmutableList;
@@ -172,6 +175,10 @@ public class Calcite2TrinoUDFConverter {
       if (call.getOperator() instanceof GenericProjectFunction) {
         return GenericProjectToTrinoConverter.convertGenericProject(rexBuilder, call);
       }
+      // MAP(k1, v1, k2, v2) should be MAP(ARRAY(k1, k2), ARRAY(v1, v2))
+      if (call.getOperator() instanceof SqlMapValueConstructor) {
+        return this.convertMapValueConstructor(rexBuilder, call);
+      }
 
       if (call.getOperator().getName().equals("from_utc_timestamp")) {
         Optional<RexNode> modifiedCall = visitFromUtcTimestampCall(call);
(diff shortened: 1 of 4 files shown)


coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/TrinoSqlDialect.java
```diff
@@ -62,4 +74,26 @@ public void unparseIdentifier(SqlWriter writer, SqlIdentifier identifier) {
...
+    for (int i = 1; i < call.operandCount(); i += 2) {
+      call.operand(i).unparse(writer, leftPrec, rightPrec);
+      // Last comma is redundant
+      if (i < call.operandCount() - 1) {
+        writer.sep(",");
+      }
+    }
+    writer.keyword("]");
+    writer.endList(frame);
+  }
```
wmoustafa, line 98: Should not we handle the transformation in `Calcite2TrinoUDFConverter`? I think if the operands need transformation, this will not take care of them.

JiajunBernoulli: Yes, I understand. 
So far, it seems operands  number must be determined because  `CalciteTrinoUDFMap.getUDFTransformer(String calciteOpName, int numOperands)`. However `map` is variable length parameter, I can't put it in `UDF_MAP`.
Do we plan to support functions with variable length parameters? I thinks this is meaningful topic.

JiajunBernoulli:
Because I can't put `map` in `UDF_MAP`, I can only handle the transformation in `Calcite2TrinoUDFConverter.convertMapValueConstructor` and replace "[]" with "()" in `TrinoSqlDialect.unparseMapValueConstructor`.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/TrinoSqlDialect.java
wmoustafa: Agreed. Maybe we can introduce `getUDFTransformer(String calciteOpName)` that means the `Op` takes variable length parameters?

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
```diff
@@ -194,5 +203,23 @@ private RexCall adjustInconsistentTypesToEqualityOperator(RexCall call) {
...
+      // Even numbers are keys
+      List<RexNode> keys = new ArrayList<>();
+      for (int i = 0; i < sourceOperands.size(); i += 2) {
+        keys.add(sourceOperands.get(i));
+      }
+      results.add(rexBuilder.makeCall(SqlStdOperatorTable.ARRAY_VALUE_CONSTRUCTOR, keys));
+      // Odd numbers are values
+      List<RexNode> values = new ArrayList<>();
+      for (int i = 1; i < sourceOperands.size(); i += 2) {
+        values.add(sourceOperands.get(i));
```
ljfgem, line 219: If value is also a map, we failed to convert it.
For example, if the view definition is:
```
SELECT MAP('key1', MAP('a', 'b', 'c', 'd'), 'key2', MAP('a', 'b', 'c', 'd')) AS col_map
```
it would only be converted to
```
SELECT MAP (ARRAY['key1', 'key2'], ARRAY[MAP ('a', 'b', 'c', 'd'), MAP ('a', 'b', 'c', 'd')]) AS "col_map"
```
However, the result should be
```
SELECT MAP (ARRAY['key1', 'key2'], ARRAY[MAP (ARRAY['a', 'c'], ARRAY['b', 'd']), MAP (ARRAY['a', 'c'], ARRAY['b', 'd'])]) AS "col_map"
```
I think we can change this part to be:
```
if (sourceOperands.get(i) instanceof RexCall) {
  values.add(visitCall((RexCall) sourceOperands.get(i)));
} else {
  values.add(sourceOperands.get(i));
}
```
and add corresponding unit tests.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/TrinoSqlDialect.java
```diff
@@ -62,4 +74,14 @@ public void unparseIdentifier(SqlWriter writer, SqlIdentifier identifier) {
   public boolean requireCastOnString() {
     return true;
   }
+
+  private void unparseMapValueConstructor(SqlWriter writer, SqlCall call, int leftPrec, int rightPrec) {
+    writer.keyword(call.getOperator().getName()); // "MAP"
+    final SqlWriter.Frame frame = writer.startList("(", ")"); // not is "[" and "]"
```
ljfgem, line 80: nit: // not "[" and "]"

ljfgem (changes requested):
Thank you for your PR! I think we need to handle the case `map(key, map(...))`

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
JiajunBernoulli: I agree with you and will try to fix it.

JiajunBernoulli:
I updated this PR, Do you have any suggestions?

wmoustafa (approved):

wmoustafa:
@ljfgem PTAL.

ljfgem:
> @ljfgem PTAL.

Thank you for reminding, integration test is running, will merge if there is no regression.

PR #100, opened 2021-06-14
Manually registered some UDFs

ljfgem:
Based on the test result of latest 3 month production views, these UDFs need to be registered.
Tested on the affected views, which could be translated with this patch.

### DIFF
--- a/coral-hive/build.gradle
+++ b/coral-hive/build.gradle
@@ -3,7 +3,7 @@ apply plugin: 'antlr'
 
 dependencies {
   antlr deps.'antlr'
-  compile('com.linkedin.calcite:calcite-core:1.21.0.146') {
+  compile('com.linkedin.calcite:calcite-core:1.21.0.148') {
     artifact {
       name = 'calcite-core'
       extension = 'jar'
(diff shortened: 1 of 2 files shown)


wmoustafa:
Could you please add comments as of why dynamic UDF resolution does not work with those functions? Are there things that we can fix there?

ljfgem:
Reasons of the exceptions:
`Unable to find org.apache.hadoop.hive.ql.udf.generic.GenericUDF.initialize()`: the UDFs are not generic UDF, cannot find `initialize` method.

`Unable to call org.apache.hadoop.hive.ql.udf.generic.GenericUDF.initialize()`: there are many different reasons of this exception, the following 2 are reasons of affected views from latest 3 month views.
1. Argument type is union type, we converted it to struct, so there is a mismatch between the provided type and required type
2. Exception: `java.lang.NoSuchMethodError: com.google.common.collect.ImmutableSet.toImmutableSet()Ljava/util/stream/Collector;` while calling the `initialize` method, seems to be because the versions of `guava` are different.

Since only 3 views are affected by the above UDF exceptions, I don't think we should spend time fixing for them at present, as there are some more important issues to be fixed.

funcheetah (approved):

shenodaguirguis:
since this is a temporary fix, can we please create issue here and corresponding one in internally (in the team's backlog) to detail the observations, the temporary fix and if any hints for the long term/permanent fix (& link to this PR)?

PR #110, opened 2021-07-14
Make Coral-Schema return correct schema while selecting field from a nested struct

ljfgem:
Previously, `RelToAvroSchemaConverter.visitFieldAccess` just handles the case `SELECT struct_col.int_field`, if the struct is nested in map/array/another struct, coral-schema fails to return the correct schema. This patch adds the support for these cases.

Tests:
1. unit test
2. tested on affected production views, could return the correct schema
3. integration test, no regression

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
@@ -460,7 +460,7 @@ public class RelToAvroSchemaConverter {
           fieldNames.push(((RexFieldAccess) referenceExpr).getField().getName());
           referenceExpr = ((RexFieldAccess) referenceExpr).getReferenceExpr();
         } else {
-          return super.visitFieldAccess(rexFieldAccess);
+          throw new IllegalArgumentException("Unsupported referenceExpr kind: " + referenceExpr.getKind());
         }
       }
       String oldFieldName = rexFieldAccess.getField().getName();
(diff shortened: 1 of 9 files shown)


coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
```diff
@@ -450,31 +450,63 @@ public RexNode visitRangeRef(RexRangeRef rexRangeRef) {
 
     @Override
     public RexNode visitFieldAccess(RexFieldAccess rexFieldAccess) {
```
shenodaguirguis, line 452: just a suggestion for code readability: this method is now long with several looped and nested if-else, which makes it a bit hard to read/follow the logic (thanks for the comments!). If, on the other hand we extract each if/else body and each loop body into a separate method with expressive method name, it will be much easier to read.

coral-schema/src/test/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverterTests.java
```diff
@@ -612,6 +612,19 @@ public void testSelectStarFromNestComplex() {
         TestUtils.loadSchema("testSelectStarFromNestComplex-expected.avsc"));
   }
 
+  @Test
+  public void testSelectNestedStructFieldFromNestComplex() {
```
shenodaguirguis, line 616: can we add tests for deep nested fields (e.g., `struct.array_col`, `sturct.innter_struct.map_col`, ...)?

shenodaguirguis (commented):
LGTM!
just 2 minor comments

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
ljfgem: Thank you for suggestion, I extracted the second `while` loop. It is kind of hard to extract the first `while` loop because it needs to return `super.visitFieldAccess(rexFieldAccess);` in an if-else branch.

funcheetah (approved):
Thanks @ljfgem for adding this patch. It looks good.

PR #111, opened 2021-07-20
Use default value for HIVE_QUOTEDID_SUPPORT when hiveConf=null

nagarathnam200:
This patch aims to prevent the NullPointerException we get when the hiveConf is Null. This patch is similar to 
https://issues.apache.org/jira/secure/attachment/12788767/HIVE-13101.1.patch


ljfgem (approved):
LGTM, no regression in integration test

PR #113, opened 2021-07-21
Add warn log to `useSQL11ReservedKeywordsForIdentifier`

ljfgem:
Add warn log if there is an error in `useSQL11ReservedKeywordsForIdentifier`


shenodaguirguis (approved):
thanks!

PR #114, opened 2021-07-27
Support `LATERAL VIEW` with other UDTFs besides `UNNEST`

ljfgem:
This patch adds the support for UDTF which extends Hive GenericUDTF.

For generic UDTFs, we treat them as LinkedIn UDFs and make the following conversion in `ParseTreeBuilder`
```
SELECT a, t.col1
FROM test.tableOne
LATERAL VIEW `com.linkedin.coral.hive.hive2rel.CoralTestUDTF`(`tableone`.`a`) `t`
->
SELECT a, t.col1
FROM test.tableOne
LATERAL COLLECTION_TABLE(`com.linkedin.coral.hive.hive2rel.CoralTestUDTF`(`tableone`.`a`)) AS `t` (`col1`)
```
Tests:
1. Unit tests
2. Tested on affected production views, which could be translated well

### DIFF
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
@@ -381,6 +381,34 @@ public class StaticHiveFunctionRegistry implements HiveFunctionRegistry {
     createAddUserDefinedFunction("com.linkedin.etg.business.common.udfs.MapSfdcProductCode", HiveReturnTypes.STRING,
         STRING);
     createAddUserDefinedFunction("com.linkedin.etg.business.common.udfs.MapSfdcProductId", ReturnTypes.INTEGER, STRING);
+    createAddUserDefinedFunction("udfs.SeoReferrerTrkUdf", HiveReturnTypes.STRING, STRING_STRING_STRING);
+    createAddUserDefinedFunction("com.linkedin.vector.daliview.udf.PresentMediaType", HiveReturnTypes.STRING,
+        family(SqlTypeFamily.ANY));
+    createAddUserDefinedFunction("com.linkedin.tscp.reporting.dali.udfs.AdClickClassifier",
+        HiveReturnTypes.rowOf(ImmutableList.of("clicks", "landingPageClicks", "totalEngagements", "otherEngagements",
+            "likes", "commentLikes", "comments", "shares", "follows", "oneClickLeadFormOpens", "companyPageClicks",
+            "fullScreenPlays", "viralClicks", "viralLandingPageClicks", "viralLikes", "viralCommentLikes",
+            "viralComments", "viralShares", "viralFollows", "viralOneClickLeadFormOpens", "viralCompanyPageClicks",
+            "viralFullScreenPlays", "viralTotalEngagements", "viralOtherEngagements", "adUnitClicks", "actionClicks",
+            "textUrlClicks", "opens", "cardClicks", "viralCardClicks", "costInUsd", "costInLocalCurrency"),
+            ImmutableList.of(SqlTypeName.INTEGER, SqlTypeName.INTEGER, SqlTypeName.INTEGER, SqlTypeName.INTEGER,
+                SqlTypeName.INTEGER, SqlTypeName.INTEGER, SqlTypeName.INTEGER, SqlTypeName.INTEGER, SqlTypeName.INTEGER,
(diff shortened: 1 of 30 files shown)


coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
```diff
@@ -501,6 +518,31 @@ public static void createAddUserDefinedFunction(String functionName, SqlReturnTy
...
 
+  /**
+   * Adds the generic UDTF, which is almost same as how we register for LinkedIn UDFs except that we need to register
+   * the return field names in `UDTF_RETURN_FIELD_NAME_MAP`
+   */
+  public static void createAddUserDefinedTableFunction(String functionName, ImmutableList<String> returnFieldNames,
+      ImmutableList<?> returnFieldTypes, SqlOperandTypeChecker operandTypeChecker) {
+    Preconditions.checkArgument(!returnFieldTypes.isEmpty() && returnFieldTypes.size() == returnFieldNames.size()
+        && (returnFieldTypes.stream().allMatch(type -> type instanceof SqlTypeName)
+            || returnFieldTypes.stream().allMatch(type -> type instanceof SqlReturnTypeInference)));
```
shenodaguirguis, line 529: just curious, why restricting return types to be either all `SqlTypeName` or `SqlReturnTypeInference` as opposed to have a mix of both?

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/CoralTestUDTF.java
```diff
@@ -0,0 +1,50 @@
...
+
+// This is used in TestUtils to set up as dali function
+// This needs in a separate file for Hive to correctly load for setup
+
+
+/**
+ * CoralTestUDTF outputs the number of rows seen, twice. It's output twice
+ * to test outputting of rows on close with lateral view.
+ */
+public class CoralTestUDTF extends GenericUDTF {
```
funcheetah, line 25: Not specific to this PR. Do you think if there is way to reuse the UDF/UDTF class in other modules instead of duplicating the code?

shenodaguirguis (approved):
LGTM in general, left minor comments

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/CoralTestUDTF.java
```diff
@@ -0,0 +1,50 @@
...
+
+  Integer count = 0;
+  Object[] forwardObj = new Object[1];
+
+  @Override
+  public void close() throws HiveException {
+    forwardObj[0] = count;
+    forward(forwardObj);
+    forward(forwardObj);
```
funcheetah, line 34: Why call `forward` twice here?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
ljfgem: Because we need to call either `HiveReturnTypes.rowOf(ImmutableList<String> fieldNames, ImmutableList<SqlTypeName> types)` or `HiveReturnTypes.rowOfInference(ImmutableList<String> fieldNames, ImmutableList<SqlReturnTypeInference> types)` to get the return type.

We need 2 methods because `HiveReturnTypes.rowOfInference` is useful to create row types whose fields are complex types and hence cannot be represented through `SqlTypeName` to be used in `HiveReturnTypes.rowOf`

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/CoralTestUDTF.java
ljfgem: At present, I have no idea about how to reuse these classes since the tests are independent, and there will be exceptions if we don't create the UDF classes for each module's tests.

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/CoralTestUDTF.java
ljfgem: As described in the comment of CoralTestUDTF:
```
CoralTestUDTF outputs the number of rows seen, twice. It's output twice
to test outputting of rows on close with lateral view.
```
There is no special reason, and the time doesn't matter as we just want to create a new table using table function.

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/CoralTestUDTF.java
funcheetah: Fair.

funcheetah (approved):
LGTM. Thanks @ljfgem !

PR #115, opened 2021-07-28
Passing docstring from underlying sources to views.

varunbharill:
Code changes to support passing down docstring from underlying tables to views created on top of them. 
6 UTs added covering different scenarios. 
Some special cases like literals, etc. still need to be handled.


ljfgem (approved):
LGTM, thank you @varunbharill!

funcheetah (commented):
Let's run regression tests once before check in

funcheetah (approved):
Thanks @varunbharill for this patch!

PR #117, opened 2021-08-05
Delete incorrect code

huangxiaopingRD:
This code will be converted to the wrong UDF, resulting in the final SQL execution failure.It needs to be deleted.


coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/CalciteTrinoUDFMap.java
```diff
@@ -65,11 +65,6 @@ private CalciteTrinoUDFMap() {
     createRuntimeUDFMapEntry(UDF_MAP, hiveToCalciteOp("decode"), 2,
         "[{\"regex\":\"(?i)('utf-8')\", \"input\":2, \"name\":\"from_utf8\"}]", "[{\"input\":1}]", null);
 
-    // FIXME: this is incorrect. Adding this to test correctness of the overall system
```
huangxiaopingRD, line 68: This code will be converted to the wrong UDF, resulting in the final SQL execution failure. It needs to be deleted.

wmoustafa (approved):

PR #128, opened 2021-08-20
Return DATE type in Hive date() function

ebyhr:
I'm not sure about old version's and other distribution's behavior, but it returns `DATE` type in HDP 3.1. 

```sql
> CREATE VIEW view_with_date AS SELECT date('2021-08-20') as a;

> DESC view_with_date;

| a         | date       |          |
```


ebyhr:
cc: @ljfgem

ljfgem (approved):
Thank you @ebyhr, the return type should be `DATE`.

PR #134, opened 2021-08-23
coral-hive: Added supported for lateral view explode without column aliases for MAP type

uzshao:
Queries like

`SELECT key, value FROM (SELECT MAP('key1', 'value1') as m) tmp LATERAL VIEW EXPLODE(m) m_alias`

can run successfully on Hive because the default column names `key` and `value` are automatically given.

This diff makes it possible to process such queries by adding the column aliases automatically.

Unfortunately it's not possible to handle the other case where the type is ARRAY, and the default column name is `col` since we don't have metadata information at this stage (ASTNode Tree).

cc @ljfgem


ljfgem:
Thank you @uzshao for this patch!
Since we cannot recognize the type (map or array), I think it is not necessary to support only MAP for now, because it would throw an exception if we use `lateral view ARRAY` without alias. Actually, `lateral view ARRAY` is used much more than `lateral view MAP` in our production views. 
Therefore, I think we can check it in if we can resolve the `TODO: handle the case when op1 is array type, in which case we should add 'col' automatically` item.

uzshao:
Makes sense.

Any thoughts on how to solve it for both `lateral view ARRAY` and `lateral view MAP`?  I would dig into the Calcite SqlTree validation process to see what is customizable there.  Please let me know if you have a simpler idea.

uzshao:
This is now superseded by #137

PR #135, opened 2021-08-24
Avoid duplicating visit(SqlCall) FuzzyUnionSqlRewriter

autumnust:
The edited code block in `FuzzyUnionSqlRewriter.java` is exactly the same as `super.visit(call)` as I read through the code base. We should directly invoke the method in super class instead of copying them. 

Also make `getFuzzyUnionView` method in test class in parity with `com.linkedin.coral.hive.hive2rel.HiveToRelConverter#convertView` 

Since there's no actual change in the functionality, there's no additional tests done other than existing unit tests.


ljfgem (approved):
Thanks @autumnust !
Integration test succeeded.

PR #137, opened 2021-08-28
Improved calcite hive lateral view explode compatibility

uzshao:
This patch fixes two problems:
1. While `LATERAL VIEW EXPLODE(my_array) t AS col`, `LATERAL VIEW EXPLODE(my_map) t AS key, value` are supported, `LATERAL VIEW EXPLODE(my_array_or_map) t` was not supported until this patch;
2. coral-trino generates queries like `FROM t CROSS JOIN LATERAL (SELECT xxx FROM UNNEST(t.arr_col) AS t2(c2))` which is not supported by the version of Trino that I am using.  With this patch, it generates `FROM t CROSS JOIN UNNEST(t.arr_col) AS t2(c2)` which is well supported by Presto as noted [in trino official doc](https://trino.io/docs/current/sql/select.html#unnest)

### DIFF
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveRelBuilder.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveRelBuilder.java
@@ -1,6 +1,15 @@
+/**
+ * Copyright 2021 LinkedIn Corporation. All rights reserved.
+ * Licensed under the BSD-2 Clause license.
+ * See LICENSE in the project root for license information.
+ */
 package com.linkedin.coral.hive.hive2rel;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import com.google.common.base.Preconditions;
+
 import org.apache.calcite.plan.Context;
 import org.apache.calcite.plan.Contexts;
 import org.apache.calcite.plan.RelOptCluster;
@@ -14,10 +23,8 @@ import org.apache.calcite.tools.Frameworks;
 import org.apache.calcite.tools.RelBuilder;
 import org.apache.calcite.tools.RelBuilderFactory;
 import org.apache.calcite.util.Pair;
-import com.linkedin.coral.hive.hive2rel.rel.HiveUncollect;
 
-import java.util.ArrayList;
-import java.util.List;
+import com.linkedin.coral.hive.hive2rel.rel.HiveUncollect;
 
 import static org.apache.calcite.rel.core.RelFactories.DEFAULT_AGGREGATE_FACTORY;
 import static org.apache.calcite.rel.core.RelFactories.DEFAULT_EXCHANGE_FACTORY;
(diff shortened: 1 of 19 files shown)


uzshao:
@ljfgem You are the expert in this part of the code.  Can you help review this?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilder.java
```diff
@@ -292,22 +298,16 @@ private SqlNode visitLateralViewExplode(List<SqlNode> sqlNodes, List<SqlNode> al
...
-    aliasCallOperands.addAll(aliasOperands.subList(1, operandCount));
-    SqlCall aliasCall = SqlStdOperatorTable.AS.createCall(ZERO, aliasCallOperands);
+    unnestCall = HiveExplodeOperator.EXPLODE.createCall(ZERO, unnestOperand);
+
+    List<SqlNode> asOperands = new ArrayList<>();
+    asOperands.add(unnestCall);
+    asOperands.addAll(aliasOperands.subList(1, aliasOperands.size()));
+    SqlNode as = SqlStdOperatorTable.AS.createCall(ZERO, asOperands);
+
+    SqlNode lateralCall = SqlStdOperatorTable.LATERAL.createCall(ZERO, as);
```
wmoustafa, line 308: Could you add a comment to the code explaining how this generically handles all cases of array, map, and array/map with auto derived aliases? (just by explaining the mapping of the operators created here to `op0... op3` and their meaning in different cases)

wmoustafa (commented):
Thanks @uzshao  for this patch. I have added a few comments and will continue later. @findepi and @losipiuk FYI.

coral-spark/src/main/java/com/linkedin/coral/spark/SparkRelToSparkSqlConverter.java
```diff
@@ -139,7 +145,21 @@ public Result visit(Uncollect e) {
...
+    // SqlOperator with HiveExplodeOperator (or its parent SqlUnnestOperator) is not allowed (only
+    // SqlSetOperator, SqlStdOperatorTable.AS, and SqlStdOperatorTable.VALUES are allowed).
+    /** See {@link org.apache.calcite.rel.rel2sql.SqlImplementor#wrapSelect(SqlNode)}*/
+
+    final List<SqlNode> asOperands = createAsFullOperands(e.getRowType(), unnestNode, x.neededAlias);
+    final SqlNode asNode = SqlStdOperatorTable.AS.createCall(POS, asOperands);
+
+    return result(asNode, ImmutableList.of(Clause.FROM), e, null);
+
+    // return result(unnestNode, ImmutableList.of(Clause.FROM), e, null);
```
ljfgem, line 162: Remove it?

coral-spark/src/test/java/com/linkedin/coral/spark/CoralSparkTest.java
```diff
@@ -191,8 +191,8 @@ public void testNoUdf() {
   public void testLateralView() {
     RelNode relNode = TestUtils.toRelNode(
         String.join("\n", "", "SELECT a, t.ccol", "FROM complex", "LATERAL VIEW explode(complex.c) t as ccol"));
-    String targetSql = String.join("\n", "SELECT complex.a, t1.ccol",
-        "FROM default.complex LATERAL VIEW EXPLODE(complex.c) t1 AS ccol");
+    String targetSql =
+        "SELECT complex.a, t00.ccol\n" + "FROM default.complex LATERAL VIEW EXPLODE(complex.c) t00 AS ccol";
```
ljfgem, line 195: Just for my understanding, could you explain why `t1` will become `t00`?

coral-spark/src/test/java/com/linkedin/coral/spark/CoralSparkTest.java
```diff
@@ -294,22 +294,23 @@ public void testNamedStructViewWithSelectSupported() {
   }
 
   /**
-   * Following Queries are not supported
+   * Following Queries are supported now
```
ljfgem, line 297: Remove this comment?

ljfgem (commented):
Thanks @uzshao for this patch, will continue later.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilder.java
```diff
@@ -265,9 +267,13 @@ private SqlNode visitLateralViewUDTF(List<SqlNode> sqlNodes, List<SqlNode> alias
...
-    // array if operandCount == 3: LATERAL VIEW EXPLODE(op0) AS op1(op2)
-    // map if operandCount == 4: LATERAL VIEW EXPLODE(op0) AS op1(op2, op3)
-    checkState(operandCount == 3 || operandCount == 4,
+    // array if operandCount == 3: LATERAL VIEW EXPLODE(op0) op1 AS op2
+    // map if operandCount == 4: LATERAL VIEW EXPLODE(op0) op1 AS op2, op3
+    // if operandCount == 2: LATERAL VIEW EXPLODE(op0) op1
+    //   if op0 is a map, it implies "AS key, value" where key, value are auto-named by Hive
+    //   if op0 is an array, is implies "AS col" where col is auto-named by Hive
+    //   The logic above will be implemented as part of Calcite SqlNode validation
+    checkState(operandCount == 2 || operandCount == 3 || operandCount == 4,
```
ljfgem, line 276: For the `operandCount == 2` case, I think we also need to add the support if `isOuter == true`.
i.e. 
```
SELECT a_alias.col FROM (SELECT ARRAY('a1', 'a2') as a) tmp LATERAL VIEW OUTER EXPLODE(a) a_alias
```

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveRelBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+        statement) -> new HiveRelBuilder(config.getContext(), cluster, relOptSchema));
+  }
+
+  /** Creates a {@link RelBuilderFactory}, a partially-created RelBuilder.
+   * Just add a {@link RelOptCluster} and a {@link RelOptSchema} */
+  public static RelBuilderFactory proto(final Context context) {
+    return (cluster, schema) -> new HiveRelBuilder(context, cluster, schema);
+  }
+
+  public static final RelBuilderFactory LOGICAL_BUILDER =
```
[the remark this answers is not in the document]
uzshao: The only difference is that this static function calls the static function `HiveRelBuilder.proto` while the one in the parent calls `RelBuilder.proto`.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveRelBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+      HiveRelBuilder.proto(Contexts.of(DEFAULT_PROJECT_FACTORY, DEFAULT_FILTER_FACTORY, DEFAULT_JOIN_FACTORY,
+          DEFAULT_SORT_FACTORY, DEFAULT_EXCHANGE_FACTORY, DEFAULT_SORT_EXCHANGE_FACTORY, DEFAULT_AGGREGATE_FACTORY,
+          DEFAULT_MATCH_FACTORY, DEFAULT_SET_OP_FACTORY, DEFAULT_VALUES_FACTORY, DEFAULT_TABLE_SCAN_FACTORY,
+          DEFAULT_SNAPSHOT_FACTORY, DEFAULT_SPOOL_FACTORY, DEFAULT_REPEAT_UNION_FACTORY));
```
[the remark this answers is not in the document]
uzshao: I copied this from the parent class so they supposed to be the same.  Did you mean that the `LOGICAL_BUILDER` in the parent class have those two additional items while mine don't have them?

coral-schema/src/test/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverterTests.java
```diff
@@ -423,8 +423,8 @@ public void testLateralViewArrayWithComplexType() {
 
   @Test
   public void testLateralViewMap() {
-    String viewSql = "CREATE VIEW v AS " + "SELECT bl.Id AS Id_View_Col, t.Col1, t.Col2 " + "FROM baselateralview bl "
-        + "LATERAL VIEW explode(bl.Map_Col_String) t as Col1, Col2";
+    String viewSql = "CREATE VIEW v AS " + "SELECT bl.Id AS Id_View_Col, t.col1, t.col2 " + "FROM baselateralview bl "
+        + "LATERAL VIEW explode(bl.Map_Col_String) t as col1, col2";
```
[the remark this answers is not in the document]
uzshao: The existing test case has a bug (I think) that in the Avro schema, the column names are `Id_View_Col`, `col1`, `col2`: the first one the same as the query, the latter two somehow lower-cased.  The existing code somehow would produce this inconsistent cases for different parameters, until my patch when the code generates Avro schema with exactly the cases as in the query.

So I have to either change the query or change the Avro schema to make the test pass.  I decide to change the query since that's easier.

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/HiveToRelConverterTest.java
```diff
@@ -189,6 +189,59 @@ public void testWindowWithNoPartitionNoOrder() {
     assertEquals(relString, expected);
   }
 
+  @Test
+  public void testLateralViewArray() {
+    // Test if the code can handle the use of window functions
```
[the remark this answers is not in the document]
uzshao: Nice catch.  Will fix.

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilderTest.java
```diff
@@ -124,10 +125,30 @@ public static void beforeClass() throws HiveException, IOException, MetaExceptio
   }
 
   @DataProvider(name = "validateSql")
-  public Object[][] getValidateSql() {
-    return new Object[][] {
+  public Iterator<Object[]> getValidateSql() {
+    List<List<String>> convertAndValidateSql = ImmutableList.of(
+        // test lateral view explode with a map
```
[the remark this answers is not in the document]
uzshao: Will do

coral-spark/src/main/java/com/linkedin/coral/spark/SparkRelToSparkSqlConverter.java
uzshao: Will do

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilder.java
uzshao: Added in the next commit.

coral-spark/src/test/java/com/linkedin/coral/spark/CoralSparkTest.java
uzshao: It's related to the table alias generation code.  In the earlier code, the aliases that got generated were `t0` and `t1` since they are both based on the basename `t`.  In the new code, the first table alias generated `t0` is used to generate the second table alias so an extra `0` is added to the name of the table.

The main reason for this change is the simplification of the use of `SELECT` and `AS` nodes in the handling of Explode function.  The old code was doing: `AS`, `SELECT`, `AS`, while the new code use a single `AS`.

coral-spark/src/test/java/com/linkedin/coral/spark/CoralSparkTest.java
uzshao: Will remove in the next commit

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilder.java
uzshao: You are right that `operandCount == 2` `OUTER` case is not supported right now.

I couldn't think about a way to support it easily since we don't know the type of the parameter at this point.

Let me throw an exception and add a test case to catch this.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveRelBuilder.java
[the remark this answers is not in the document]
wmoustafa: Thanks! Looking at the code again, this `LOGICAL_BUILDER` is meant to replace `RelFactories.LOGICAL_BUILDER`. That makes sense.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveRelBuilder.java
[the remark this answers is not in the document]
wmoustafa: I meant `RelBuilder` uses them. But now that I see the definition of `RelFactories.LOGICAL_BUILDER`, I can see they match. Interesting that `RelFactories.LOGICAL_BUILDER` does not use `DEFAULT_TABLE_FUNCTION_SCAN_FACTORY` and `DEFAULT_CORRELATE_FACTORY`.

uzshao:
@ljfgem Thanks for the review.  I just addressed all review comments.  Can you take a look again?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilder.java
```diff
@@ -265,9 +267,15 @@ private SqlNode visitLateralViewUDTF(List<SqlNode> sqlNodes, List<SqlNode> alias
...
-    checkState(operandCount == 3 || operandCount == 4,
+    // array if operandCount == 3: LATERAL VIEW EXPLODE(op0) op1 AS op2
+    // map if operandCount == 4: LATERAL VIEW EXPLODE(op0) op1 AS op2, op3
+    // if operandCount == 2: LATERAL VIEW EXPLODE(op0) op1
+    //   if op0 is a map, it implies "AS key, value" where key, value are auto-named by Hive
+    //   if op0 is an array, is implies "AS col" where col is auto-named by Hive
+    //   The logic above will be implemented as part of Calcite SqlNode validation
+    //   Note that "operandCount == 2 && isOuter" is not supported yet due to the lack of type information needed
+    //   to derive the correct IF function parameters.
+    checkState((operandCount == 2 && !isOuter) || operandCount == 3 || operandCount == 4,
```
ljfgem, line 278: Is it better to throw the exception for `operandCount == 2` in the following `if (isOuter)` block? Otherwise, if user uses `operandCount == 2 && isOuter`, they will get the misleading error message because we do support `operandCount == 2 && !isOuter`

ljfgem (commented):
Thanks for the update @uzshao, just added one nit.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilder.java
uzshao: Good catch @ljfgem !  Just fixed.  Please take another look.

uzshao:
@ljfgem I've fixed the nit.  Can you take a look again?  Thanks!

ljfgem:
> @ljfgem I've fixed the nit. Can you take a look again? Thanks!

Hi @uzshao , integration test just finished, but there is a regression.
Could you try to translate 
```
SELECT arr.alias FROM test.tableA tmp LATERAL VIEW EXPLODE(ARRAY('a', 'b')) arr as alias
```
by adding the following test in Coral-Trino?
```
  @Test
  public void testLateralViewArray2() {
    RelNode relNode = hiveToRelConverter
        .convertSql("SELECT arr.alias FROM test.tableA tmp LATERAL VIEW EXPLODE(ARRAY('a', 'b')) arr as alias");

    String targetSql = "SELECT \"t00\".\"alias\" AS \"alias\"\n" + "FROM \"test\".\"tablea\",\n"
        + "UNNEST(ARRAY['a', 'b']) AS \"t0\" (\"alias\") AS \"t00\"";
    RelToTrinoConverter relToTrinoConverter = new RelToTrinoConverter();
  }
```
The result seems to be incorrect. And when I query the result in trino, there will be exception like:
```
mismatched input 'AS'. Expecting: ',', 'CROSS', 'EXCEPT', 'FETCH', 'FULL', 'GROUP', 'HAVING', 'INNER', 'INTERSECT', 'JOIN', 'LEFT', 'LIMIT', 'NATURAL', 'OFFSET', 'ORDER', 'RIGHT', 'TABLESAMPLE', 'UNION', 'WHERE', 'WINDOW', <EOF>
```
The issue seems also in Coral-Spark.

uzshao:
@ljfgem Thanks for the report.  The issue happens with `LATERAL VIEW EXPLODE` that is not correlated.

I've updated the code to fix such a situation.

Also fixed an incorrect existing test case https://github.com/linkedin/coral/pull/137/files#diff-d76c70181ecb53951e593e111ca716b3c3121bd6f08b0d40297fd301e51855b7L253

Please take a look again.

uzshao:
@ljfgem Can you help run integration test again?  This diff touches several files so the cost of rebase is pretty high for me.  Hope to get this in soon after the review is complete.  Thanks!

ljfgem:
> @ljfgem Can you help run integration test again? This diff touches several files so the cost of rebase is pretty high for me. Hope to get this in soon after the review is complete. Thanks!

Hi @uzshao, I have started the integration test several hours ago, will update once it finishes.

uzshao:
@wmoustafa @funcheetah it will be great if you can take a look (or if you are comfortable with it, just stamp it).

funcheetah (approved):
Thanks @uzshao . LGTM.

uzshao:
@wmoustafa Would you take a look also (or stamp it if you are comfortable with it)?

wmoustafa (approved):
Thanks @uzshao! Appreciate all the discussions on this patch.

PR #143, opened 2021-09-07
Coral-Trino: Use `cast(xxx as timestamp)` for `timestamp` operator

ljfgem:
`timestamp xxx` only supports pure date string like `"2020-01-02"`, if the operand is a function returning string like `timestamp substr(xxx)`, Trino will throw an exception. Therefore, we use `cast(xxx as timestamp)` for `timestamp` operator instead.

Tests:
1. unit test
2. tested on the affected production views, which could be translated well with this patch
3. integration test, no regression


ljfgem:
Opened an issue for trino: https://github.com/trinodb/trino/issues/9140

wmoustafa (approved):

PR #144, opened 2021-09-08
Coral-Trino: Remove extra artificial wrapping `ROW` for `LATERAL VIEW EXPLODE` over ARRAY of STRUCTS

ljfgem:
#93 added an extra artificial wrapping `ROW` for `LATERAL VIEW EXPLODE` over ARRAY of STRUCTS, which will cause the issue while querying the translated Trino SQL.

i.e.
```
CREATE TABLE test.table_with_struct_array(a int, b array<struct<sa: int, sb: string>>)
CREATE VIEW test.view_with_explode_struct_array AS SELECT a, c.sa FROM test.table_with_struct_array LATERAL VIEW EXPLODE(b) t AS c
```
The translated Trino SQL of view `test.view_with_explode_struct_array` with this PR would be:
```
SELECT "$cor0"."a" AS "a", "t0"."c"."sa" AS "sa"
FROM "test"."table_with_struct_array" AS "$cor0"
CROSS JOIN UNNEST(TRANSFORM("$cor0"."b", x -> ROW(x))) AS "t0" ("c")
```
Since we used `TRANSFORM("$cor0"."b", x -> ROW(x))`, the schema of `t0.c` would be `row(row(sa integer, sb varchar))`, therefore, trino couldn't find `sa` from `t0.c` and will throw the exception `Column 't0.c.sa' cannot be resolved`.

Before patch #93, without `TRANSFORM("$cor0"."b", x -> ROW(x))`, the schema of `t0.c` was `row(sa integer, sb varchar)` and `t0.c.sa` could be resolved and the translated trino SQL could run well.

This PR is to remove the change made by #93 

Tests:
1. Modified unit test, the translated Trino SQL could not be resolved before this PR.
2. Tested on the affected production views

### DIFF
--- a/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/RelToTrinoConverter.java
+++ b/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/RelToTrinoConverter.java
@@ -18,20 +18,15 @@ import org.apache.calcite.rel.core.*;
 import org.apache.calcite.rel.rel2sql.RelToSqlConverter;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeField;
-import org.apache.calcite.rel.type.RelDataTypeFieldImpl;
-import org.apache.calcite.rel.type.RelRecordType;
 import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.sql.*;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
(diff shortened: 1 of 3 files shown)


funcheetah (approved):

ljfgem:
In case I missed something, @wmoustafa since you approved #93, do you think we can revert #93 using this patch?

findepi (changes requested):
Assuming this is a revert of https://github.com/linkedin/coral/pull/93, this is not a correct change. I've added a comment to the ongoing discussion there.

ljfgem:
Closing it since we should not revert #93, we will figure out a way to handle the internal legacy unnest behavior for Coral-Trino.

PR #145, opened 2021-09-08
Coral-Hive: Treat `CHAR` with length > 255 as `VARCHAR` in `TypeConverter.convertPrimtiveType`

kxu1026:
Function with a string parameter will be converted to SqlStringLiteral Node as an operand. The SqlStringLiteral Node is a char type node. And, The validation stage will rely on the node type to validate the value via hive serde. However, char type in HIVE is limited to 255. We need to CHAR convertPrimativeType to either char or varchar based on the length of the string.


ljfgem (approved):
LGTM, thanks @kxu1026
Integration test is running, will merge once it succeeded.

PR #147, opened 2021-09-09
coral incorrectly get rid of CAST in converting double constants to integer

uzshao:
This test case shows a bug.  One possible solution is to modify the code in `HiveConvertletTable.java`'s `convertCast` to always call `return cx.getRexBuilder().makeAbstractCast(castType, leftRex);`:

CC @antumbde Any thoughts?  Since you added the `if` condition in the code

```
  public RexNode convertCast(SqlRexContext cx, SqlCastFunction cast, SqlCall call) {
    final SqlNode left = call.operand(0);
    if (SqlUtil.isNullLiteral(left, false)) {
      RexNode leftRex = cx.convertLiteral((SqlLiteral) left);
      SqlDataTypeSpec dataType = call.operand(1);
      RelDataType castType = dataType.deriveType(cx.getValidator(), true);
      // can not call RexBuilder.makeCast() since that optimizes to remove the cast
      // we don't want to remove the cast
      return cx.getRexBuilder().makeAbstractCast(castType, leftRex);
    }
    // this is odd but we want to re-use as much code from calcite as possible
    return StandardConvertletTable.INSTANCE.get(call).convertCall(cx, call);
  }
```

### DIFF
--- a/coral-spark/src/test/java/com/linkedin/coral/spark/CoralSparkTest.java
+++ b/coral-spark/src/test/java/com/linkedin/coral/spark/CoralSparkTest.java
@@ -342,6 +342,14 @@ public class CoralSparkTest {
   }
 
   @Test
+  public void testCastByTypeName() {
+    RelNode relNode = TestUtils.toRelNode("SELECT DOUBLE(1), INT(1.5), STRING(2.3), TIMESTAMP(1631142817), BOOLEAN('')  FROM default.complex");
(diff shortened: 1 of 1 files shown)


ljfgem:
Hi @uzshao, thanks for reporting this issue. I agree that we should always call `return cx.getRexBuilder().makeAbstractCast(castType, leftRex);`, would you mind modifying this PR to add the fix?

PR #150, opened 2021-09-10
Coral-Spark: Avoid casting `call` to `SqlLateralJoin` in `SqlLateralJoin.unparse`

ljfgem:
Previously, we cast `call` to `SqlLateralJoin` in `SqlLateralJoin.unparse`, which might cause the following exception:
```
java.lang.ClassCastException: org.apache.calcite.sql.SqlBasicCall cannot be cast to com.linkedin.coral.spark.functions.SqlLateralJoin
```
We can get the needed information using `call.getOperandList()` directly without the casting.

Tests:
1. Tested on affected views, could be translated well with this patch
2. Integration test, no regression

### DIFF
--- a/coral-spark/src/main/java/com/linkedin/coral/spark/functions/SqlLateralJoin.java
+++ b/coral-spark/src/main/java/com/linkedin/coral/spark/functions/SqlLateralJoin.java
@@ -5,6 +5,9 @@
  */
 package com.linkedin.coral.spark.functions;
 
+import java.util.List;
+
+import org.apache.calcite.sql.JoinType;
 import org.apache.calcite.sql.SqlCall;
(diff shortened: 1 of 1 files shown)


funcheetah:
Nice PR @ljfgem ! Is it possible to add a unit test for this?

ljfgem:
> Nice PR @ljfgem ! Is it possible to add a unit test for this?

Thanks for the review @funcheetah , I tried but failed to reproduce the issue of that affected production view since the definition of that affected production view is not very special. 

I think it is fine since there are many existing unit tests for `lateral view explode` in Coral-Spark. This PR just changed to use a safer way to unparse the `SqlLateralJoin`

funcheetah (approved):

PR #158, opened 2021-09-17
Coral-Trino: Add config for LinkedIn's internal use

ljfgem:
This PR adds the following configs for LinkedIn's internal use:
1. Legacy unnest array of struct: our internal trino only unnests `ARRAY` in `ARRAY<ROW<...` instead of both `ARRAY` and `ROW`, so we add this config which can avoid the extra wrapping added by #93 
2. Avoid transforming internally registered trino UDF `to_date`

All the external uses of Coral-Trino wouldn't be affected by this patch.

Tests:
1. Unit tests
2. Integration test, no regression

### DIFF
--- a/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/CoralTrinoConfigKeys.java
+++ b/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/CoralTrinoConfigKeys.java
@@ -9,6 +9,21 @@ package com.linkedin.coral.trino.rel2trino;
  * Contains configuration keys for {@link RelToTrinoConverter#configs}
  */
 public class CoralTrinoConfigKeys {
+  /**
+   * SQL standard (which Trino follows) defines that when we do CROSS JOIN UNNEST over ARRAY(ROW(...)), both ARRAY and ROW are unnested.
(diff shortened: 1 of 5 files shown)


coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/CoralTrinoConfigKeys.java
```diff
@@ -0,0 +1,14 @@
...
+ * Licensed under the BSD-2 Clause license.
+ * See LICENSE in the project root for license information.
+ */
+package com.linkedin.coral.trino.rel2trino;
+
+/**
+ * Contains configuration keys for {@link RelToTrinoConverter#configs}
+ */
+public class CoralTrinoConfigKeys {
+  public static final String SUPPORT_LEGACY_UNNEST_ARRAY_OF_STRUCT = "SUPPORT_LEGACY_UNNEST_ARRAY_OF_STRUCT";
```
funcheetah, line 12: Please add documentation on what each of these config key means since this is a central place where those keys are stored.

funcheetah (approved):
Thanks @ljfgem for this PR

PR #160, opened 2021-09-23
Add a visit rule to cancel out extract_union in rto views trino trans…

rzhang10:
…lation

This is a demo pr for now since the `reverse_extract_union_placeholder` function is to be implemented by rto team to make this whole thing work.

### DIFF
--- a/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
+++ b/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
@@ -201,12 +201,33 @@ public class Calcite2TrinoUDFConverter {
         }
       }
 
+      if (operatorName.equalsIgnoreCase("reverse_extract_union_placeholder")) {
+        Optional<RexNode> modifiedCall = cancelExtractUnion(call);
+        if (modifiedCall.isPresent()) {
+          return modifiedCall.get();
+        }
+      }
+
       final UDFTransformer transformer = CalciteTrinoUDFMap.getUDFTransformer(operatorName, call.operands.size());
       if (transformer != null && shouldTransformOperator(operatorName)) {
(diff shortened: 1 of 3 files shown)


coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
```diff
@@ -209,6 +216,22 @@ public RexNode visitCall(RexCall call) {
       return super.visitCall(modifiedCall);
     }
 
+    private Optional<RexNode> cancelExtractUnion(RexCall call) {
+      if (!(call.getOperands().get(0) instanceof RexCall)) {
+        throw new IllegalStateException(
+            "reverse_extract_union_placeholder should always be called with an internal extract_union function");
+      }
+      RexCall internalCall = (RexCall) call.getOperands().get(0);
```
ljfgem, line 224: I think we need to call
```
List<RexNode> convertedOperands = visitList(call.getOperands(), (boolean[]) null);
```
and use `convertedOperands` afterwards. Otherwise, the inner operators would not be handled.
i.e. `reverse_extract_union_placeholder(extract_union(reverse_extract_union_placeholder(extract_union(a))))`

ljfgem (commented):
Thanks @rzhang10 
Will review again when it is ready, left a comment in case I forget it.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
rzhang10: Yes, that's a good point, generally we need to do this, but for this case, `extract_union` is a very special function which will only be applied to a column reference, so we can guarantee we will only have `extract_union(col)`, we won't have`extract_union(arbitrary_transformation...(a))`.

PR #161, opened 2021-09-24
Update udf map to point latest spark3-compatible transport udf urls

rzhang10:
Previously those null values mean that those transport udfs haven't had their spark2/3 separate artifacts, this patch updates the urls to their corresponding artifacts url using classifiers.

I verified all these transport udfs are already migrated and pcl passed and the versions are on the artifactory.

cc @HotSushi


HotSushi (approved):
Thank you @rzhang10

HotSushi:
Just make sure the build passes

PR #162, opened 2021-09-25
Coral-Hive: Add `hive-exec` if `GenericUDF` class could not be found

ljfgem:
If `GenericUDF` class could not be found, add `hive-exec:core` in `_udfDependencies` to download the missing class.

Tested on the affected views, which could be translated well with this PR.

### DIFF
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveGenericUDFReturnTypeInference.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveGenericUDFReturnTypeInference.java
@@ -105,7 +105,7 @@ public class HiveGenericUDFReturnTypeInference implements SqlReturnTypeInference
       if (error.getMessage().contains("GenericUDF")) {
         // If GenericUDF class could not be found, add `hive-exec:core` in `_udfDependencies` to download the missing class
         _udfClassLoader = null; // set it to null to re-download
-        _udfDependencies.add("com.linkedin.hive:hive-exec:1.1.0.+:core");
+        _udfDependencies.add("org.apache.hive:hive-exec:1.1.0");
         return Class.forName(_udfClassName, true, getUdfClassLoader());
       }
       throw error;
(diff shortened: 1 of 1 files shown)


coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveGenericUDFReturnTypeInference.java
```diff
@@ -99,7 +99,17 @@ private URLClassLoader getUdfClassLoader() {
...
 
   private final Class getDynamicallyLoadedUdfClass() throws ClassNotFoundException {
-    return Class.forName(_udfClassName, true, getUdfClassLoader());
+    try {
+      return Class.forName(_udfClassName, true, getUdfClassLoader());
+    } catch (NoClassDefFoundError error) {
+      if (error.getMessage().contains("GenericUDF")) {
+        // If GenericUDF class could not be found, add `hive-exec:core` in `_udfDependencies` to download the missing class
+        _udfClassLoader = null; // set it to null to re-download
+        _udfDependencies.add("com.linkedin.hive:hive-exec:1.1.0.+:core");
```
wmoustafa, line 108: Is there a way to infer this automatically from the UDF dependencies? Upstream dependencies may not match the Hive version, the `com.linkedin.hive` group name, or the `core` classifier.

ljfgem: I tried but didn't find a way to infer this automatically from the UDF dependencies, could we change it to `org.apache.hive:hive-exec:1.1.0` so that the upstream dependencies could also match and retrieve it.

wmoustafa: Yes, if it works internally.

ljfgem: Thanks! Updated and it could work internally.

wmoustafa (approved):

PR #165, opened 2021-10-01
Wrap timestamp with "to_unixtime" when casting from timestamp to decimal

jongn:
Hive allows casting TIMESTAMP type to DECIMAL type, which casts the timestamp into unix time with decimal type.

```
0: jdbc:hive2://localhost:10000> describe default.timestamp_test_base_2;
| t1        | timestamp  |          |
```
```
0: jdbc:hive2://localhost:10000> select cast(timestamp_test_base_2.t1 as decimal(10,0)) as d from default.timestamp_test_base_2;
+-------------+
|      d      |
+-------------+
| 1631137219  |
+-------------+
```

Trino/Presto does not allow for this kind of casting, and attempting the same query results in
`Cannot cast timestamp to decimal(10,0)`

 So we propose mapping
`CAST(timstamp_col AS DECIMAL(10,0))`

To

`CAST("to_unixtime"(timstamp_col) AS DECIMAL(10,0))`

In order to achieve the same behavior.

### DIFF
--- a/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
+++ b/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
@@ -39,6 +39,7 @@ import org.apache.calcite.sql.SqlOperator;
 import org.apache.calcite.sql.fun.SqlMapValueConstructor;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.type.SqlTypeFamily;
+import org.apache.calcite.sql.type.SqlTypeName;
 
 import com.linkedin.coral.com.google.common.collect.ImmutableList;
 import com.linkedin.coral.com.google.common.collect.ImmutableMultimap;
@@ -201,6 +202,13 @@ public class Calcite2TrinoUDFConverter {
         }
       }
 
+      if (operatorName.equalsIgnoreCase("cast")) {
+        Optional<RexNode> modifiedCall = visitCast(call);
+        if (modifiedCall.isPresent()) {
+          return modifiedCall.get();
+        }
+      }
+
       final UDFTransformer transformer = CalciteTrinoUDFMap.getUDFTransformer(operatorName, call.operands.size());
       if (transformer != null && shouldTransformOperator(operatorName)) {
         return super.visitCall((RexCall) transformer.transformCall(rexBuilder, call.getOperands()));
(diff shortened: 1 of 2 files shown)


coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
```diff
@@ -279,6 +287,23 @@ public RexNode visitCall(RexCall call) {
       return Optional.empty();
     }
 
+    private Optional<RexNode> visitCast(RexCall call) {
```
ljfgem, line 290: Could you please add a commend with an example here?

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
```diff
@@ -279,6 +287,23 @@ public RexNode visitCall(RexCall call) {
...
+      final SqlOperator op = call.getOperator();
+      if (op.getKind() != SqlKind.CAST) {
+        return Optional.empty();
+      }
+
+      RexNode leftOperand = call.getOperands().get(0);
```
ljfgem, line 296: I think we need to use
```
  List<RexNode> convertedOperands = visitList(call.getOperands(), (boolean[]) null);
  RexNode leftOperand = convertedOperands.get(0);
```
here to take care of the possible nested transformations.

For example, for SQL
```
select cast(from_utc_timestamp(2592000.0,'PST') as decimal(10, 0))
```
we also need to transform the inner `from_utc_timestamp`.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
```diff
@@ -279,6 +287,23 @@ public RexNode visitCall(RexCall call) {
...
+
+      if (call.getType().getSqlTypeName().equals(SqlTypeName.DECIMAL)
+          && leftOperand.getType().getSqlTypeName().equals(SqlTypeName.TIMESTAMP)) {
```
ljfgem, line 299: To be aligned with the previous syntax, could we modify it to be:
```
call.getType().getSqlTypeName() == DECIMAL && leftOperand.getType().getSqlTypeName() == TIMESTAMP
```
?

ljfgem (commented):
Thanks @jongn for the PR! Left some comments.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
jongn: Added in revision

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
jongn: Good catch, thanks! Fixed this in the revision and added a unit test for this case as well.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
jongn: Fixed in revision

ljfgem (approved):
LTGM, thanks @jongn 
Integration test passed.

PR #109, opened 2021-07-08
Add support for from_utc_timestamp in Hive->Trino translation

losipiuk:
Matching Trino PR: https://github.com/trinodb/trino/pull/8502

@wmoustafa @hashhar @findepi PTAL

PR #169, opened 2021-10-08
Coral-Trino: Add `with_timezone` wrapper while casting timestamp to decimal

ljfgem:
As discussed in https://github.com/trinodb/trino/pull/9530#discussion_r723399144
We need to add a wrapper `with_timezone"(..., 'UTC')` before calling `to_unixtime`, so it is timezone agnostic like how #109 did.


findepi:
@losipiuk ptal

PR #172, opened 2021-10-14
coral-schema: Generate documentation for literal fields in views

sherryhli:
### Summary
An enhancement on #115, which added field-level documentation to generated Avro schemas of view fields directly selected from underlying tables. This PR adds support for generating documentation for fields created from literals, e.g. `Additional_Int` in the view below:

```sql
CREATE VIEW v AS
SELECT
  bc.Id AS Id_View_Col,
  100 AS Additional_Int
FROM basecomplexwithdoc bc
```

The documentation is in the form of: `"Field created from view literal with value: <VALUE-OF-LITERAL>"`.

### Notes
The current logic for merging schemas in the case of unions will always select the documentation of fields from the union's left side. This can result in somewhat odd cases where `Field` in a view like
```sql
CREATE VIEW v AS SELECT NULL Field FROM basecomplex UNION ALL SELECT Id Field FROM basecomplex
```
has the documentation `"Field created from view literal with value: null"`, even though it may be more natural to favour the documentation of the non-literal `basecomplex.Id`. However, at the time of merging the schemas, we cannot differentiate between fields originating from literals and non-literals. As discussed with @funcheetah, we will maintain the behaviour of using the left side's documentation.

### Testing
There are existing unit test cases covering literals (e.g. `testSelectWithMultipleLiterals`, `testSelectWithLiteralsWithDoc`). The `.avsc` files containing expected schemas for those cases have been updated to include the documentation field.


funcheetah:
Thanks for the nice description and PR @sherryhli !

ljfgem (approved):
LGTM, thanks @sherryhli for this nice PR.

funcheetah (approved):
Thanks @sherryhli . Look forward to the followup PRs.

PR #177, opened 2021-10-19
Wrap non-string input to substring function

jongn:
Hive substr/substring [function](https://cwiki.apache.org/confluence/display/hive/languagemanual+udf) allows user to input either a String or byte array, but Calcite's substring function only allows String input.

```
describe string_timestamp_table;
+--------------+------------+----------+
|   col_name   | data_type  | comment  |
+--------------+------------+----------+
| a_string     | string     |          |
| a_timestamp  | timestamp  |          |
+--------------+------------+----------+
```
```
select substr(string_timestamp_table.a_timestamp, 12, 8) as newtime from default.string_timestamp_table;
+-----------+
|  newtime  |
+-----------+
| 22:54:24  |
| 22:54:54  |
+-----------+
```

One way we could try to emulate this behavior would be to attempt to wrap any non-String input in substr/substring function to String type. This might be a bit funky since we need to remove the mapping to Calcite's substring and instead register it as a UDF. Additionally this fix only works for 'substr' but not 'substring', since Calcite will actually [override](https://github.com/linkedin/linkedin-calcite/blob/c13ba44c6974574d07a47766e32e695e5c5e042f/core/src/main/java/org/apache/calcite/sql/SqlUtil.java#L329) the UDF.

Open to suggestions if there are better approaches to this.


jongn:
Something happened to my feature branch and now the diff is wonky

Going to re-make the PR

jongn:
https://github.com/linkedin/coral/pull/190

PR #48, opened 2021-02-18
Support Hive get_json_object to Presto json_extract

ebyhr:

PR #182, opened 2021-10-21
Coral-Trino: Modify converted function for `get_json_object`

ljfgem:
In #48, `get_json_object` is converted to `json_extract`. `get_json_object` in Hive returns string type but `json_extract` in Trino returns json type. This PR is to modify the converted function to [json_extract_scalar](https://trino.io/docs/current/functions/json.html?highlight=json_extract#json_extract_scalar) which also returns varchar type to be aligned with the return type of `get_json_object`.

@ebyhr @findepi PTAL, thanks!


jerryleooo:
If I am correct, the root cause is Trino has no exact mapping for `get_json_object`, which return string even the return value is not a scalar. 
@findepi maybe in Trino we can change the behaviour of `json_extract_scalar` by implicit casting to return varchar even the return value is not scalar, or providing another exact mapping to `get_json_object`?

PR #132, opened 2021-08-22
Support some date calculation functions for hive2trino

JiajunBernoulli:
Some date calculation functions are different betweend hive and trino， here are some examples.
- [to_date](https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFToDate.java)
```c
/*Hive*/
select to_date('2021-08-20'); //2021-08-20
select to_date('2021-08-20 00:00:00'); //2021-08-20
/*Trino*/
select date(TIMESTAMP '2021-08-20'); //2021-08-20
select date(TIMESTAMP '2021-08-20 00:00:00'); //2021-08-20
```
- [date_add](https://prestodb.io/docs/current/functions/datetime.html?highlight=date_add#date_add) 

```c
/*Hive*/
select date_add('2021-08-20', 1); // '2021-08-21'
select date_add('2021-08-20 00:00:00', 1); // '2021-08-21'
/*Trino*/
select date_add('day', 1, TIMESTAMP '2021-08-20 00:00:00');  
//2021-08-21 00:00:00 is timestamp. In order to be consistent with hive's results, we need cast it as date.
select date_add('day', 1, DATE(TIMESTAMP '2021-08-20'));   // '2021-08-21'
 select date_add('day', 1, DATE(TIMESTAMP '2021-08-20 00:00:00'));  // '2021-08-21'
```
- [date_sub](https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFDateSub.java)
```c
/*Hive*/
select date_sub('2021-08-20', 1); // '2021-08-19'
select date_sub('2021-08-20 00:00:00', 1); // '2021-08-19'
/*date_sub is not in the Trino*/
select date_add('day', -1, date(TIMESTAMP '2021-08-20')); //2021-08-19
select date_add('day', -1, date(TIMESTAMP '2021-08-20 00:00:00')); //2021-08-19
```

- [date_diff](https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/udf/generic/GenericUDFDateDiff.java)

```c
/*Hive*/
select datediff('2021-08-20', '2021-08-21'); //-1, input1-input2
select datediff('2021-08-20', '2021-08-19');  //1
select datediff('2021-08-20 00:00:00', '2021-08-19 23:59:59'); //1
/*Trino*/
[...]

PR #185, opened 2021-10-28
Coral-Trino: Add type cast to avoid return type mismatch between Hive function and Trino function

ljfgem:
#132 made the following conversion:
```
datediff('2021-08-20', '2021-08-21') -> "date_diff"('day', "date"(CAST('2021-08-21' AS TIMESTAMP)), "date"(CAST('2021-08-20' AS TIMESTAMP)))
```
However, `datediff` in Hive returns `int` type, but `date_diff` in Trino returns `bigint` type, the type discrepancy would cause the issue while querying the view on Trino, so we need to add the CAST for it.
This PR introduces a method `adjustReturnTypeWithCast` to add the CAST for functions with required return types using a map. 
Similar use case:
1. `Cardinality`: Refer to #55 for more info, this PR can help with that issue
2. `date_add`: at LinkedIn, as we are using Hive < 2.1.0, `date_add` returns `string` type instead of `date` type. However, Trino's `date_add` returns `date` type. So we add a new config `CAST_DATEADD_TO_STRING` for our internal use.

Tests:
1. Unit test
2. Test on affected production views
3. Enhanced integration test, no regression


funcheetah (approved):
Nice PR! Thanks @ljfgem

PR #186, opened 2021-10-28
coral-schema: Generate documentation for aggregates in views

sherryhli:
### Summary

Previously, field-level documentation in schemas produced by the `ViewToAvroSchemaConverter` was only available for fields directly selected from underlying base tables (#115), or those created from literal values (#172). This PR adds documentation for fields created from aggregate function calls (e.g. `COUNT`, `SUM`, `AVG`, etc).

The documentation is in the form of: `"Field created in view by applying aggregate function of type: <TYPE>"`

### Testing
One new unit test has been added (`testMultipleAggregates`). Some existing test cases include aggregate function calls and the `.avsc` files with their expected schemas have been updated to include the expected documentation.

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
@@ -295,7 +295,8 @@ public class RelToAvroSchemaConverter {
       for (Pair<AggregateCall, String> aggCall : logicalAggregate.getNamedAggCalls()) {
         String fieldName = SchemaUtilities.toAvroQualifiedName(aggCall.right);
         RelDataType fieldType = aggCall.left.getType();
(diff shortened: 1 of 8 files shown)


ljfgem (approved):
LGTM, thanks @sherryhli !
No regression in i-test.
Looks like there is another part not having doc: https://github.com/linkedin/coral/blob/b515576e89cf14432a485ffc615ca9cb595ba4f6/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java#L332
Are we planning to handle it in the future?

sherryhli:
@ljfgem Thanks for the review!

We weren't planning on handling the above case, but I can definitely look into it later. The one other case we're handling is function calls (UDFs and operators):
https://github.com/linkedin/coral/blob/b515576e89cf14432a485ffc615ca9cb595ba4f6/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java#L424

funcheetah (approved):
Nice PR. Thanks @sherryhli !

PR #191, opened 2021-11-03
Pass down precision/scale when casting to DECIMAL

jongn:
Related Trino issue:
https://github.com/trinodb/trino/issues/6450

When casting column to DECIMAL type, precision was not being preserved during the Hive -> Rel translation, resulting in default precision/scale of 10,0 respectively.

This PR is to pass the precision/scale in Hive -> Rel translation, if applicable.


jongn:
@ljfgem PTAL if you get the chance, thanks!

ljfgem:
@jongn Thank you so much for fixing it.
Could you please also add the unit tests for coral-spark?

jongn:
Thanks @ljfgem, added the spark tests.

I'll add corresponding Trino tests once this is merged.

ljfgem (approved):
LGTM, thanks @jongn !
i-test passed

PR #192, opened 2021-11-09
[Coral-Common] Convert Hive uniontype into a struct-RelDataType that conforms Trino' schema

autumnust:
Background: 
- Previously to make union type translatable between engines, we represent the union-type in Coral-IR as a struct where each member field of union becomes a subfield within a struct with each carrying the name `tag_N`. For example we would see `uniontype<int,string>` becomes something like `struct<1,null>` or `struct<null, "h">`.  This format is based on Hive's `extract_union` UDF. 
- Trino supports natively reading union field by exploding the union into a struct, with a different schema: https://github.com/trinodb/trino/pull/3483 
- This PR conforms the Coral IR's union-after-explosion representation to Trino's format, so that there won't be type differences between engine's deserialized format versus IR of a view's field. 

Also added unit tests for the different union conversion logic. 

--- 
Updates from long discussion in this PR: 
- For some of the corner cases like empty union, we decided to keep consistent with Trino's PR where it will convert the empty union into a struct with only one field named as `tag`. 
- The tag field is in "TINYINT" type. Note that this type does not exist in Iceberg so we will need further testing to verify the whole flow works.


autumnust:
@rzhang10  @funcheetah  Can you take a look when you get a chance? thanks !

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/TestUtils.java
```diff
@@ -357,6 +357,10 @@ public static RelNode convertView(String db, String view) {
     return new HiveToRelConverter(hiveMetastoreClient).convertView(db, view);
   }
 
+  public static RelNode convertView(String sql) {
+    return new HiveToRelConverter(hiveMetastoreClient).convertSql(sql);
+  }
+
```
ljfgem, line 363: Why do we need this method?

coral-spark/src/main/java/com/linkedin/coral/spark/IRRelToSparkRelTransformer.java
```diff
@@ -323,6 +323,7 @@ public RexNode visitCall(RexCall call) {
       return Optional.empty();
     }
 
+    // TODO:This removal is no longer needed
```
ljfgem, line 326: Can we handle it in this PR and modify the corresponding tests in Coral-Spark?
`testUnionExtractUDF` in `CoralSparkTest` will fail if we don't modify it.

ljfgem (commented):
Thanks @autumnust !
Could you also add the unit tests for Coral-Trino in `HiveToTrinoConverterTest`?

coral-spark/src/main/java/com/linkedin/coral/spark/IRRelToSparkRelTransformer.java
ljfgem: And before committing, we can run `./gradlew clean build` to make sure that the Github CI can pass.

coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
```diff
@@ -141,11 +141,16 @@ public static RelDataType convert(StructTypeInfo structType, final RelDataTypeFa
...
 
   // Mimic the StructTypeInfo conversion to convert a UnionTypeInfo to the corresponding RelDataType
+  // The schema of output Struct conforms to https://github.com/trinodb/trino/pull/3483
   public static RelDataType convert(UnionTypeInfo unionType, RelDataTypeFactory dtFactory) {
     List<RelDataType> fTypes = unionType.getAllUnionObjectTypeInfos().stream()
         .map(typeInfo -> convert(typeInfo, dtFactory)).collect(Collectors.toList());
-    List<String> fNames = IntStream.range(0, unionType.getAllUnionObjectTypeInfos().size()).mapToObj(i -> "tag_" + i)
+    List<String> fNames = IntStream.range(0, unionType.getAllUnionObjectTypeInfos().size()).mapToObj(i -> "field" + i)
         .collect(Collectors.toList());
+    if (fNames.size() > 0) {
```
rzhang10, line 150: I think we don't need this if statement since an empty union is a wrong schema, to begin with, maybe we could add a `Preconditions.checkState` to check it.

coral-spark/src/main/java/com/linkedin/coral/spark/IRRelToSparkRelTransformer.java
autumnust: thx for the suggestion. will address.

autumnust: actually what about temporarily disabling the `testUnionExtractUDF` as part of this PR ? I do want to separate them out since change in this PR is happening in the Coral IR that could have broader impact.

ljfgem: Sure, I am fine with separating them.

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/TestUtils.java
ljfgem: I would prefer to remove the method if it's not used, and we can add if we need it in the future.

autumnust: ack. will address.

coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
wmoustafa: Seems that here https://github.com/trinodb/trino/pull/3483/files# the size is not checked. Does the Avro or ORC standard say anything about this?

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/HiveTableTest.java
```diff
@@ -63,6 +63,34 @@ public void testTable() throws Exception {
     RelDataTypeField colC = rowType.getField("c", false, false);
     assertEquals(colC.getType().getSqlTypeName(), SqlTypeName.ARRAY);
     assertEquals(colC.getType().getComponentType().getSqlTypeName(), SqlTypeName.DOUBLE);
+
+    // test handling of union
```
wmoustafa, line 67: Should this code go to a separate test method?

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/HiveTableTest.java
```diff
@@ -63,6 +63,34 @@ public void testTable() throws Exception {
...
+    Table unionTable = getTable("default", "union_table");
+    // union_table:(foo uniontype<int, double, array<string>, struct<a:int,b:string>>)
+    // expected outcome schema: struct<tag:int, field0:int, field1:double, field2:array<string>, field3:struct<a:int,b:string>>
+    rowType = unionTable.getRowType(typeFactory);
+    assertNotNull(rowType);
+    assertTrue(rowType.isStruct());
+    // top-level: it is just a struct
+    assertEquals(rowType.getFieldCount(), 1);
```
wmoustafa, line 75: Can we refactor the assertions so instead of checking on things like field count of certain field names, they assert the whole expected schema?

autumnust: I figured the easiest way to do such assertion is to create a table with equivalent exploded union schema. If we would like to have a utility method to achieve "given a schema described in certain way and return the equivalent RelDataType", it will be pretty much the same thing as the visitor in Hive type to RelDataType conversion which I felt redundant.

coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
autumnust: ORC spec doesn't mention it while I do see `uniontype<>` happening in production. Regardless, this check is there to keep parity with the original (or we could end up with an additional `tag` field).

coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
```diff
@@ -141,11 +141,16 @@ public static RelDataType convert(StructTypeInfo structType, final RelDataTypeFa
...
+      fTypes.add(0, dtFactory.createSqlType(SqlTypeName.INTEGER));
```
ljfgem, line 151: According to the corresponding PR:
https://github.com/trinodb/trino/pull/3483/files#diff-6a37b030f26bfeb6eca302ca157050cad2ead0476c1785fad57fc5f14fab88cfR264
the type of tag is `TINYINT`, rather than `INT`, so it should be 
```
fTypes.add(0, dtFactory.createSqlType(SqlTypeName.TINYINT));
```
querying the view will fail if the types mismatch.
I think the test also needs to be modified.

Enhanced coral-trino test suite helped me to find this hidden issue 😉

ljfgem (commented):
LTGM except the type issue, thanks @autumnust !

coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
rzhang10: So if using the current code, `uniontype<>` ends up in an empty struct? I recall there are some issues with empty struct somewhere, either in Iceberg or in Spark. So I'm not sure if this whole corner case will work or not.
My point is if we explicitly announce this special case's behavior is undefined, then users have the responsibility to create the correct schema, it's a tradeoff between us and the users, I'm willing to accept either way.

coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
autumnust: Wow that's a great catch, thanks ! I will address this.

coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
wmoustafa: Our references should be: 1- what the standard says (again for Avro or ORC) 2- what the Trino transformation does. Hopefully all of them align. If not, we can discuss how to move forward.

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/HiveTableTest.java
wmoustafa: Does `RelDataType#toString()` or `RelDataType#getFullTypeString()` help?

coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
funcheetah: Another special case is union with only a single type `uniontype<[type]>`. We should consider whether we want to support it and make sure its semantics is consistent across Trino and Iceberg implementation

coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
funcheetah: Nice work for enhanced coral-trino test suite! This is awesome! We can catch more issues before hitting production.

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/HiveTableTest.java
ljfgem: +1 to use `rowType.toString()` here so that we don't need to create another table `exploded_union` for test.

ljfgem:
FYI, enhanced i-test succeeded after fixing the type mismatch issue, we can merge it once all the comments are addressed.

coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
autumnust: There aren't any spec that I could find regarding this point in either Avro or ORC, and looks like Trino doesn't check the size either. 
Taking a step back, I also don't think Coral is in right position to gate such usage. So I am leaning towards let such case pass through (which means a check like `fNames.size > 0 ` is necessary since we see `struct<>` as a legit type). 

I am also OK to add a precondition check to fail the translation. But again, leaning towards the former option.

autumnust:
> FYI, enhanced i-test succeeded after fixing the type mismatch issue, we can merge it once all the comments are addressed.

I addressed the RowType comparison comment. The remaining discussion is mostly on how does Coral deal with corner cases in `struct` representation. Let's get some agreement on that, if that's needed before merging.

coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
ljfgem: +1 to the current approach

wmoustafa: Discussed offline. The objective is to match the Trino schema. We should remove the check from here since it is not used there.

wmoustafa (approved):

PR #193, opened 2021-11-10
Coral-Spark: Build awareness in Coral for coalsce_struct UDF

autumnust:
This patch includes change to swap `extract_union` UDF with `coalesce_struct` UDF in Coral-Spark during Coral `RelNode` -> Spark `RelNode`. 

In https://github.com/linkedin/coral/pull/192, we have changed the interpretation of union field in IR to a struct that aligns with Trino's after-explosion schema. The output struct which contains `tag` field and names each member field as `fieldN` is backward incompatible with the schema from `extract_union` UDF. This patch is need to bridge the schema difference between both. 

--- 

Note that `coalesce_struct` UDF will be statically registered.

### DIFF
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/CoalesceStructUtility.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/CoalesceStructUtility.java
@@ -60,7 +60,7 @@ public class CoalesceStructUtility {
    *
    * Further details:
    * struct_tr looks like:
-   * struct&lt;tag:tinyint, field0:type0, field1:type1, ... fieldN:typeN&gt;
+   * struct&lt;tag:int, field0:type0, field1:type1, ... fieldN:typeN&gt;
    *
    * struct_ex looks like:
    * struct&lt;tag_0:type0, tag_1:type1, ... tag_N:typeN&gt;
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/CoalesceStructUtility.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/CoalesceStructUtility.java
@@ -43,6 +43,7 @@ public class CoalesceStructUtility {
       return opBinding.getOperandType(0).getFieldList().get(ordinal).getType();
     }
   };
+
   /**
    * Represents the return type for the coalesce_struct UDF that is built for bridging the schema difference
    * between extract_union UDF's processed schema of union field in Coral IR (let's call it struct_ex) and
@@ -92,7 +93,7 @@ public class CoalesceStructUtility {
   }
 
   /**
-   * Converting a {@link RelDataType} that could potentially contains a Trino-format exploded-union(i.e. a struct
+   * Converting a {@link RelDataType} that could potentially contain a Trino-format exploded-union(i.e. a struct
    * in a format of {tag, field0, field1, ..., fieldN} to represent a union after being deserialized)
    * into a exploded-union that complies with Hive's extract_union UDF format
    * (i.e. a struct as {tag_0, tag_1, ..., tag_{N}} to represent a union after being deserialized)
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/CoalesceStructUtility.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/CoalesceStructUtility.java
@@ -3,18 +3,17 @@
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
-
 package com.linkedin.coral.hive.hive2rel.functions;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import com.google.common.annotations.VisibleForTesting;
+
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rel.type.RelDataTypeField;
-
-import com.google.common.annotations.VisibleForTesting;
 
 
 /**
@@ -30,7 +29,6 @@ public class CoalesceStructUtility {
   private CoalesceStructUtility() {
     // Utility class, does nothing in constructor
   }
-
 
   /**
    * Converting a {@link RelDataType} that could potentially contains a Trino-format exploded-union(i.e. a struct
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/CoalesceStructUtility.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/CoalesceStructUtility.java
@@ -10,10 +10,12 @@ import java.util.List;
 import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.sql.type.SqlReturnTypeInference;
 
 
 /**
@@ -23,8 +25,67 @@ import org.apache.calcite.rel.type.RelDataTypeField;
  */
 public class CoalesceStructUtility {
 
-  private static final String NEW_PREFIX = "field";
-  private static final String OLD_PREFIX = "tag_";
+  /**
+   * The semantics for the extract_union is now pass-through: Assuming the engine's reader could deal with
+   * union type and explode it into a struct, this extract_union UDF's return type will simply follow exploded struct's
+   * schema based on how many arguments passed by users.
+   */
+  public static final SqlReturnTypeInference EXTRACT_UNION_FUNCTION_RETURN_STRATEGY = opBinding -> {
+    int numArgs = opBinding.getOperandCount();
+    Preconditions.checkState(numArgs == 1 || numArgs == 2);
+    // 1-arg case
+    if (numArgs == 1) {
+      return opBinding.getOperandType(0);
+    }
+    // 2-arg case
+    else {
+      int ordinal = opBinding.getOperandLiteralValue(1, Integer.class);
+      return opBinding.getOperandType(0).getFieldList().get(ordinal).getType();
+    }
+  };
+  /**
+   * Represents the return type for the coalesce_struct UDF that is built for bridging the schema difference
+   * between extract_union UDF's processed schema of union field in Coral IR (let's call it struct_ex) and
+   * Trino's schema when deserializing union field from its reader.
+   * (Let's call it struct_tr, See https://github.com/trinodb/trino/pull/3483 for details).
+   *
+   * The main reason we need this briding capability is that we have existing users relying on the
+   * schema of struct_ex. While the underlying reader(e.g. the trino one referenced above) starts to interpret the union
+   * in its own format, Coral tries to maintain backward compatibility on top of that. Notably we also have
+   * Iceberg reader does the same, see Linkedin's (temporary) fork on Iceberg:
+   * https://github.com/linkedin/iceberg/pull/84 (Avro)
+   * https://github.com/linkedin/iceberg/pull/85 (ORC)
+   *
+   *
+   * Further details:
+   * struct_tr looks like:
+   * struct&lt;tag:int, field0:type0, field1:type1, ... fieldN:typeN&gt;
+   *
+   * struct_ex looks like:
+   * struct&lt;tag_0:type0, tag_1:type1, ... tag_N:typeN&gt;
+   *
+   * This new UDF could be stated as the following signatures:
+   * def coalesce_struct(struct:struct_tr) : struct_ex = {...}
+   * def coalesce_struct(struct:struct_tr, ordinal: int): field_at_ordinal = {...}
+   *
+   */
+  public static final SqlReturnTypeInference COALESCE_STRUCT_FUNCTION_RETURN_STRATEGY = opBinding -> {
+    int numArgs = opBinding.getOperandCount();
+    RelDataTypeFactory typeFactory = opBinding.getTypeFactory();
+    Preconditions.checkState(numArgs == 1 || numArgs == 2);
+    RelDataType coalescedDataType = coalesce(opBinding.getOperandType(0), typeFactory);
+    // 1-arg case
+    if (numArgs == 1) {
+      return coalescedDataType;
+    }
+    // 2-arg case
+    else {
+      int ordinal = opBinding.getOperandLiteralValue(1, Integer.class);
+      return coalescedDataType.getFieldList().get(ordinal).getType();
+    }
+  };
+  private static final String TRINO_PREFIX = "field";
+  private static final String HIVE_EXTRACT_UNION_PREFIX = "tag_";
 
   private CoalesceStructUtility() {
     // Utility class, does nothing in constructor
@@ -41,7 +102,7 @@ public class CoalesceStructUtility {
   @VisibleForTesting
   static RelDataType coalesce(RelDataType inputNode, RelDataTypeFactory typeFactory) {
     // Using type information implicitly carried in the object of RelDateType
-    // instead of get down to SqlTypeName since the former contains enough categorization
+    // instead of getting down to SqlTypeName since the former contains enough categorization
     // of types to achieve the purpose for this method.
 
     if (inputNode.isStruct()) {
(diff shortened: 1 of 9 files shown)


wmoustafa:
Should the coalesce_struct UDF be part of Coral?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveReturnTypes.java
```diff
@@ -61,6 +67,40 @@ private HiveReturnTypes() {
...
+   * extract_union Hive UDF, let's call it struct_old) and Trino's schema when deserializing union field from its reader.
+   * (Let's call it struct_new, See https://github.com/trinodb/trino/pull/3483 for details).
+   *
+   * struct_new looks like:
+   * struct&lt;tag:int, field0:type0, field1:type1, ... fieldN:typeN&gt;
+   *
+   * struct_old looks like:
+   * struct&lt;tag_0:type0, tag_1:type1, ... tag_N:typeN&gt;
+   *
+   * This new UDF could be stated as the following signature:
```
wmoustafa, line 82: `The `coalesce_struct` function signature can be expressed as follows:`
Also I would separate to two signatures.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveReturnTypes.java
```diff
@@ -61,6 +67,40 @@ private HiveReturnTypes() {
...
+   * def coalesce_struct(struct:struct_new, [ordinal: int]) : struct_old = {}
+   *
+   */
+  public static final SqlReturnTypeInference COALESCE_STRUCT_FUNCTION_RETURN_STRATEGY = opBinding -> {
+    int numArgs = opBinding.getOperandCount();
+    RelDataTypeFactory typeFactory = opBinding.getTypeFactory();
+    Preconditions.checkState(numArgs == 1 || numArgs == 2);
+    // 1-arg case
+    if (numArgs == 1) {
+      List<RelDataTypeField> oldStructFieldList = opBinding.getOperandType(0).getFieldList();
```
wmoustafa, line 92: old --> input?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveReturnTypes.java
```diff
@@ -61,6 +67,40 @@ private HiveReturnTypes() {
...
+      List<RelDataType> types = oldStructFieldList.stream().map(RelDataTypeField::getType).collect(Collectors.toList());
+      List<String> names = oldStructFieldList.stream().map(RelDataTypeField::getName).collect(Collectors.toList());
+      return typeFactory.createStructType(types.subList(1, types.size()), names.subList(1, names.size()));
+    }
+    // 2-arg case
+    else {
+      int ordinal = opBinding.getOperandLiteralValue(1, Integer.class);
+      return opBinding.getOperandType(0).getFieldList().get(ordinal).getType();
+    }
```
wmoustafa, line 101: The correct conversion is not implemented yet, right?

wmoustafa (commented):
Should both Spark and Trino be handled in the same PR?

autumnust:
> Should the coalesce_struct UDF be part of Coral?
Why is that if that is going to be a Hive UDF, or do you think it should be statically defined in Coral and being loaded into engines when being used ?

autumnust:
> Should both Spark and Trino be handled in the same PR?

Before the change, only Spark IR Transformer (during IR -> Spark RelNode) handles the `extract_union` UDF (by getting rid of it). Trino path simply cannot execute SQL statements with `extract_union` as part of it. Do we want to preserve that behavior or you believe we should also get rid of `extract_union` in the trino SQL whenever appearing ?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveReturnTypes.java
autumnust: In terms of return type inference this is the implementation. Do you mean the UDF itself ?

wmoustafa: I expect the UDF should convert structs from Trino flavor to structs of extract_union flavor, recursively?

wmoustafa:
> > Should the coalesce_struct UDF be part of Coral?
> 
> Why is that if that is going to be a Hive UDF, or do you think it should be statically defined in Coral and being loaded into engines when being used ?

I meant the GenericUDF definition, yes. But, I think it is okay if the function is not part of Coral.

wmoustafa:
> > Should both Spark and Trino be handled in the same PR?
> 
> Before the change, only Spark IR Transformer (during IR -> Spark RelNode) handles the `extract_union` UDF (by getting rid of it). Trino path simply cannot execute SQL statements with `extract_union` as part of it. Do we want to preserve that behavior or you believe we should also get rid of `extract_union` in the trino SQL whenever appearing ?

Ok, makes sense. We can restrict `extract_union` usage to Spark.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveReturnTypes.java
autumnust: that's right. I will have to add the recursion part for the 1-arg case.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveReturnTypes.java
autumnust: Remove the whole part. This isn't applicable any more.

autumnust:
@wmoustafa  Can you take a second look ? thanks !

coral-spark/src/test/java/com/linkedin/coral/spark/TestUtils.java
```diff
@@ -193,6 +193,9 @@ public static void initializeViews() throws HiveException, MetaException {
 
     run(driver,
         "CREATE TABLE IF NOT EXISTS union_table(foo uniontype<int, double, array<string>, struct<a:int,b:string>>)");
+
+    // Recursive union case
+    run(driver, "CREATE TABLE IF NOT EXISTS recursive_union(a uniontype<int, uniontype<int, double>>)");
```
funcheetah, line 198: Could you please add test case for `uniontype<struct<uniontype>>`? We found this case is error prone and better to have unit test to guard it.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
```diff
@@ -376,6 +376,8 @@ public RelDataType inferReturnType(SqlOperatorBinding opBinding) {
     createAddUserDefinedFunction("sort_array", ARG0, ARRAY);
     createAddUserDefinedFunction("extract_union", HiveReturnTypes.EXTRACT_UNION_FUNCTION_RETURN_STRATEGY,
         or(ANY, family(SqlTypeFamily.ANY, SqlTypeFamily.INTEGER)));
+    createAddUserDefinedFunction("coalesce_struct", HiveReturnTypes.COALESCE_STRUCT_FUNCTION_RETURN_STRATEGY,
+        or(ANY, family(SqlTypeFamily.ANY, SqlTypeFamily.INTEGER)));
```
ljfgem, line 380: Given this method is registered, I think we'd better add the tests to verify the converted RelNode's type of SQLs like `SELECT coalesce_struct(a) from recursive_union`. That will give us more confidence than `CoalesceStructUtilityTest` only.
But yeah, it needs #192 to be checked in first.

coral-spark/src/test/java/com/linkedin/coral/spark/CoralSparkTest.java
```diff
@@ -382,11 +383,16 @@ public void testSchemaPromotionView() {
...
-    RelNode relNode2 = TestUtils.toRelNode("SELECT extract_union(foo, 2) from union_table");
-    String targetSql2 = String.join("\n", "SELECT foo.tag_2", "FROM default.union_table");
+    RelNode relNode1 = TestUtils.toRelNode("SELECT extract_union(foo, 2) from union_table");
+    String targetSql1 = String.join("\n", "SELECT coalesce_struct(foo, 3)", "FROM default.union_table");
+    assertEquals(CoralSpark.create(relNode1).getSparkSql(), targetSql1);
+
+    // Recursion case
+    RelNode relNode2 = TestUtils.toRelNode("SELECT extract_union(a) from recursive_union");
+    String targetSql2 = String.join("\n", "SELECT coalesce_struct(a)", "FROM default.recursive_union");
     assertEquals(CoralSpark.create(relNode2).getSparkSql(), targetSql2);
```
ljfgem, line 396: I don't think the recursion case here is necessary given it does not verify the type of translated SQL's relNode, previous 2 test cases should be fine.
I think we need to test the type of RelNode from SQL like `SELECT coalesce_struct(a) FROM default.recursive_union` where `a` is nested union as mentioned above.

coral-spark/src/test/java/com/linkedin/coral/spark/TestUtils.java
ljfgem: +1, and I think `nested_union` is better than `recursion_union`?

ljfgem (commented):
Thanks @autumnust for this PR!

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/CoalesceStructUtility.java
```diff
@@ -0,0 +1,111 @@
...
+      boolean flag = true;
+      StringBuilder fieldNameRef = new StringBuilder("field");
+      for (int i = 1; i < fieldNames.size(); i++) {
+        int index = i - 1;
+        fieldNameRef.append(index);
+        if (!fieldNameRef.toString().equals(fieldNames.get(i))) {
+          flag = false;
+          break;
+        }
+        fieldNameRef.deleteCharAt(5);
```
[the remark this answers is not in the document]
ljfgem: And could you add the corresponding unit test for this case?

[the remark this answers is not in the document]
autumnust: Good catch ... added unit test for that.

autumnust:
Looks like the Github action is failing with weird reason: 

```/home/runner/work/coral/coral/coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/functions/CoalesceStructUtilityTest.java:36: error: method setup in class ToRelConverterTestUtils cannot be applied to given types;

> Task :coral-hive:compileTestJava FAILED
    ToRelConverterTestUtils.setup();
                           ^
  required: HiveConf
  found: no arguments
  reason: actual and formal argument lists differ in length
Note: /home/runner/work/coral/coral/coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilderTest.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
1 error```

@ljfgem  have you seen this ? this failure couldn't e reproduced locally and the error message doesn't seem to make sense since there's only one `setup` method without no arguments. Binding issue ?

**Comment by wmoustafa** [comment]
> @ljfgem have you seen this ? this failure couldn't be reproduced locally and the error message doesn't seem to make sense since there's only one `setup` method without no arguments. Binding issue ?

You probably need to rebase. This method has changed in https://github.com/linkedin/coral/pull/198.

File: coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/functions/CoalesceStructUtilityTest.java [hunk 8]
```diff
@@ -0,0 +1,111 @@
...
+
+public class CoalesceStructUtilityTest {
+  private RelDataTypeFactory typeFactory;
+  private RelDataType trinoStruct;
+  private RelDataType extractUnionStruct;
+  private RelDataType nonTrinoStruct;
+
+  @BeforeClass
+  public void setup() throws Exception {
+    ToRelConverterTestUtils.setup();
```
ljfgem, line 36: Given #198 has been checked in, we need to modify this part, like:
https://github.com/linkedin/coral/blob/master/coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/FuzzyUnionTest.java#L29-L40

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/functions/CoalesceStructUtilityTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+public class CoalesceStructUtilityTest {
+  private RelDataTypeFactory typeFactory;
+  private RelDataType trinoStruct;
+  private RelDataType extractUnionStruct;
+  private RelDataType nonTrinoStruct;
+
+  @BeforeClass
+  public void setup() throws Exception {
+    ToRelConverterTestUtils.setup(TestUtils.loadResourceHiveConf());
```
ljfgem, line 37: We need to have a class variable like `HiveConf conf` and instantiate it here (`conf = TestUtils.loadResourceHiveConf()`), and add a `afterClass` method to clean the workspace after all the tests in this class. 
i.e. https://github.com/linkedin/coral/blob/master/coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/FuzzyUnionTest.java#L29-L40

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/functions/CoalesceStructUtilityTest.java
```diff
@@ -0,0 +1,112 @@
...
+    typeFactory = ToRelConverterTestUtils.createRelBuilder().getTypeFactory();
+
+    List<String> names = ImmutableList.of("tag", "field0", "field1");
+    List<RelDataType> types = ImmutableList.of(typeFactory.createSqlType(SqlTypeName.INTEGER),
```
ljfgem, line 41: I think it should be `SqlTypeName.TINYINT`, although it wouldn't affect the test.

ljfgem (commented):
Thanks @autumnust for this patch!

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/CoalesceStructUtility.java
```diff
@@ -0,0 +1,173 @@
...
+      List<String> fieldNames = inputNode.getFieldNames();
+      return coalesceStruct(inputNode, isTrinoStructPattern(fieldNames), typeFactory);
+    } else if (inputNode.getKeyType() != null) {
+      return coalesceMap(inputNode, typeFactory);
+    } else if (inputNode.getComponentType() != null) {
+      return coalesceCollection(inputNode, typeFactory);
+    } else {
+      return inputNode;
+    }
+  }
```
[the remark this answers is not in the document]
autumnust: I was actually looking into that. However the enum `SqlTypeName` isn't quite clear to me (for example: there could be multiple members mapping to struct (`Row` and `Structured`)). 

I also looked into `com.linkedin.coral.common.FuzzyUnionSqlRewriter#getUnionDataType(java.util.List<org.apache.calcite.rel.type.RelDataType>)` and looks like is also uses `isStruct` to determine struct type. Given that I would prefer keeping this way since it is even more confusing to have one if-branch plus a switch statement.

coral-spark/src/main/java/com/linkedin/coral/spark/IRRelToSparkRelTransformer.java
```diff
@@ -323,16 +325,34 @@ public RexNode visitCall(RexCall call) {
...
         else if (call.getOperands().size() == 2) {
-          int ordinal = ((RexLiteral) call.getOperands().get(1)).getValueAs(Integer.class);
-          return Optional.of(rexBuilder.makeFieldAccess(call.getOperands().get(0), ordinal));
+          int ordinal = ((RexLiteral) call.getOperands().get(1)).getValueAs(Integer.class) + 1;
+          List<RexNode> operandsCopy = Lists.newArrayList(call.getOperands());
+          operandsCopy.set(1, rexBuilder.makeExactLiteral(new BigDecimal(ordinal)));
+          return Optional.of(rexBuilder.makeCall(
+              createUDF("coalesce_struct", CoalesceStructUtility.COALESCE_STRUCT_FUNCTION_RETURN_STRATEGY),
+              operandsCopy));
         }
```
ljfgem, line 356: If we want to add `coalesce_struct` in coral, I think we could add the corresponding UDF info here, like:
```
sparkUDFInfos.add(new SparkUDFInfo("<class_name>", "coalesce_struct", Collections.singletonList(URI.create("<ivy_link>")), SparkUDFInfo.UDFTYPE.HIVE_CUSTOM_UDF);
```

autumnust: Thanks for the pointers. I will go with that route in a separate PR once we decided where to put the UDF.

ljfgem (approved):
LGTM, thanks 
@wmoustafa @funcheetah PTAL if you get a chance.

funcheetah (approved):
Thanks @autumnust ! LGTM

PR #196, opened 2021-11-15
coral-schema: Generate documentation for function calls

sherryhli:
### Summary

Previously, field-level documentation in schemas produced by the `ViewToAvroSchemaConverter` was only available for fields that were:
* directly selected from base tables (#115)
* created from literal values (#172)
* created from aggregate function calls (#186)

This PR adds documentation for fields created from other types of function calls (i.e. UDFs and operators).

The documentation is in the form of: `"Field created in view by applying UDF / operator '<FUNCTION-NAME>' with argument(s): <LIST-OF-ARGUMENTS>"`

### Testing
There are existing unit tests covering function calls. The `.avsc` files containing expected schemas for those test cases have been updated to include the documentation field.

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
@@ -12,6 +12,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.StringJoiner;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -22,12 +23,16 @@ import com.google.common.annotations.VisibleForTesting;
 
 import org.apache.avro.Schema;
 import org.apache.avro.SchemaBuilder;
+import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.core.AggregateCall;
+import org.apache.calcite.rel.logical.LogicalAggregate;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexInputRef;
 import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.validate.SqlUserDefinedFunction;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.hive.metastore.api.FieldSchema;
 import org.apache.hadoop.hive.metastore.api.Table;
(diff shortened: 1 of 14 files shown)


coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -273,20 +278,63 @@ static String getFieldName(String oldName, String suggestedNewName) {
...
+   * Given an input {@link RelNode} and the index of a field in the {@link RelNode}'s corresponding Avro schema,
+   * determine if the field with the specified index is a column from a table.
+   * @param fieldIndex the index of a field in the <code>inputRelNode</code>'s corresponding Avro schema
+   * @param inputRelNode the input {@link RelNode}
+   * @return true if the field at <code>fieldIndex</code> is a column from a table
+   */
+  private static boolean isColumn(int fieldIndex, @Nonnull RelNode inputRelNode) {
+    return !(inputRelNode instanceof LogicalAggregate)
+        || fieldIndex < ((LogicalAggregate) inputRelNode).getGroupSet().cardinality();
+  }
```
sherryhli, line 301: Given an input `RelNode` and a field index, how can we generalize the logic for determining whether or not the field is referring to a column from a table? I don't think this current approach is very robust.

@funcheetah @ljfgem wondering if you have any suggestions? 🙂

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -273,20 +278,63 @@ static String getFieldName(String oldName, String suggestedNewName) {
...
+
+    for (RexNode rexNode : rexCall.getOperands()) {
+      SqlKind nodeKind = rexNode.getKind();
+      switch (nodeKind) {
+        case LITERAL:
+          args.add(getLiteralValueAsString((RexLiteral) rexNode));
+          break;
+        case INPUT_REF:
+          int fieldIndex = ((RexInputRef) rexNode).getIndex();
+          if (isColumn(fieldIndex, inputRelNode)) {
```
sherryhli, line 323: To provide more context: we only want to extract the field name if the field refers to a column, otherwise the name is not meaningful (they could be auto-assigned names such as `EXPR_3` or `_f0`).

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
ljfgem: I would prefer to determine it by:
```
return !inputRelNode.getRowType().getFieldNames().get(fieldIndex).contains("$");
```
`$` wouldn't appear in a normal column name, but would appear in the auto-assigned column name in the row type.

sherryhli: Thank you, I'll give this a try!

sherryhli: @ljfgem This seems to work for all current test cases, but it looks like from Hive 0.13 onwards, `$` can be a valid character in a column name.

From Apache Hive [docs](https://cwiki.apache.org/confluence/display/hive/languagemanual+ddl#LanguageManualDDL-CreateTable):
>* In Hive 0.12 and earlier, only alphanumeric and underscore characters are allowed in table and column names.
>* In Hive 0.13 and later, column names can contain any Unicode character (see HIVE-6013), however, dot (.) and colon (:) yield errors on querying, so they are disallowed in Hive 1.2.0 (see HIVE-10120). Any column name that is specified within backticks (`) is treated literally. Within a backtick string, use double backticks (``) to represent a backtick character. Backtick quotation also enables the use of reserved keywords for table and column identifiers.

ljfgem: Hmmm, then I am good with the current logic for determining it.

sherryhli: Sounds good. Are there any special cases other than `LogicalAggregate` that I should account for?

ljfgem: Nothing on top of my mind at present, we can add if we encounter a corner case in the future.

ljfgem (approved):
LGTM, thanks @sherryhli for this great work!
i-test passed.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
funcheetah: I'm okay with this approach to determine whether it is a column.

funcheetah (approved):
Nice work! Thanks for your contribution to Coral @sherryhli !

PR #198, opened 2021-11-21
Create special workspace for each test class and clean it after test

ljfgem:
### Summary
Previously, the workspaces (something like `/var/folders/wy/lgjfncg538j1symjt4cbxcxw001d0t/T/coral`) are not cleaned after the tests, and once there is an backward incompatible change in the setup schema, following build will fail with the issue like:
```
com.linkedin.coral.sparkplan.SparkPlanToIRRelConverterTest.beforeClass FAILED
    java.lang.RuntimeException at SparkPlanToIRRelConverterTest.java:29
        Caused by: java.lang.RuntimeException at SparkPlanToIRRelConverterTest.java:29
            Caused by: java.lang.reflect.InvocationTargetException at SparkPlanToIRRelConverterTest.java:29
                Caused by: javax.jdo.JDOFatalDataStoreException at SparkPlanToIRRelConverterTest.java:29
                    Caused by: java.sql.SQLException at SparkPlanToIRRelConverterTest.java:29
                        Caused by: java.sql.SQLException at SparkPlanToIRRelConverterTest.java:29
                            Caused by: org.apache.derby.impl.jdbc.EmbedSQLException at SparkPlanToIRRelConverterTest.java:29
                                Caused by: org.apache.derby.impl.jdbc.EmbedSQLException at SparkPlanToIRRelConverterTest.java:29
                                    Caused by: org.apache.derby.iapi.error.StandardException at SparkPlanToIRRelConverterTest.java:29

4 tests completed, 1 failed, 3 skipped
```
To solve this issue, we need to delete the workspace manually, which is not friendly to developers especially new contributors, and we have received many complaints about it.

With this PR, for each test class, a corresponding test workspace will be created before running the tests, and it will be cleaned once all the tests in the class finish. 
i.e. for test class `CoralSparkTest`, a workspace `/var/folders/wy/lgjfncg538j1symjt4cbxcxw001d0t/T/coral/spark/<random uuid>` will be created just for it, and will be deleted after all the tests in `CoralSparkTest` finish.

We need to use `uuid` to create different workspaces for different test classes, because if we just create one workspace for many classes, it might be deleted by one finished test class while another unfinished test class still needs it.
### Tests
1. `./gradlew clean build`
2. i-test


coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/TestUtils.java
```diff
@@ -64,12 +68,11 @@ public IMetaStoreClient getMetastoreClient() throws HiveException, MetaException
     }
   }
 
-  public static TestHive setupDefaultHive() throws IOException {
+  public static TestHive setupDefaultHive(HiveConf conf) throws IOException {
     if (hive != null) {
       return hive;
     }
-    System.out.println(System.getProperty("java.io.tmpdir"));
```
funcheetah, line 71: Could we keep similar log in new implementation in case a manually look into the directory is needed?

funcheetah (approved):
Awesome! This is a long pending issue which causes confusion to Coral developers. Thanks @ljfgem

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/TestUtils.java
ljfgem: Thanks for reminding, added back

PR #200, opened 2021-11-24
Coral-Trino: Convert the name of unregistered UDF to lowercase underscore format

ljfgem:
### Summary
Previously, for a UDF not registered in `CalciteTrinoUDFMap#UDF_MAP`, we would return the full name like `com.linkedin.jemslookup.udf.hive.UrnArrayToIdStringConverter` in translated Trino SQL, which will definitely cause the `UDF not registered` issue while querying the view on Trino, given we register the Trino UDFs in the lowercase underscore format. This PR is to convert such the full name to the lowercase underscore format to increase the query success rate of the translated Trino SQL.
i.e. `com.linkedin.jemslookup.udf.hive.UrnArrayToIdStringConverter -> urn_array_to_id_string_converter`

### Tests
1. Enhanced i-test
2. Test on affected production views


funcheetah (approved):
Thanks @ljfgem . LGTM

PR #202, opened 2021-12-06
Coral-Common: Using INTEGER for trino-struct tag field

autumnust:
Based on previous PR https://github.com/linkedin/coral/pull/192/files, this PR makes the modification in the data type of `tag` field due to compatibility with other platforms like Avro and Iceberg in which `tinyint` is not supported.

### DIFF
--- a/coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
+++ b/coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
@@ -138,12 +138,17 @@ public class TypeConverter {
 
   // Mimic the StructTypeInfo conversion to convert a UnionTypeInfo to the corresponding RelDataType
   // The schema of output Struct conforms to https://github.com/trinodb/trino/pull/3483
+  // except we adopted "integer" for the type of "tag" field instead of "tinyint" in the Trino patch
+  // for compatibility with other platforms that Iceberg at currently doesn't support tinyint type.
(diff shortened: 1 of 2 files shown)


coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
```diff
@@ -138,12 +138,17 @@ public static RelDataType convert(StructTypeInfo structType, final RelDataTypeFa
 
   // Mimic the StructTypeInfo conversion to convert a UnionTypeInfo to the corresponding RelDataType
   // The schema of output Struct conforms to https://github.com/trinodb/trino/pull/3483
+  // except we adopted "integer" for the type of "tag" field instead of "tinyint" in the Trino patch
+  // for compatibility with other platforms that Iceberg at currently doesn't support tinyint type.
```
ljfgem, line 142: nit: `that Iceberg currently doesn't support tinyint type.`?

ljfgem (approved):
LGTM, left one nit.
Thanks @autumnust

funcheetah (approved):

coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
autumnust: Addressed, thanks for catching

PR #204, opened 2021-12-08
Coral-Hive: Modify the return type of IF function

ljfgem:
### Summary
Previously, the return type of `if(..., a, b)` is the first non-null type of a and b, therefore, the return type of `if(..., 0, bigint_type_field) as bigint_type_field` is `int` rather than `bigint`, which will cause the following issue while querying the view on Trino:
```
View 'xxx' is stale or in invalid state: column [bigint_type_field] of type bigint projected from query view 
at position x cannot be coerced to column [bigint_type_field] of type integer stored in view definition
```
This PR will ensure that the return type of `if(..., 0, bigint_type_field) as bigint_type_field` is `bigint` rather than `int`.

We pick the non-literal one if the types mismatch because the non-literal one (which is usually a field of table) is picked by Trino and we need to ensure the type picked by Trino is the same as the type returned by coral.

### Test
1. Unit test
2. Enhanced i-test, this issue was caught by the enhanced i-test to begin with
3. Test on affected views

### DIFF
--- a/coral-common/src/main/java/com/linkedin/coral/common/functions/FunctionReturnTypes.java
+++ b/coral-common/src/main/java/com/linkedin/coral/common/functions/FunctionReturnTypes.java
@@ -28,13 +28,16 @@ public final class FunctionReturnTypes {
 
   }
 
-  public static final SqlReturnTypeInference ARG1_OR_ARG2 = opBinding -> {
+  public static final SqlReturnTypeInference IF_FUNC_RETURN_TYPE = opBinding -> {
     Preconditions.checkState(opBinding.getOperandCount() == 3);
-    if (!opBinding.isOperandNull(1, false)) {
-      return opBinding.getOperandType(1);
-    } else {
-      return opBinding.getOperandType(2);
+    final RelDataType type1 = opBinding.getOperandType(1);
+    final RelDataType type2 = opBinding.getOperandType(2);
+    if (type1 == type2) {
+      return type1;
     }
+    // If the types mismatch, the non-literal one's type should be picked
+    // i.e. the type of `if(..., 0, bigint_type_field)` should be bigint rather than int
+    return opBinding.isOperandLiteral(1, false) ? type2 : type1;
   };
 
   public static final SqlReturnTypeInference STRING = ReturnTypes.explicit(SqlTypeName.VARCHAR);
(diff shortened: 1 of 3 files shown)


coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
```diff
@@ -129,7 +129,7 @@
     // to SQL to be odd although correct. So, we add 'if' as UDF
     // TODO: add check to verify 2nd and 3rd operands are same
```
autumnust, line 130: This comment can be removed now ?

coral-common/src/main/java/com/linkedin/coral/common/functions/FunctionReturnTypes.java
```diff
@@ -28,13 +28,16 @@ private FunctionReturnTypes() {
...
-      return opBinding.getOperandType(1);
-    } else {
-      return opBinding.getOperandType(2);
+    final RelDataType type1 = opBinding.getOperandType(1);
+    final RelDataType type2 = opBinding.getOperandType(2);
+    if (type1 == type2) {
+      return type1;
     }
+    // If the types mismatch, the non-literal one's type should be picked
+    // i.e. the type of `if(..., 0, bigint_type_field)` should be bigint rather than int
```
autumnust, line 39: Can you point me a reference about `if` grammar? Looks like it is used to replace `case` statement but the grammar of `case` seems to be a bit different though ...

autumnust (changes requested):

coral-common/src/main/java/com/linkedin/coral/common/functions/FunctionReturnTypes.java
ljfgem: The grammar can be found [here](https://cwiki.apache.org/confluence/display/hive/languagemanual+udf#LanguageManualUDF-ConditionalFunctions)
But it seems that in Hive, the second and third type don't need to be exactly same, SQL like `select if(true, 1, 1.0);` works well, and there are many mismatching types in our production views.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
ljfgem: Thanks for catching! Will remove it.

coral-common/src/main/java/com/linkedin/coral/common/functions/FunctionReturnTypes.java
autumnust: can you elaborate more on why we are pick non-literal type as part of comments?

autumnust (approved):
LGTM other than one small ask

coral-common/src/main/java/com/linkedin/coral/common/functions/FunctionReturnTypes.java
ljfgem: Thanks for reminding, because the non-literal one (which is usually a field of table) is picked by Trino and we need to ensure the type picked by Trino is the same as the type returned by coral.

Also added it in the PR summary.

PR #208, opened 2021-12-10
Coral-Spark | Coral-Hive : Statically register coalesce_struct UDF in Coral-Spark

autumnust:
This PR statically register `coalesce_struct` UDF in Coral-Spark. The jar containing UDF will be downloaded in `org.apache.spark.sql.hive.HiveExternalCatalog#restoreTableMetadata` where the `view` is handled. 

~~This patch is created for early feedback, the integration test within properly environment is in-progress~~ 

This PR additionally contains an one-liner fix for the issue we discovered during the tests. Briefly, given the decision to swap `extract_union` to `coalesce_struct` to preserve the original output schema of `extract_union` HiveUDF, we need to ensure the `RelDataType` of a node in AST `extract_union(union)` to be its original output schema, which is `[tag_0, tag_1, ... , tag_N]`. To achieve that, the registration of `extract_union` needs to be modified to return `COALESCE_STRUCT_FUNCTION_RETURN_STRATEGY` instead of `EXTRACT_UNION_FUNCTION_RETURN_STRATEGY`.

### DIFF
--- a/coral-spark/src/main/java/com/linkedin/coral/spark/IRRelToSparkRelTransformer.java
+++ b/coral-spark/src/main/java/com/linkedin/coral/spark/IRRelToSparkRelTransformer.java
@@ -341,6 +341,12 @@ class IRRelToSparkRelTransformer {
      */
     private Optional<RexNode> swapExtractUnionFunction(RexCall call) {
       if (call.getOperator().getName().equalsIgnoreCase("extract_union")) {
+        // When there's necessity to register coalesce_struct UDF
+        sparkUDFInfos.add(new SparkUDFInfo("com.linkedin.coalescestruct.GenericUDFCoalesceStruct",
+            "coalesce_struct",
+            ImmutableList.of(URI.create("ivy://com.linkedin.coalesce-struct:coalesce-struct-impl:0.0.1")),
+            SparkUDFInfo.UDFTYPE.HIVE_CUSTOM_UDF));
+
         // one arg case: extract_union(field_name)
         if (call.getOperands().size() == 1) {
           return Optional.of(rexBuilder.makeCall(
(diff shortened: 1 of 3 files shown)


autumnust:
@ljfgem  I hope this approach looks reasonable based on our previous discussion ?

ljfgem:
@autumnust Yeah, it looks good to me.

coral-spark/src/main/java/com/linkedin/coral/spark/IRRelToSparkRelTransformer.java
```diff
@@ -339,6 +339,12 @@ public RexNode visitCall(RexCall call) {
      */
     private Optional<RexNode> swapExtractUnionFunction(RexCall call) {
       if (call.getOperator().getName().equalsIgnoreCase("extract_union")) {
+        // Only when there's a necessity to register coalesce_struct UDF
```
rzhang10, line 342: If there are multiple `extract_union` occurrences in the SQL query, the `sparkUDFInfos` list will contain multiple copies of the same `SparkUDFInfo`, will this potential redundancy cause any problems?

ljfgem: Good catch, I don't think it will cause problems for Spark, but I would also prefer to check if it exists before adding it.

rzhang10: I was thinking maybe define `sparkUDFInfos` to a `Set`, instead of `List`? But this seems to need to refactor some other areas of code too.

ljfgem: Yeah I had the same thought before, but changing it to `Set` also needs to modify the code in Spark. Another way is to keep using `List` and deduplicate before constructing `CoralSpark`

autumnust: Resolving this conversation as there's a separate PR addressing this issue. https://github.com/linkedin/coral/pull/209/files

ljfgem (approved):
LGTM, thanks!
@funcheetah @wmoustafa do you want to take another look?

funcheetah (approved):
LGTM. Good to run integration test for this change and make sure no regression.

PR #209, opened 2021-12-11
Coral-Spark: Deduplicate sparkUDFInfoList

ljfgem:
### Summary
This PR deduplicates `sparkUDFInfoList` to avoid registering the same UDF many times in Spark. Didn't choose to use `Set` in `CoralSpark` and `SparkRelInfo` directly because it needs modifications from Spark side.

### Tests
1. Unit test
2. i-test

### DIFF
--- a/coral-spark/src/main/java/com/linkedin/coral/spark/containers/SparkUDFInfo.java
+++ b/coral-spark/src/main/java/com/linkedin/coral/spark/containers/SparkUDFInfo.java
@@ -7,6 +7,7 @@ package com.linkedin.coral.spark.containers;
 
 import java.net.URI;
 import java.util.List;
+import java.util.Objects;
 
 
 /**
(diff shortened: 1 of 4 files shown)


coral-spark/src/main/java/com/linkedin/coral/spark/containers/SparkUDFInfo.java
```diff
@@ -94,4 +95,22 @@ public String toString() {
...
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    SparkUDFInfo that = (SparkUDFInfo) o;
+    return Objects.equals(className, that.className) && Objects.equals(functionName, that.functionName)
+        && Objects.equals(artifactoryUrls, that.artifactoryUrls) && udfType == that.udfType;
+  }
+
+  @Override
+  public int hashCode() {
```
autumnust, line 113: you could use IDE to generate both equal and hashCode method btw. I don't think there are problems in this implementation but just FYI.

autumnust (approved):
One comment but not blocking, LGTM

coral-spark/src/main/java/com/linkedin/coral/spark/containers/SparkUDFInfo.java
ljfgem: Yeah these 2 methods were auto-generated by IntelliJ.

PR #210, opened 2021-12-14
coral-schema: Handle NULL type schema in setupNestedNamespace

rzhang10:
This patch fixes the bug where the `setupNestedNamespace` method failed to handle `NULL` primitive type in Avro, which makes it fail to handle complex union view RelNode to Avro schema conversion in `ViewToAvroSchemaConverterTests.testComplexUnionType()`.


autumnust (approved):
LGTM.  Thanks for fixing the issue.

ljfgem (approved):
LGTM, thanks!

Issue #212, opened 2021-12-16
Incorrect schema for views with unnamed columns in the view SQL


shardulm94:
If we create a Hive view with columns that are not explicitly named in the SQL, Hive with autogenerate column names for such columns persisting the view to the Metastore. E.g. Consider the following view
```
hive> CREATE VIEW v AS SELECT TRUE, lower('STR') FROM some_table;
```
Here is the persisted information in the metastore
```
hive> DESC FORMATTED v;
OK
# col_name            	data_type           	comment

_c0                 	boolean
_c1                 	string
.
.
.
# View Information
View Original Text: 	SELECT TRUE, lower('STR') FROM some_table
View Expanded Text: 	SELECT TRUE, lower('STR') FROM `some_table`
```

Trying to retrieve view schema using `coral-schema` for this view returns `EXPR_0` and `EXPR_1` as column names, which I assume are auto generated by Calcite. I think we should respect the column names provided by Hive here i.e. `_c0` and `c1` since they have been persisted as column metadata in the metastore.

ljfgem:
Thanks @shardulm94 for reporting it, opened a PR #214 to fix it.

PR #214, opened 2021-12-16
Coral-Schema: Respect column names provided by Hive for unnamed columns

ljfgem:
### Summary
This PR is to solve #212 
For unnamed columns, we should convert the name to the one provided by Hive. i.e. `EXPR$0 -> _c0`

### Tests
1. Unit tests
2. i-test

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
@@ -202,9 +202,8 @@ public class RelToAvroSchemaConverter {
       Queue<String> suggestedFieldNames = new LinkedList<>();
 
       for (int i = 0; i < logicalProject.getChildExps().size(); ++i) {
-        final RexNode rexNode = logicalProject.getChildExps().get(i);
         final String fieldName = logicalProject.getRowType().getFieldList().get(i).getName();
-        if (rexNode instanceof RexLiteral && fieldName.toLowerCase().startsWith("expr$")) {
+        if (fieldName.startsWith("EXPR$")) {
           // we should respect the column names provided by Hive here for unnamed columns
           // i.e. EXPR$0 -> _c0
           suggestedFieldNames.offer("_c" + fieldName.substring(5));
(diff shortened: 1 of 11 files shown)


coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
```diff
@@ -200,8 +200,16 @@ public RelNode visit(LogicalProject logicalProject) {
       Schema inputSchema = schemaMap.get(logicalProject.getInput());
 
       Queue<String> suggestedFieldNames = new LinkedList<>();
-      for (RelDataTypeField field : logicalProject.getRowType().getFieldList()) {
-        suggestedFieldNames.offer(field.getName());
+
+      for (int i = 0; i < logicalProject.getChildExps().size(); ++i) {
+        final String fieldName = logicalProject.getRowType().getFieldList().get(i).getName();
+        if (fieldName.startsWith("EXPR$")) {
```
shardulm94, line 206: I am not sure if this is the best way to go about the fix.
1. It is possible that some column is really named `EXPR$...` which will be renamed to `_c0`.
2. Some systems may use a different convention to generate names for unnamed columns. E.g. If you create the same view through Spark, the schema persisted in the metastore is a different. It as an escaped version of the expression itself.
```
hive (u_smahadik)> DESC FORMATTED v6;
OK
# col_name            	data_type           	comment

true                	boolean
lower(str)          	string
```
I was wondering if we should handle this at an earlier layer when reading view metadata from Hive.

ljfgem: Thanks for your review and thoughts!
For 1, I tried on our internal Hive, we couldn't use `$` in column name while creating a table/view.
For 2, yeah it is an issue of this implementation. But to handle it at an earlier layer when reading view metadata from Hive, we might need big refactoring in coral-schema, because we only handle the base table (not view) metadata in `RelToAvroSchemaConverter` at present. Not sure if it's doable and necessary in terms of ROI.
@funcheetah @wmoustafa feel free to share your thoughts.

shardulm94: You can use `$` in column names in Hive by using the backticks notation, e.g.
```
CREATE VIEW test AS
SELECT 1 AS `EXPR$1`
FROM t
```
At this point, I don't think this fix is super urgent for us. So I would be fine if the fix took some more time but considered more cases. But I will leave it to you guys to decide the best path forward.

PR #216, opened 2021-12-17
Cast current_timestamp to timestamp

jongn:
### Summary
This PR fixes https://github.com/trinodb/trino/issues/5688

Hive's `current_timestamp` returns `TIMESTAMP` while Trino's `current_timestamp` returns `TIMESTAMP WITH TIME ZONE`, which cannot be coerced into `TIMESTAMP` type. This fix adds the cast `CAST(current_timestamp AS TIMESTAMP(3))`

If this seems like the right way to fix it, we could consider supporting current_timestamp(n) for completeness of the intermediate representation

### Testing
Update unit test in RelToTrino and HiveToTrino

Added/ran Trino unit tests locally for correctness (not equality of results)


ljfgem (approved):
LGTM, thanks @jongn!
Enhanced i-test passed.

PR #217, opened 2021-12-21
Coral-Spark: Convert back for LATERAL VIEW OUTER EXPLODE

ljfgem:
### Summary
Given Spark supports `LATERAL VIEW OUTER EXPLODE` itself, we don't need to transform `unnest(b)` to `unnest(if(b is null or cardinality(b) = 0, ARRAY(null)/MAP(null, null), b))` in https://github.com/linkedin/coral/blob/3e388ab839e199270fe81189be14fc0bf9312dce/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilder.java#L223-L243.
But we don't want to remove this part of code given it's still useful for Trino translation and any other engines which don't support `LATERAL VIEW OUTER EXPLODE`, so we can convert it back in Coral-Spark.

i.e. this PR converts
```
LATERAL VIEW OUTER EXPLODE(if(array_struct_table.as IS NOT NULL AND size(array_struct_table.as) > 0,
array_struct_table.as, ARRAY (NULL))) t0 AS exploded_results a
```
back to 
```
LATERAL VIEW OUTER EXPLODE(array_struct_table.as) t0 AS exploded_results a
```
We need to do this conversion because if `array_struct_table.as` is an array of struct, query using previous translated Spark SQL will fail with the exception like:
```
java.lang.ClassCastException: org.apache.spark.sql.types.NullType$ cannot be cast to org.apache.spark.sql.types.StructType
```

### Tests
1. Modified unit tests
2. Test on affected production views with Spark-shell
3. i-test

### DIFF
--- a/coral-spark/src/main/java/com/linkedin/coral/spark/functions/SqlLateralJoin.java
+++ b/coral-spark/src/main/java/com/linkedin/coral/spark/functions/SqlLateralJoin.java
@@ -69,16 +69,35 @@ public class SqlLateralJoin extends SqlJoin {
       final SqlWriter.Frame joinFrame = writer.startList(SqlWriter.FrameTypeEnum.JOIN);
       left.unparse(writer, leftPrec, getLeftPrec());
 
-      switch (joinType) {
-        case COMMA:
-          writer.literal("LATERAL VIEW");
-          writer.setNeedWhitespace(true);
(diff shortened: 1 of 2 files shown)


funcheetah (approved):
Thanks @ljfgem

PR #218, opened 2022-01-07
Add sonatype settings and apply spotless check

ljfgem:
According to the [doc](https://github.com/gradle-nexus/publish-plugin):
```
Important. Users registered in Sonatype after 24 February 2021 need to customize the following URLs:

nexusPublishing {
    repositories {
        sonatype {  //only for users registered in Sonatype after 24 Feb 2021
            nexusUrl.set(uri("https://s01.oss.sonatype.org/service/local/"))
            snapshotRepositoryUrl.set(uri("https://s01.oss.sonatype.org/content/repositories/snapshots/"))
        }
    }
}
```
we need to add these settings for release.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ buildscript {
     }
   }
   dependencies {
-    classpath "io.github.gradle-nexus:publish-plugin:1.0.0"
+    classpath "io.github.gradle-nexus:publish-plugin:1.1.0"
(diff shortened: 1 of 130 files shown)


rzhang10:
Thanks you Jiefan, LGTM.
A minor nit: maybe bump the gradle-nexus plugin version from 1.0.0 to 1.1.0?

PR #219, opened 2022-01-11
coral-schema: return partition column as nullable

funcheetah:
This PR makes coral schema return partition column as nullable

1. Modified unit tests
2. itests (no regression)


ljfgem (approved):
LGTM, thanks @funcheetah

Issue #205, opened 2021-12-08
Shade trino-parser dependency of coral-trino


joshthoward:
When bumping the Coral dependency version in Trino, we noticed that `coral-trino` has a new dependency on `trino-parser`.

One of our maintainers brought up the fact that the `trino-parser` is not designed to be depended on by other projects i.e. not guaranteed to maintain backward compatibility. In order to preempt future incompatibilities, would the folks in Coral consider shading the `trino-parser` dependency?

For reference, the conversation is [here](https://github.com/trinodb/trino/pull/10226).

wmoustafa:
Added a comment in the Trino PR referenced above.

joshthoward:
@shenodaguirguis Do you have an expectation for when you will be able to shade the dependency?

shenodaguirguis:
@joshthoward thanks for checking, I will work on this next week. Will share the PR here once ready

joshthoward:
@shenodaguirguis That works. Thanks again for looking into this.

shenodaguirguis:
We changed trino-parser dependency to compileOnly/Implementation in https://github.com/linkedin/coral/pull/222

PR #220, opened 2022-01-13
[WIP] Shade trino-parser dependency of coral-trino #205

shenodaguirguis:
shading trino-parser dependeny [https://github.com/linkedin/coral/issues/205]


settings.gradle
```diff
@@ -5,7 +5,8 @@ def modules = [
   'coral-spark',
   'coral-schema',
   'coral-spark-plan',
-  'coral-trino'
+  'coral-trino',
+  'shading'
```
autumnust, line 9: Shall we be more specific on what it shades, or are we foresee other shading components ?

autumnust (changes requested):

shading/coral-trino-parser/build.gradle
```diff
@@ -0,0 +1,22 @@
+import com.github.jengelman.gradle.plugins.shadow.tasks.ConfigureShadowRelocation
```
rzhang10, line 1: I'm not sure if it's a good idea to make the shaded libraries a separate module, and let the previous module depend on it (rather than just shade within the previous module directly)

But if we go this route, we need to also publish this shaded module (ref: https://imperceptiblethoughts.com/shadow/publishing/#publishing-with-maven-publish-plugin) , so that the downstream can continue to consume coral as a library.

wmoustafa:
This caused release failures due to the following error:
```Event: Failed: Signature Validation
Monday, January 24, 2022 16:58:46 PST (GMT-0800)
typeId	signature-staging
failureMessage	Missing Signature: '/com/linkedin/coral/coral-trino-parser/2.0.46/coral-trino-parser-2.0.46-all.jar.asc' does not exist for 'coral-trino-parser-2.0.46-all.jar'.
failureMessage	Missing Signature: '/com/linkedin/coral/coral-trino-parser/2.0.46/coral-trino-parser-2.0.46.pom.asc' does not exist for 'coral-trino-parser-2.0.46.pom'.```

We should not publish the `trino-parser` shaded jar and only use the classes internally I think.

Issue #170, opened 2021-10-10
Support Presto as the input SQL


wenruimeng:
We had a few offline conversations about the n-to-one-to-n proposal, which is to support any engine query to any other engine query. 

This issue is the feature quest to introduce Presto SQL as the input query and translate it into intermediate Coral representation and then it can be translated to other engines that are already supported in Coral output. 

There are a few stages for this feature:
1) Parse Presto SQL to Calcite SqlNode
2) Presto SqlNode to Presto RelNode (most work is about the function resolver)
3) Presto RelNode to Intermediate Coral RelNode
4) End to end test

PR #224, opened 2022-01-20
[Coral-Trino] Implement basic end-to-end Trino to Rel conversion workflow

KevinGe00:
Implementation of steps 2-4 from https://github.com/linkedin/coral/issues/170, completing a basic end-to-end Trino to Rel conversion workflow. Step 1 (Parse Trino SQL to Calcite SqlNode) was completed by @wenruimeng. 

**Classes created**

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-87a3883f-7fff-826e-80d4-8cffffb1cd0d"><div dir="ltr" style="margin-left:0pt;" align="left">

Class | Purpose
-- | --
`TrinoSqlToRelConverter` | Class to specialize the conversion of `SQLNode` to `RelNode`. This method is used in the implementation of `TrinoToRelConverter.getSqlToRelConverter()`.
`TrinoToRelConverter` | Class that drives the end to end process of Trino Sql to Coral IR conversion and serves as the main entry point for clients to convert. It Implements `ToRelConverter`.
`TrinoViewExpander` | Class that implements `ViewExpander` interface to support expansion of Trino Views to relational algebra.
`Trino2CoralOperatorConverter` | This class extends the `SqlShuttle` class and overrides the “visit(SqlCall call)” method. This class transforms Trino SqlOperator to Coral SqlOperator for a `SqlNode` using the transformer map in `Trino2CoralOperatorTransformerMap`.
`Trino2CoralOperatorTransformerMap` | A mapping of the Trino operator name to the corresponding `OperatorTransformer`. It is used in `Trino2CoralOperatorConverter` to make the transformation happen.
`Trino2CoralOperatorTransformerMapUtils` | Functions that help create mappings in `Trino2CoralOperatorTransformerMap`, also contains a function to fetch by key in the transformer map.
`OperatorTransformer` | A generic class to express operator transformations. This class is a re-implementation of `UDFTransformer` functionality, but at the SQLNode level. The plan is to have this class [replace the UDFTransformer in all use cases](https://github.com/linkedin/coral/issues/237).

</div></b>

Future work includes: handling corner cases and improving the tests to include a conversion from the converted RelNode back to Sql.


ljfgem (commented):
Great work Kevin!
Left some nits and may return to go through the details once I get a chance. According to my previous experiment, the current implementation can handle most of the common Trino SQLs.
@wmoustafa @funcheetah please also take a look if you get a chance.

wmoustafa (commented):
Thanks @KevinGe00! That is a very good start. We probably need to ensure we have not over-simulated HiveToRel. Let us add a basic test case and see what code can be removed so the only remaining code is what is needed to support the test case.

wmoustafa:
> According to my previous experiment, the current implementation can handle most of the common Trino SQLs in Wenrui's test.

We had unresolved questions on the function registry, and whether each input dialect will have its own function registry or all input dialects can reuse the same function registry. Also, we need to make sure that we convert to the same Rel that Hive QL translates to for a semantically equivalent query. For example, a query in Trino using the `strpos` function should be converted to Coral IR with `instr` function.

ljfgem:
@wmoustafa, for this PR, I would prefer to try to support all the supported SQLs in `ParseTreeBuilderTest#getSupportedSql` and clean the unnecessary code. And then, we can add `functionResolver` to function conversion like `strpos -> instr` and other corner cases. Do you think we need to use this one PR to finish all the remaining work?

And for the test, I would prefer to instantiate `TrinoToRelConverter` with `localMetastore` instead of writing a lot of SQLs to create db/table. What's your opinion on that?

wmoustafa:
@ljfgem did you mean to support all the supported SQLs in `ParseTreeBuilderTest#getSupportedSql` except the corner cases (because I expect the corner cases will show up there)? My preference is to start from a function translation that requires a rewrite (e.g., `strpos` to `instr`), because that will shed light on how to go about those translations that preserve semantics, which is actually the most important part of the Rel conversion.

ljfgem:
@wmoustafa in `ParseTreeBuilderTest#getSupportedSql`, looks like there is no such a corner case you mentioned, so I didn't consider it.
But yeah, we can include the function translation that requires a rewrite in this PR for a test SQL like `select strpos('aaa', 'a')`.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/StaticTrinoFunctionRegistry.java
```diff
@@ -0,0 +1,485 @@
...
+ * Static implementation of TrinoFunctionRegistry that has hard-coded list of all
+ * function names. This has a major disadvantage that the user defined functions are
+ * not available to the registry without manually adding the entry here and uploading
+ * a new version of library.
+ * TODO: Provide function registry catalog
+ *
+ * Note that Coral maintains a copy of StaticTrinoFunctionRegistry for read only at run time.
+ * For individual query, we create a copy of function registry in a RelConTextProvider object.
+ */
+public class StaticTrinoFunctionRegistry implements FunctionRegistry {
```
funcheetah, line 67: Are we planning to refactor non trino specific functions out?

funcheetah:
Thanks @KevinGe00 for the great PR! Left a few nits.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/StaticTrinoFunctionRegistry.java
ljfgem: This PR has removed all the `com.linkedin.xxx` functions, and we can remove other non-trino functions and do a further code cleaning (for other classes) in the following PRs.

ljfgem (commented):
Thanks Kevin!
Left some nits, and you may need to run `./gradlew spotlessApply` to solve the building issue, please make sure that `./gradlew clean build` with Java 8 could work.

coral-trino/src/test/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverterTest.java
```diff
@@ -0,0 +1,203 @@
...
+    return ImmutableList.<List<String>> builder()
+        .add(ImmutableList.of("select * from foo",
+            "LogicalProject(show=[$0], a=[$1], b=[$2], x=[$3], y=[$4])\n"
+                + "  LogicalTableScan(table=[[hive, default, foo]])\n"))
+        .add(ImmutableList.of("select * from foo /* end */",
+            "LogicalProject(show=[$0], a=[$1], b=[$2], x=[$3], y=[$4])\n"
+                + "  LogicalTableScan(table=[[hive, default, foo]])\n"))
+        .add(ImmutableList.of("/* start */ select * from foo",
+            "LogicalProject(show=[$0], a=[$1], b=[$2], x=[$3], y=[$4])\n"
+                + "  LogicalTableScan(table=[[hive, default, foo]])\n"))
```
wmoustafa, line 57: We should probably have a SQL to SQL test (e.g., Trino to Hive/Spark along the lines of Hive to Trino). Could you summarize current HiveToRel testing flavors? We may want to follow them.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoOperatorTable.java
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.calcite.util.Util;
+
+import com.linkedin.coral.common.functions.Function;
+
+
+/**
+ * Class to resolve Trino function names in SQL definition based on
+ * the mapping stored in table parameters in the metastore.
+ */
+public class TrinoOperatorTable implements SqlOperatorTable {
```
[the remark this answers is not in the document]
ljfgem: `DaliOperatorTable` is coupled with `HiveFunctionResolver` which we are not planning to create in `trino2rel` at present, so I would prefer to create another operator table for now. We may do further refactoring afterwards, but I don't expect that to be a part of internship project.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/ParseTreeBuilder.java
```diff
@@ -365,17 +362,26 @@ protected SqlNode visitFunctionCall(FunctionCall node, ParserVisitorContext cont
...
-    SqlIdentifier functionName = convertQualifiedName(node.getName(), getPos(node));
+    SqlIdentifier functionIdentifier = convertQualifiedName(node.getName(), getPos(node));
     if (node.getName().toString().equalsIgnoreCase("COUNT") && operands.isEmpty()) {
       operands.add(createStarIdentifier(ZERO));
     }
-    SqlUnresolvedFunction unresolvedFunction =
-        new SqlUnresolvedFunction(functionName, null, null, null, null, SqlFunctionCategory.USER_DEFINED_FUNCTION);
+
+    SqlOperator convertedFunction;
+    Collection<Function> functions = functionRegistry.lookup(functionName);
```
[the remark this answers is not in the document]
ljfgem: I think it can handle recursion, since it's using visitor pattern and I tried SQL like `select foo(foo(1))`, it worked well.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/ParseTreeBuilder.java
```diff
@@ -365,17 +362,26 @@ protected SqlNode visitFunctionCall(FunctionCall node, ParserVisitorContext cont
...
+
+    if (!functions.isEmpty()) {
+      convertedFunction = functions.iterator().next().getSqlOperator();
```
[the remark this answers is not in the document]
ljfgem: No, I think we may need something like `UDFTransformer` for this case.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/ParseTreeBuilder.java
```diff
@@ -852,8 +858,17 @@ protected SqlNode visitInsert(Insert node, ParserVisitorContext context) {
...
+    String callName = node.getName().toString().toLowerCase(Locale.ROOT);
+    Collection<Function> functions = functionRegistry.lookup(callName);
+    SqlCall call;
+
+    if (!functions.isEmpty()) {
+      call = createCall(functions.iterator().next().getSqlOperator(), toListOfSqlNode(node.getArguments(), context));
+
+    } else {
+      call = getUnresolvedFunction(node.getName().toString(), toListOfSqlNode(node.getArguments(), context));
+    }
```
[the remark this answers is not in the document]
ljfgem: At present, given there are only 2 parts using function map, I think it's fine to keep them in `ParseTreeBuilder` as how coral-hive did, we may create a helper method to get the converted function.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/StaticTrinoFunctionRegistry.java
```diff
@@ -0,0 +1,485 @@
...
+    addFunctionEntry("reflect", HiveReflectOperator.REFLECT);
+    addFunctionEntry("java_method", HiveReflectOperator.REFLECT);
+
+    // Context functions
+    addFunctionEntry("current_user", CURRENT_USER);
+
+    // Trino specific functions
+    createAddTrinoFunction("foo", "foo", ReturnTypes.INTEGER, NILADIC); // for testing
+    createAddTrinoFunction("foo", "foo", ReturnTypes.INTEGER, family(SqlTypeFamily.ANY)); // for testing
+    createAddTrinoFunction("strpos", "instr", ReturnTypes.INTEGER, STRING_STRING);
```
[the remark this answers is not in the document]
ljfgem: `createAddUserDefinedFunction` can be removed once we clean the non-trino functions

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/StaticTrinoFunctionRegistry.java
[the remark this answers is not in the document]
funcheetah: +1 on this question. It is more like function map in my understanding.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoCalciteUDFTransformer.java
```diff
@@ -0,0 +1,372 @@
...
+ *
+ * 4. Operator transformers: Check the second parameter (scheme) matches 'utf-8' with any casing using Java Regex.
+ * [ {
+ *    "regex" : "^.*(?i)(utf-8).*$",
+ *    "input" : 2,
+ *    "name" : "from_utf8"
+ *   }
+ * ]
+ */
+class TrinoCalciteUDFTransformer {
```
KevinGe00, line 121: any advice on how to create a common interface that both TrinoCalciteUDFTransformer and UDFTransformer (from rel2trino) can implement? The problem right now is that one has functions that deal with RexNodes and the other with SqlNodes.

coral-trino/src/test/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverterTest.java
```diff
@@ -174,19 +183,8 @@ public void afterClass() throws IOException {
         .add(ImmutableList.of("call foo(3)",
             "LogicalProject(EXPR$0=[foo(3)])\n" + "  LogicalValues(tuples=[[{ 0 }]])\n"))
         .add(ImmutableList.of("call foo()", "LogicalProject(EXPR$0=[foo()])\n" + "  LogicalValues(tuples=[[{ 0 }]])\n"))
-        .add(ImmutableList.of("select count(*) from foo",
```
KevinGe00: deleting any tests with "count" as it's failing an unknown identifier test, will add support for this as part of corner cases

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoCalciteUDFTransformer.java
```diff
@@ -0,0 +1,372 @@
...
+    if (value == null) {
+      throw new IllegalArgumentException("JSON node for transformation should be either op, input, or value");
+    }
+    if (!value.isJsonPrimitive()) {
+      throw new IllegalArgumentException("Value should be of primitive type: " + value);
+    }
+
+    final JsonPrimitive primitive = value.getAsJsonPrimitive();
+    if (primitive.isString()) {
+      return createStringLiteral(primitive.getAsString(), SqlParserPos.ZERO);
```
KevinGe00, line 313: are these primitives equivalent to the RexNode versions seen here: https://github.com/linkedin/coral/blob/6ad24e180af64130bb0b3dede576d59faa24d575/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/UDFTransformer.java#L316

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
```diff
@@ -579,10 +579,17 @@ public boolean isOptional(int i) {
...
    */
-  public ImmutableMultimap<String, Function> getRegistry() {
+  public ImmutableMultimap<String, Function> getRegistryCopy() {
     return ImmutableMultimap.copyOf(FUNCTION_MAP);
   }
 
+  /**
+   * @return copy of internal function registry
+   */
+  public Multimap<String, Function> getRegistry() {
```
KevinGe00, line 589: needed this since TrinoToRelConverter needs the most up to date version of the map, the immutable map doesn't contain the new entries added in TrinoCalciteUDFMap

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/ParseTreeBuilder.java
```diff
@@ -27,16 +54,186 @@
 
 import com.linkedin.coral.common.HiveTypeSystem;
 import com.linkedin.coral.common.calcite.CalciteUtil;
-import com.linkedin.coral.hive.hive2rel.functions.HiveFunctionResolver;
-import com.linkedin.coral.hive.hive2rel.functions.StaticHiveFunctionRegistry;
 
-import coral.shading.io.trino.sql.tree.*;
```
ljfgem, line 33: Why do we remove it and add so many lines of imports?

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoCalciteUDFMapUtils.java
```diff
@@ -0,0 +1,91 @@
...
+   * @param numOperands Number of operands
+   * @param calciteSqlOperator The Calcite Sql Operator that is used as the target operator in the map
+   * @param operandTransformer Operand transformers, null for identity transformation
+   * @param resultTransformer Result transformer, null for identity transformation
+   */
+  static void createUDFMapEntry(Map<String, TrinoCalciteUDFTransformer> transformerMap, SqlOperator trinoOp,
+      int numOperands, SqlOperator calciteSqlOperator, String operandTransformer, String resultTransformer) {
+
+    StaticHiveFunctionRegistry.createAddUserDefinedFunction(calciteSqlOperator.getName(),
+        calciteSqlOperator.getReturnTypeInference(), calciteSqlOperator.getOperandTypeChecker());
```
ljfgem, line 71: Could you add a comment here explaining the reason of adding calcite operator to `StaticHiveFunctionRegistry`?

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoCalciteUDFMapUtils.java
```diff
@@ -0,0 +1,91 @@
...
+
+    transformerMap.put(getKey(trinoOp.getName(), numOperands), TrinoCalciteUDFTransformer.of(trinoOp.getName(),
+        calciteSqlOperator, operandTransformer, resultTransformer, null));
+  }
+
+  public static SqlOperator createUDF(String transformerName, SqlReturnTypeInference typeInference) {
+    return new SqlUserDefinedFunction(new SqlIdentifier(ImmutableList.of(transformerName), SqlParserPos.ZERO),
+        typeInference, null, null, null, null);
+  }
```
ljfgem, line 80: I think we can remove this method, just keep `createUDF(String functionName, SqlReturnTypeInference returnTypeInference, SqlOperandTypeChecker operandTypeChecker)`

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoCalciteUDFTransformer.java
ljfgem: Great question!
@wmoustafa and I have discussed it, we may investigate the approach of transforming `SqlOperator` directly, then refactor both `trino2rel` and `rel2trino` modules with that change accordingly.
The investigation and refactoring wouldn't be included in the internship project though.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoCalciteUDFTransformer.java
ljfgem: I think so.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverter.java
```diff
@@ -0,0 +1,103 @@
...
+
+/*
+ * We provide this class as a public interface by providing a thin wrapper
+ * around TrinoSqlToRelConverter. Directly using TrinoSqlToRelConverter will
+ * expose public methods from SqlToRelConverter. Use of SqlToRelConverter
+ * is likely to change in the future if we want more control over the
+ * conversion process. This class abstracts that out.
+ */
+public class TrinoToRelConverter extends ToRelConverter {
+  private final ParseTreeBuilder parseTreeBuilder = new ParseTreeBuilder();;
```
ljfgem, line 44: Double `;`

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoCalciteUDFTransformer.java
```diff
@@ -0,0 +1,372 @@
...
+
+      if (Pattern.matches(matcher, sourceOperands.get(index).toString())) {
+        return TrinoCalciteUDFMapUtils.createUDF(functionName, operator.getReturnTypeInference());
+      }
+    }
+    return operator;
+  }
+
+  /**
+   * TODO(ralam): Add this as a general utility in coral-calcite or look for other base libraries with this function.
```
ljfgem, line 360: May remove this `TODO`

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoCalciteUDFMap.java
```diff
@@ -0,0 +1,47 @@
...
+import static org.apache.calcite.sql.type.OperandTypes.*;
+
+
+public class TrinoCalciteUDFMap {
+  private TrinoCalciteUDFMap() {
+  }
+
+  private static final Map<String, TrinoCalciteUDFTransformer> TRANSFORMER_MAP = new HashMap<>();
+
+  static {
```
ljfgem, line 25: We may add a `TODO` here to add more functions in the future.

ljfgem (commented):
Thanks @KevinGe00!

ljfgem:
@wmoustafa @funcheetah PTAL if you get a chance, thanks!

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoSqlValidator.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.calcite.sql.validate.SqlValidatorScope;
+
+import com.linkedin.coral.common.functions.FunctionFieldReferenceOperator;
+
+
+/**
+ * Validates the parse tree of a Trino SQL statement,
+ * and provides semantic information about the parse tree.
+ */
+class TrinoSqlValidator extends SqlValidatorImpl {
```
wmoustafa, line 29: If the validator now works on Coral IR, why are we copying the Hive validator?

coral-trino/src/test/resources/hive-from-trino.xml
```diff
@@ -0,0 +1,33 @@
...
+    Licensed under the BSD-2 Clause license.
+    See LICENSE in the project root for license information.
+
+-->
+<configuration>
+  <property>
+    <name>hive.exec.scratchdir</name>
+    <value>${coral.from.trino.test.dir}/hive-scratch-dir</value>
+  </property>
+  <property>
```
wmoustafa, line 13: Do we need to copy this file?

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoCalciteUDFMap.java
```diff
@@ -0,0 +1,47 @@
...
+    return TRANSFORMER_MAP.get(getKey(trinoOpName, numOperands));
+  }
+
+  /**
+   * @return copy of internal transformer map
+   */
+  public static Map<String, TrinoCalciteUDFTransformer> getTransformerMap() {
+    return TRANSFORMER_MAP;
+  }
+
```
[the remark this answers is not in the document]
KevinGe00: It's referenced from a static context in TrinoToRelConverterTest

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoCalciteUDFMap.java
```diff
@@ -0,0 +1,47 @@
...
+import static com.linkedin.coral.trino.trino2rel.TrinoCalciteUDFMapUtils.*;
+import static org.apache.calcite.sql.type.OperandTypes.*;
+
+
+public class TrinoCalciteUDFMap {
+  private TrinoCalciteUDFMap() {
+  }
+
+  private static final Map<String, TrinoCalciteUDFTransformer> TRANSFORMER_MAP = new HashMap<>();
```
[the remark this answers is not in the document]
KevinGe00: I think `TRANSFORMER_MAP` works here since it's a mapping from Trino operator name to a `TrinoCalciteOperatorTransformer` object.

coral-trino/src/test/resources/hive-from-trino.xml
KevinGe00: Not needed, using just `hive.xml `in `ToRelTestUtils` and it's working fine.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoSqlValidator.java
KevinGe00: Using `HiveSqlValidator` in coral-trino, will delete `TrinoSqlValidator` for now since it's the same code.

coral-trino/src/test/resources/hive-from-trino.xml
wmoustafa: So are you planning to delete this file?

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoSqlValidator.java
wmoustafa: Do you mean you will delete it as part of this PR?

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoCalciteUDFMap.java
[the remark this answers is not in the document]
wmoustafa: I was referring to the fact of whether there is a value in wrapping a static variable using a static function, as opposed to exposing the static variable directly.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/ParseTreeBuilder.java
KevinGe00: I think something must have happened when I ran refactoring with the IDE, reverting this back to just `import coral.shading.io.trino.sql.tree.*;`

KevinGe00: Also reverted all other non-necessary imports.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoCalciteUDFTransformer.java
```diff
@@ -0,0 +1,372 @@
...
+ *    "name" : "from_utf8"
+ *   }
+ * ]
+ */
+class TrinoCalciteUDFTransformer {
+  private static final Map<String, SqlOperator> OP_MAP = new HashMap<>();
+
+  // Operators allowed in the transformation
+  static {
+    OP_MAP.put("+", SqlStdOperatorTable.PLUS);
```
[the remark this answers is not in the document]
KevinGe00: `OP_MAP` is used to figure out the corresponding SqlOperator for primitive operators within a json transformer. For example we register `"*"` as `SqlStdOperatorTable.MULTIPLY`, so that if we encounter a json transformer like this it knows what the "op" means when doing the transformation:  
```
 [
   { "op":"*",
     "operands":[
       {"input":1}, // input 0 is reserved for result transformer. source inputs start from 1
       { "op":"^",
         "operands":[
           {"value":10},
           {"input":2}]}]}]
```

In contrast, the `TRANSFORMER_MAP` maps a Trino Sql operator names to the corresponding `TrinoCalciteOperatorTransformer`. This map is manually populated.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoSqlConformance.java
```diff
@@ -0,0 +1,20 @@
...
+import org.apache.calcite.sql.validate.SqlConformanceEnum;
+import org.apache.calcite.sql.validate.SqlDelegatingConformance;
+
+
+public class TrinoSqlConformance extends SqlDelegatingConformance {
+
+  public static final SqlConformance TRINO_SQL = new TrinoSqlConformance();
+
+  private TrinoSqlConformance() {
+    super(SqlConformanceEnum.PRAGMATIC_2003);
```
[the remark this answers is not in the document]
KevinGe00: Left it the same as coral-hive, doesn't seem to cause any trouble right now. There does exist a [Trino specific Conformance Enum ](https://calcite.apache.org/javadocAggregate/org/apache/calcite/sql/validate/SqlConformanceEnum.html#PRESTO) in Calcite but the version of Calcite in Coral doesn't have it.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoSqlToRelConverter.java
```diff
@@ -0,0 +1,114 @@
...
+      case UNNEST:
+        convertUnnestFrom(bb, from);
+        break;
+      default:
+        super.convertFrom(bb, from);
+        break;
+    }
+  }
+
+  //TODO: make convertUnnestFrom Trino specific
```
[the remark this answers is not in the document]
KevinGe00: It seems like most of this code is still needed for coral-trino, the only "Hive specific" part is this:
```
      // In Hive, "LATERAL VIEW EXPLODE(arr) t" is equivalent to "LATERAL VIEW EXPLODE(arr) t AS col".
      // Use the default column name "col" if not specified.
```

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoConvertletTable.java
```diff
@@ -0,0 +1,42 @@
...
+  @SuppressWarnings("unused")
+  public RexNode convertCast(SqlRexContext cx, SqlCastFunction cast, SqlCall call) {
+    final SqlNode left = call.operand(0);
+    RexNode leftRex = cx.convertExpression(left);
+    SqlDataTypeSpec dataType = call.operand(1);
+    RelDataType castType = dataType.deriveType(cx.getValidator(), true);
+    // can not call RexBuilder.makeCast() since that optimizes to remove the cast
+    // we don't want to remove the cast
+    return cx.getRexBuilder().makeAbstractCast(castType, leftRex);
+  }
```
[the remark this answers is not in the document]
KevinGe00: As of now, we still need this. This step happens during the SqlNode to RelNode phase meaning that not all the conversion happens during the SqlNode layer. Maybe this could be something to work on in the future.

coral-trino/src/test/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverterTest.java
```diff
@@ -25,28 +25,34 @@
...
-import static org.testng.Assert.*;
+import static org.testng.AssertJUnit.assertEquals;
 
 
 public class TrinoToRelConverterTest {
   private static HiveConf conf;
 
   @BeforeClass
   public void beforeClass() throws HiveException, IOException, MetaException {
+    // Simulating a Coral environment where "foo" exists
```
KevinGe00, line 41: Make sure we register UDFs before instantiating TrinoToRelConverter so they are included in the registry

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoCalciteTransformerMapUtils.java
```diff
@@ -0,0 +1,86 @@
...
+   * @param trinoOp Trino SQL operator
+   * @param numOperands Number of operands
+   * @param calciteSqlOperator The Calcite Sql Operator that is used as the target operator in the map
+   * @param operandTransformer Operand transformers, null for identity transformation
+   * @param resultTransformer Result transformer, null for identity transformation
+   */
+  static void createTransformerMapEntry(Map<String, TrinoCalciteOperatorTransformer> transformerMap,
+      SqlOperator trinoOp, int numOperands, SqlOperator calciteSqlOperator, String operandTransformer,
+      String resultTransformer) {
+
```
KevinGe00, line 72: No longer registering coral operators here, if there is a Coral function we want to translate Trino to, it should already exist or to be manually added. It doesn't make sense to just "invent" coral operators here.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoConvertletTable.java
```diff
@@ -1,42 +0,0 @@
-/**
```
KevinGe00: Deleting as we only need one convertlet table for all of cora since all SqlNodes that arrive here should already be in Coral IRl, this will be in a future PR.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/DaliOperatorTable.java
```diff
@@ -30,12 +32,17 @@
 public class DaliOperatorTable implements SqlOperatorTable {
   // TODO: support injection framework to inject same function resolver here and ParseTreeBuilder.
   // For now, we create another instance since the function registry is simple.
-  private final HiveFunctionResolver funcResolver;
+  private HiveFunctionResolver funcResolver;
+  private Multimap<String, Function> funcRegistry;
```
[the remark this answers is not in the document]
ljfgem: +1, @wmoustafa you meant an issue on GitHub, or open an internal ticket?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveSqlValidator.java
```diff
@@ -22,7 +22,7 @@
 import com.linkedin.coral.common.functions.FunctionFieldReferenceOperator;
 
 
-class HiveSqlValidator extends SqlValidatorImpl {
+public class HiveSqlValidator extends SqlValidatorImpl {
```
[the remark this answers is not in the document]
ljfgem: +1

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoCalciteOperatorTransformer.java
```diff
@@ -0,0 +1,350 @@
...
+ *
+ * 4. Operator transformers: Check the second parameter (scheme) matches 'utf-8' with any casing using Java Regex.
+ * [ {
+ *    "regex" : "^.*(?i)(utf-8).*$",
+ *    "input" : 2,
+ *    "name" : "from_utf8"
+ *   }
+ * ]
+ */
+class TrinoCalciteOperatorTransformer {
```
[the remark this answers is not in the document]
ljfgem: +1, we may also need to change all the related parameter names.
i.e. `trinoOperatorName` -> `fromOperatorName`

ljfgem (commented):
Left some nits, thanks Kevin!

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/DaliOperatorTable.java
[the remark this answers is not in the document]
KevinGe00: ^ just tested this and it works.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/ParseTreeBuilder.java
```diff
@@ -27,16 +26,14 @@
...
 import com.linkedin.coral.common.calcite.CalciteUtil;
-import com.linkedin.coral.hive.hive2rel.functions.HiveFunctionResolver;
-import com.linkedin.coral.hive.hive2rel.functions.StaticHiveFunctionRegistry;
 
 import coral.shading.io.trino.sql.tree.*;
 
 import static com.linkedin.coral.common.calcite.CalciteUtil.*;
-import static java.lang.String.format;
-import static org.apache.calcite.sql.SqlFunctionCategory.USER_DEFINED_FUNCTION;
+import static java.lang.String.*;
```
funcheetah, line 33: nit: should we avoid wildcard imports and import classes needed explicitly?

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/Trino2CalciteOperatorConverter.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.Locale;
+
+import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.util.SqlShuttle;
+
+
+/**
+ * Rewrites the SqlNode tree to replace Trino SQL operators with Calcite Operators to obtain a Calcite-compatible plan.
```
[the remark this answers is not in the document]
funcheetah: +1. It is good to add a conversion example.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/DaliOperatorTable.java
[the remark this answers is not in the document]
ljfgem: @wmoustafa could you share your thoughts here? This is the only one that is not clear to @KevinGe00

funcheetah (commented):
Thanks @KevinGe00 . Left a few nits. In addition, do the PR title and description still holds?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/DaliOperatorTable.java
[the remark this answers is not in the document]
wmoustafa: Ad, yes, function resolver sounds more elegant.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverter.java
```diff
@@ -44,8 +44,8 @@
 public class TrinoToRelConverter extends ToRelConverter {
   private final ParseTreeBuilder parseTreeBuilder = new ParseTreeBuilder();
   private final ParserVisitorContext parserVisitorContext = new ParserVisitorContext();
-  private final Multimap<String, Function> functionRegistry = new StaticHiveFunctionRegistry().getRegistry();
-
+  private final HiveFunctionResolver functionResolver =
```
KevinGe00, line 47: This is will become just `FunctionResolver` in https://github.com/linkedin/coral/issues/235

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/OperatorTransformer.java
```diff
@@ -0,0 +1,350 @@
...
+  public static final String REGEX = "regex";
+  public static final String NAME = "name";
+
+  public final String trinoOperatorName;
+  public final SqlOperator targetOperator;
+  public final List<JsonObject> operandTransformers;
+  public final JsonObject resultTransformer;
+  public final List<JsonObject> operatorTransformers;
+
+  private OperatorTransformer(String trinoOperatorName, SqlOperator targetOperator,
```
wmoustafa, line 145: Still references to `trino`.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/OperatorTransformer.java
```diff
@@ -0,0 +1,350 @@
...
+ *          {"value":10},
+ *          {"input":2}]}]}]
+ *
+ *
+ * 4. Operator transformers:
+ * none
+ *
+ * Example 2:
+ * In Calcite, there exists a hive-derived function to decode binary data given a format, DECODE(binary, scheme).
+ * In Trino, there is no generic decoding function that takes a decoding-scheme.
```
wmoustafa, line 92: Still references to `Calcite` and `Trino`.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/Trino2CoralOperatorTransformerMapUtils.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.sql.type.SqlOperandTypeChecker;
+import org.apache.calcite.sql.type.SqlReturnTypeInference;
+import org.apache.calcite.sql.validate.SqlUserDefinedFunction;
+
+import com.linkedin.coral.hive.hive2rel.functions.StaticHiveFunctionRegistry;
+
+
+public class Trino2CoralOperatorTransformerMapUtils {
+  private static final StaticHiveFunctionRegistry HIVE_REGISTRY = new StaticHiveFunctionRegistry();
```
wmoustafa, line 21: `HIVE_REGISTRY` is this needed?

coral-trino/src/test/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverterTest.java
```diff
@@ -0,0 +1,211 @@
...
+        null);
+  }
+
+  @AfterTest
+  public void afterClass() throws IOException {
+    FileUtils.deleteDirectory(new File(conf.get(CORAL_FROM_TRINO_TEST_DIR)));
+  }
+
+  @DataProvider(name = "support")
+  public Iterator<Object[]> getSupportedSql() {
```
wmoustafa, line 66: I have not checked the correctness of the output (expected result). Have you manually inspected that it looks good?

wmoustafa (commented):
Could you please elaborate on the PR description and list all the new classes introduced and their description?

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/OperatorTransformer.java
KevinGe00: Is this a problem in the examples? Although the it's a generic class to express operator transformers, the examples should specify the language. In this case, the transformer just happens to be from Trino to Calcite.

coral-trino/src/test/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverterTest.java
KevinGe00: I have checked for some correctness (i.e are all the columns we expect to be there, there?) but it definitely hasn't been rigorous. The problem is I can't find good Calcite documentation/examples so I can check for correctness rigorously. How was this done in HiveToRel?

wmoustafa: One idea is to convert back to SQL and read the query if it looks correct. The flip side of the transformation (CoralIR to Trino SQL) should have a transformation from "TrinoRel" (not "CoralRel") to Trino SQL ("TrinoRel" is the transformed version of "CoralRel" after applying the (coral-to-trino) transformation map). You could provide your Rel to that part and see how the SQL looks like.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/OperatorTransformer.java
wmoustafa: I am referring to the words of `Calcite` and `Trino` which I believe should not occur in this class.

wmoustafa: In this comment they appeared in the comments (JavaDoc).

KevinGe00: If I took out `Calcite` and `Trino` from the comments the examples wouldn't make sense. For example, this comment from the examples: 
`The transformation from Calcite TRUNCATE to Trino TRUNCATE is represented as follows`

Do you mean instead I should just generalize it and say something along the lines of `The transformation of TRUNCATE from one SQL language to another is represented as follows`?

ljfgem (commented):
Added 2 more nits

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/ParseTreeBuilder.java
```diff
@@ -27,16 +26,14 @@
...
 
 import coral.shading.io.trino.sql.tree.*;
 
 import static com.linkedin.coral.common.calcite.CalciteUtil.*;
 import static java.lang.String.format;
-import static org.apache.calcite.sql.SqlFunctionCategory.USER_DEFINED_FUNCTION;
+import static org.apache.calcite.sql.SqlFunctionCategory.*;
 import static org.apache.calcite.sql.fun.SqlStdOperatorTable.*;
-import static org.apache.calcite.sql.parser.SqlParserPos.ZERO;
+import static org.apache.calcite.sql.parser.SqlParserPos.*;
```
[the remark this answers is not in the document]
KevinGe00: I'm not entirely sure when these are getting changed, I think either spotlessApply or when I refactor the code using some IDE tools. Either way, I'll be more careful with these in the future.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/OperatorTransformer.java
wmoustafa: Yes, that is one step towards where I think this should go. We can improve `SQL language` to `SQL operator in one IR to SQL operators in another IR`. I expect Coral IR will be used both on the right and left hand side of this map (once in language to Coral and once in Coral to language). If so, we cannot even attribute one of the two sides (right or left) to Coral. Do you agree @ljfgem?

ljfgem: Agreed.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoSqlToRelConverter.java
```diff
@@ -0,0 +1,113 @@
...
+  TrinoSqlToRelConverter(RelOptTable.ViewExpander viewExpander, SqlValidator validator,
+      Prepare.CatalogReader catalogReader, RelOptCluster cluster, SqlRexConvertletTable convertletTable,
+      Config config) {
+    super(viewExpander, validator, catalogReader, cluster, convertletTable, config);
+  }
+
+  // This differs from base class in two ways:
+  // 1. This does not validate the type of converted rel rowType with that of validated node. This is because
+  //    hive is lax in enforcing view schemas.
+  // 2. This skips calling some methods because (1) those are private, and (2) not required for our usecase
```
wmoustafa, line 56: It is not clear what that means.

wmoustafa: Is this copied from `HiveSqlToRelConverter`?

KevinGe00: Yes it's copied from HiveSqlToRelConverter.

ljfgem (approved):
LGTM, thanks @KevinGe00 !

wmoustafa:
Thanks @KevinGe00! Great work.

PR #226, opened 2022-01-25
Coral-Schema: Correct nullability determination in coral-schema

rzhang10:
During investigation of a extract_union view coral-schema schema conversion issue, I found a bigger problem in the nullability determination logic in the existing coral-schema module codebase.

The correct determination of the nullability should be using the `RelDataType`'s builtin `isNullable` method, this means, for example, to determine the nullability of a UDF's return type, we should use the UDF's `SqlReturnTypeInference` which when called on the given operands will return the inferred return `RelDataType` which contains the nullability information.

Coral-schema needs to return the correct nullability information to work with engines that have strict nullability checks like spark3.

The unit test failures are expected, because the code change covers the logic change in the main code but the framework and unit tests setup needs more changes to pass all tests.

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
@@ -217,32 +217,6 @@ class SchemaUtilities {
     }
   }
 
-  static boolean isFieldNullable(@Nonnull RexCall rexCall, @Nonnull Schema inputSchema) {
-    Preconditions.checkNotNull(rexCall);
-    Preconditions.checkNotNull(inputSchema);
-
-    // the field is non-nullable only if all operands are RexInputRef
-    // and corresponding field schema type of RexInputRef index is not UNION
-    List<RexNode> operands = rexCall.getOperands();
-    for (RexNode operand : operands) {
-      if (operand instanceof RexInputRef) {
-        Schema.Field field = inputSchema.getFields().get(((RexInputRef) operand).getIndex());
-        if (Schema.Type.UNION.equals(field.schema().getType())) {
-          return true;
-        }
-      } else if (operand instanceof RexCall) {
-        boolean isNullable = isFieldNullable((RexCall) operand, inputSchema);
-        if (isNullable) {
-          return true;
-        }
-      } else {
-        return true;
-      }
-    }
-
-    return false;
-  }
-
   static void appendField(@Nonnull String fieldName, @Nonnull Schema.Field field,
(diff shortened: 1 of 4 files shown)


autumnust:
"The unit test failures are expected," 
Does it mean we will have a separate PR to fix unit test?

wmoustafa:
> During investigation of a extract_union view coral-schema schema conversion issue, I found a bigger problem in the nullability determination logic in the existing coral-schema module codebase.

Could you elaborate on whether this PR address the "bigger problem"? It is not clear which what are the two problems and which is the bigger one.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -217,32 +217,6 @@ static void appendField(@Nonnull String fieldName, @Nonnull RelDataType fieldRel
     }
   }
 
-  static boolean isFieldNullable(@Nonnull RexCall rexCall, @Nonnull Schema inputSchema) {
-    Preconditions.checkNotNull(rexCall);
-    Preconditions.checkNotNull(inputSchema);
-
-    // the field is non-nullable only if all operands are RexInputRef
```
autumnust, line 224: do we know why we were adding this ? 

It seems to me that, this PR invalidate this nullable-check method and forcing the nullability to come from `RelDataType's#isNullable` method. So if there are some scenario previously relying on this method to return `nullable` now could return different result ? Just thinking out loud, let me know if I miss something here

autumnust (changes requested):

rzhang10:
> > During investigation of a extract_union view coral-schema schema conversion issue, I found a bigger problem in the nullability determination logic in the existing coral-schema module codebase.
> 
> Could you elaborate on whether this PR addresses the "bigger problem"? It is not clear what are the two problems and which is the bigger one.

The bigger problem is the existing nullability determination logic (and functions registrations without the correct nullability information in `SqlReturnTypeInference`).

The change of `coalesce_struct` is just one example of such changes of `SqlReturnTypeInference` (the smaller problem, which is extract_union view not working in spark3)

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
funcheetah: This method is by design to infer nullability of UDF return types. The reason not using RelDataType's#isNullable is that it is not initialized with proper nullability information at the beginning due to various limitations.

ljfgem:
Closing it given #229 has been merged.

PR #229, opened 2022-02-01
Force interring extract_union nullablity using Calcite

rzhang10:
This pr is an alternative solution to #226 , which has minimum impact on affecting existing function's nullability but force the nullablilty of `extract_union` to be inferred by calcite.

This will make the extract_union view's Coral inferred schema align with spark's analysis, thus making spark able to read such views instead of throwing exception.

I manually tested this by translating the failed extract_union view using this patch, and the nullability in the schema is correct. It's a bit hard to write a unit test on this one because the Hive environment (1.2.2) that Coral depends on doesn't contain the extract_union UDF in its codebase yet (it's a feature from open source hive 2+), so we wouldn't be able to create a extract_union view in the test Hive environment to begin with.

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
@@ -51,6 +51,8 @@ class SchemaUtilities {
   private static final Logger LOG = LoggerFactory.getLogger(SchemaUtilities.class);
   private static final String DALI_ROW_SCHEMA = "dali.row.schema";
 
+  // TODO: 2/2/22 Needs to refactor this into a separate registry class
+  // if the num of functions in this set get bigger
   private static final Set<String> USE_CALCITE_NULLABILITY_FUNCS =
       Collections.unmodifiableSet(new HashSet<>(Arrays.asList("extract_union")));
 
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
@@ -7,12 +7,7 @@ package com.linkedin.coral.schema.avro;
 
 import java.io.PrintWriter;
 import java.io.StringWriter;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.StringJoiner;
+import java.util.*;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -55,6 +50,9 @@ import static org.apache.avro.Schema.Type.NULL;
 class SchemaUtilities {
   private static final Logger LOG = LoggerFactory.getLogger(SchemaUtilities.class);
   private static final String DALI_ROW_SCHEMA = "dali.row.schema";
+
+  private static final Set<String> USE_CALCITE_NULLABILITY_FUNCS =
+      Collections.unmodifiableSet(new HashSet<>(Arrays.asList("extract_union")));
 
   // private constructor for utility class
   private SchemaUtilities() {
@@ -220,6 +218,12 @@ class SchemaUtilities {
   static boolean isFieldNullable(@Nonnull RexCall rexCall, @Nonnull Schema inputSchema) {
     Preconditions.checkNotNull(rexCall);
     Preconditions.checkNotNull(inputSchema);
+
+    // we first filter against these static list of functions, whose nullability should be
+    // determined by calcite rather than avro.schema.literal
+    if (USE_CALCITE_NULLABILITY_FUNCS.contains(rexCall.getOperator().getName().toLowerCase())) {
+      return rexCall.getType().isNullable();
+    }
 
     // the field is non-nullable only if all operands are RexInputRef
     // and corresponding field schema type of RexInputRef index is not UNION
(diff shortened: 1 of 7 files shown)


coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -56,6 +51,9 @@
   private static final Logger LOG = LoggerFactory.getLogger(SchemaUtilities.class);
   private static final String DALI_ROW_SCHEMA = "dali.row.schema";
 
+  private static final Set<String> USE_CALCITE_NULLABILITY_FUNCS =
+      Collections.unmodifiableSet(new HashSet<>(Arrays.asList("extract_union")));
+
```
ljfgem, line 56: Is it better to create a registry class in Coral-Schema containing all the functions that need special handling and add detailed comments there?

ljfgem (commented):
Thanks @rzhang10 for this PR!
Could you also add a corresponding unit test?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
funcheetah: Here is not the best place to put the function list. I'm fine to put here for now if we only have this single function. However, please add comment what the list is and a TODO item to move it to a class if there are more those functions.

funcheetah:
Thanks @rzhang10 ! As we discussed, I'm okay with this approach. Left some nits.

rzhang10:
> Thanks @rzhang10 for this PR! Could you also add a corresponding unit test?

The reason I didn't add unit test is explained in the limitation in the description, do you have other methods to circumvent that?

ljfgem:
> > Thanks @rzhang10 for this PR! Could you also add a corresponding unit test?
> 
> The reason I didn't add unit test is explained in the limitation in the description, do you have other methods to circumvent that?

@rzhang10 you may derive schema from SQL directly, like:
```
new RelToAvroSchemaConverter(hiveMetastoreClient).convert(new HiveToRelConverter(hiveMetastoreClient).convertSql(sql), false);
```

rzhang10:
> > > Thanks @rzhang10 for this PR! Could you also add a corresponding unit test?
> > 
> > 
> > The reason I didn't add unit test is explained in the limitation in the description, do you have other methods to circumvent that?
> 
> @rzhang10 you may derive schema from SQL directly, like:
> 
> ```
> new RelToAvroSchemaConverter(hiveMetastoreClient).convert(new HiveToRelConverter(hiveMetastoreClient).convertSql(sql), false);
> ```

Yes I noticed such API exists too, thanks for the pointer.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
rzhang10: Added a todo item.

ljfgem:
Thanks @rzhang10! I just started the i-test, have you tested this PR on Spark shell for the affected views using `extract_union` ?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
```diff
@@ -92,6 +94,19 @@ public Schema toAvroSchema(String dbName, String tableOrViewName) {
...
+   *
+   * @param sql SQL string literal
+   * @return avro schema for a given sql query
+   */
+  public Schema toAvroSchema(String sql) {
+    Preconditions.checkNotNull(sql);
+
+    RelNode relNode = hiveToRelConverter.convertSql(sql);
+    return relToAvroSchemaConverter.convert(relNode, false);
+  }
```
funcheetah, line 108: I'm not sure about the idea to expose a public method just for testing. Could you make it only visible to test module?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
```diff
@@ -39,6 +40,7 @@
    */
   private ViewToAvroSchemaConverter(HiveMetastoreClient hiveMetastoreClient) {
     this.hiveToRelConverter = new HiveToRelConverter(hiveMetastoreClient);
+    this.relToAvroSchemaConverter = new RelToAvroSchemaConverter(hiveMetastoreClient);
```
funcheetah, line 43: Could you verify if RelToAvroSchemaConverter is stateless and there is no side effect if creation is moved to constructor?

rzhang10: It is, it only has an instance field which is a `HiveMetastoreClient`.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
rzhang10: I actually think this method is a good public API to be exposed, as does the `convertSql` is public in `HiveToRelConverter`.

rzhang10:
> Thanks @rzhang10! I just started the i-test, have you tested this PR on Spark shell for the affected views using `extract_union` ?

I did a test on gateway and it worked.

ljfgem (approved):
LGTM, i-test passed and @rzhang10 verified that the affected views could be queried on Spark shell with this PR.
@funcheetah do you want to take a look? Raymond has responded to your review comments.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
funcheetah: Thanks @rzhang10 for verifying this.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
funcheetah: Discussed offline. We need to think more about the API design and do not release it as public API for now.

funcheetah (approved):
Thanks @rzhang10 for this PR!

PR #232, opened 2022-02-15
Make nested type returned by coalesce UDF all nullable

rzhang10:
We should return all nested type as nullable since Hive types are all nullable by default.

We should also enhance the unit tests to do equality check using `getFullTypeString`, which includes the information of nullability of the type.


ljfgem:
Thanks @rzhang10!
I am wondering why we don't need to modify some existing tests in Coral-Schema. If there is no corresponding test in Coral-Schema, could you please add one?

rzhang10:
> Thanks @rzhang10! Kicked off the i-tests. I am wondering why we don't need to modify some existing tests in Coral-Schema. If there is no corresponding test in Coral-Schema, could you please add one?

I agree that in coral-schema we haven't got such a test case, but I think the tests in `CoalesceStructUtilityTest` can already suffice to ensure we get a correct IR, thus making coral-schema also correct. LMKWYT?

ljfgem (approved):
i-test and test on Spark shell passed
Thanks @rzhang10 !

PR #238, opened 2022-03-07
Improve trino2rel tests

KevinGe00:
Currently, trino2rel only tests the conversion of Trino Sql to RelNodes using `TrinoToRelConverter`, however the problem with doing this alone is that it's hard to manually verify the converted RelNode is correct. 

The solution is to convert the RelNode produced by `TrinoToRelConverter` back to Sql using `RelToTrinoConverter` like in https://github.com/linkedin/coral/blob/13a8c401e8c3d240c8fd9d99a0366031911848af/coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java#L285. 

1. This solutions adds an extra layer of verification that `TrinoToRelConverter` is working as expected.
2. It allows us to more easily verify the converted `RelNode` and Sql is equivalent to the original query since we can use converted Sql (easier to understand than RelNode) to compare 

All expected Sql strings added in this PR have been manually verified to be equivalent to the original query.

### DIFF
--- a/coral-trino/src/test/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverterTest.java
+++ b/coral-trino/src/test/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverterTest.java
@@ -26,6 +26,7 @@ import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
 import com.linkedin.coral.hive.hive2rel.functions.StaticHiveFunctionRegistry;
+import com.linkedin.coral.trino.rel2trino.RelToTrinoConverter;
 
 import static com.linkedin.coral.trino.trino2rel.ToRelTestUtils.*;
 import static com.linkedin.coral.trino.trino2rel.Trino2CoralOperatorTransformerMapUtils.*;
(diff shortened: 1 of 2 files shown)


coral-trino/src/test/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverterTest.java
```diff
@@ -67,131 +68,157 @@ public void afterClass() throws IOException {
...
+                + "    LogicalProject(EXPR$0=[999])\n" + "      LogicalTableScan(table=[[hive, default, foo]])\n",
+            "SELECT COALESCE(999, 999) AS \"ID\"\n" + "FROM (SELECT 123\n" + "FROM \"default\".\"foo\") AS \"t\"\n"
+                + "INNER JOIN (SELECT 999\n" + "FROM \"default\".\"foo\") AS \"t0\" ON 999 = 999"))
         .add(ImmutableList.of("select cast('123' as bigint)",
-            "LogicalProject(EXPR$0=[CAST('123'):BIGINT])\n" + "  LogicalValues(tuples=[[{ 0 }]])\n"))
+            "LogicalProject(EXPR$0=[CAST('123'):BIGINT])\n" + "  LogicalValues(tuples=[[{ 0 }]])\n",
+            "SELECT CAST('123' AS BIGINT)\n" + "FROM (VALUES  (0)) AS \"t\" (\"ZERO\")"))
         .add(ImmutableList.of("select a \"my price\" from \"foo\" \"ORDERS\"",
-            "LogicalProject(MY PRICE=[$1])\n" + "  LogicalTableScan(table=[[hive, default, foo]])\n"))
-        .add(ImmutableList.of(
```
KevinGe00, line 151: The following 2 tests were removed as `RelToTrinoConverter` do not have the capability to carry the conversion of this query.

wmoustafa (approved):
LGTM.

PR #239, opened 2022-03-09
Add a helper to return array of nullable type and change return signature of a UDF

maluchari:
Included a helper method to return nullable array type. Updated return type of jemslookupudf.

### DIFF
--- a/coral-common/src/main/java/com/linkedin/coral/common/functions/FunctionReturnTypes.java
+++ b/coral-common/src/main/java/com/linkedin/coral/common/functions/FunctionReturnTypes.java
@@ -15,6 +15,7 @@ import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.sql.type.ReturnTypes;
 import org.apache.calcite.sql.type.SqlReturnTypeInference;
 import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.sql.type.SqlTypeTransforms;
 
 import com.linkedin.coral.com.google.common.collect.ImmutableList;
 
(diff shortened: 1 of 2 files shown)


xkrogen:
Would it not make more sense to update `arrayOfType` to take a `boolean elementsNullable` parameter, rather than having an entirely separate method for this?

maluchari:
> Would it not make more sense to update `arrayOfType` to take a `boolean elementsNullable` parameter, rather than having an entirely separate method for this?

Sounds good to me. Updated the method.

coral-common/src/main/java/com/linkedin/coral/common/functions/FunctionReturnTypes.java
```diff
@@ -56,9 +56,15 @@ private FunctionReturnTypes() {
...
-        -1);
+  public static SqlReturnTypeInference arrayOfType(final SqlTypeName typeName, boolean elementsNullable) {
+    return opBinding -> {
+      RelDataTypeFactory typeFactory = opBinding.getTypeFactory();
+      RelDataType relType = typeFactory.createSqlType(typeName);
+      RelDataType arrayType = typeFactory.createArrayType(relType, -1);
+      if (!elementsNullable)
+        return arrayType;
+      return typeFactory.createTypeWithNullability(arrayType, true);
+    };
```
ljfgem, line 67: If the original `jobentitlementids` is like `array<bigint>`, I think this logic will make it to be `[array<bigint>, null]` rather than `array<[bigint, null]>` (`[x, null]` means `x` is nullable)
I am not sure if that is what we want, if it is, I am fine with this change.

maluchari: Thanks for pointing out. Modified the method and validated that read from affected view goes through.

coral-common/src/main/java/com/linkedin/coral/common/functions/FunctionReturnTypes.java
```diff
@@ -56,9 +57,18 @@ private FunctionReturnTypes() {
   public static final SqlReturnTypeInference ARRAY_OF_ARG0_TYPE =
       opBinding -> opBinding.getTypeFactory().createArrayType(opBinding.getOperandType(0), -1);
 
-  public static SqlReturnTypeInference arrayOfType(final SqlTypeName typeName) {
-    return opBinding -> opBinding.getTypeFactory().createArrayType(opBinding.getTypeFactory().createSqlType(typeName),
-        -1);
+  public static SqlReturnTypeInference arrayOfType(final SqlTypeName typeName, boolean elementsNullable) {
```
wmoustafa, line 60: Let us keep the old method with the old behavior for backward compatibility. We can have 2 signatures with one calling the other.

maluchari: Done. Added back the old API.

funcheetah (approved):
Thanks @maluchari . LGTM.

ljfgem (approved):
LGTM, thanks @maluchari !

Issue #242, opened 2022-03-10
Trino SqlParser doesn't accept backquotes "`" from trino2rel expanded views


KevinGe00:
A repeated error came up during testing of TrinoToRelConverter,` coral.shading.io.trino.sql.parser.ParsingException: line 1:8: backquoted identifiers are not supported; use double quotes to quote identifiers`. The issue is that during the Trino to Rel conversion, if we [expand the view](https://github.com/linkedin/coral/blob/0ee75f1510d773ddfb53733ddaf5e709a49c6c70/coral-common/src/main/java/com/linkedin/coral/common/ToRelConverter.java#L161), backquotes are added and the query is then rejected by Trino SqlParser.

PR #243, opened 2022-03-10
Replacing backquotes with equivalent double quotes

KevinGe00:
Fixes issue #242

### DIFF
--- a/coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverter.java
+++ b/coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverter.java
@@ -83,7 +83,7 @@ public class TrinoToRelConverter extends ToRelConverter {
 
   @Override
   protected SqlNode toSqlNode(String sql, Table trinoView) {
-    String trimmedSql = trimParenthesis(sql.toUpperCase());
+    String trimmedSql = trimParenthesis(standardizeSql(sql));
     SqlNode parsedSqlNode = PrestoParserDriver.parse(trimmedSql).accept(parseTreeBuilder, parserVisitorContext);
     SqlNode convertedSqlNode = parsedSqlNode.accept(new Trino2CoralOperatorConverter());
     return convertedSqlNode;
@@ -92,6 +92,10 @@ public class TrinoToRelConverter extends ToRelConverter {
   @Override
   protected RelNode standardizeRel(RelNode relNode) {
     return relNode;
+  }
+
+  private static String standardizeSql(String sql) {
+    return sql.toUpperCase().replaceAll("`", "\"");
   }
 
   private static String trimParenthesis(String value) {


coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverter.java
```diff
@@ -94,6 +94,10 @@ protected RelNode standardizeRel(RelNode relNode) {
     return relNode;
   }
 
+  private static String standardizeSql(String sql) {
+    return sql.toUpperCase().replaceAll("`", "\"");
+  }
+
   private static String trimParenthesis(String value) {
     String str = value.trim();
     if (str.startsWith("(") && str.endsWith(")")) {
```
ljfgem, line 103: I think it's better to combine these 2 methods together in one method `standardizeTrinoSql` and add comment of the reason why we need to standardize it.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverter.java
```diff
@@ -94,12 +94,15 @@ protected RelNode standardizeRel(RelNode relNode) {
...
+   * Standardizing Sql to ensure compatability with Sql Parser.
+   */
+  private static String standardizeSql(String value) {
     String str = value.trim();
     if (str.startsWith("(") && str.endsWith(")")) {
-      return trimParenthesis(str.substring(1, str.length() - 1));
+      return standardizeSql(str.substring(1, str.length() - 1));
     }
-    return str;
+    return str.toUpperCase().replaceAll("`", "\"");
```
KevinGe00, line 105: I believe putting replaceAll here should avoid unnecessary calls if we need to recurse, would appreciate confirmation on that.

wmoustafa: Why did we add `toUppercase()`. Also back quotes can be part of the literals so might accidentally change a backquote that was not supposed to be changed.

wmoustafa:
I am not clear on the motivation. So obviously, current test cases do not have this issue. Did you run to them in production views (that they use backquotes)? If yes, how did the production views get created in the first place. You might want to add more details to the description to make the workflow of the issue clear.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverter.java
KevinGe00: The `toUpperCase()` is brought over from [this line](https://github.com/linkedin/coral/blob/346055432259d79180dae037f110a0c08c0e8464/coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverter.java#L86).

wmoustafa:
By any chance are we using the wrong dialect to generate this SQL? Dialects allow customizing the quotes if that was the issue. See https://github.com/linkedin/coral/blob/master/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/TrinoSqlDialect.java#L18.

PR #236, opened 2022-02-28
Rename PrestoParserDriver to TrinoParserDriver

findinpath:
This PR renames `PrestoParserDriver` to `TrinoParserDriver`.

PR #244, opened 2022-03-10
Coral-Trino: Fix incompatible issue after renaming PrestoParserDriver to TrinoParserDriver

ljfgem:
#224 and #236 are not compatible, this PR is to fix it.
It's better to rerun the CI build test before merging any PR in the future, given the test wouldn't run again by itself once another PR is merged.


funcheetah (approved):
Thanks @ljfgem !

PR #246, opened 2022-03-23
Coral-Spark: Maintain alias even if alias is the same as field name

rzhang10:
We noticed that for view like:
  `select foo.bar as bar from t as v`
the previous coral-spark translated spark sql statement will drop the alias because the last part of the field name and the alias is the same, it will generate:
  `select foo.bar from t` for `v`'s definition.

However, when then if user try to in spark3 query the `v` with:
  `select bar from v`
spark3's sql analyzer will fail because it doesn't understand `bar` from the view, since it only understands the full name `foo.bar`, because the alias is dropped.

This change makes the `RelToSqlConverter` maintain the alias all the time. I also chose to override the super method instead of directly changing the method in place inside calcite, to minimize the effect only to the Rel -> sparkSQL part.

The test failures are expected, as I actually want to use the test case failures to showcase the impact on translating other views, since this change will add all column alias even the view definition doesn't specify `as alias` for a certain projection. (this is by the limitation that Calcite doesn't really remember if there's an alias in its `RelNode` IR representation). But overall, I don't think this effect will cause any spark2/3 regressions but I'm open for discussion.

### DIFF
--- a/coral-spark/src/main/java/com/linkedin/coral/spark/SparkRelToSparkSqlConverter.java
+++ b/coral-spark/src/main/java/com/linkedin/coral/spark/SparkRelToSparkSqlConverter.java
@@ -36,7 +36,6 @@ import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.sql.SqlOperator;
 import org.apache.calcite.sql.fun.SqlMultisetValueConstructor;
 import org.apache.calcite.sql.parser.SqlParserPos;
-import org.apache.calcite.sql.validate.SqlValidatorUtil;
 
 import com.linkedin.coral.com.google.common.collect.ImmutableList;
 import com.linkedin.coral.hive.hive2rel.functions.HiveExplodeOperator;
@@ -44,6 +43,7 @@ import com.linkedin.coral.hive.hive2rel.functions.HivePosExplodeOperator;
 import com.linkedin.coral.spark.dialect.SparkSqlDialect;
 import com.linkedin.coral.spark.functions.SqlLateralJoin;
 import com.linkedin.coral.spark.functions.SqlLateralViewAsOperator;
+import com.linkedin.coral.spark.utils.SparkSqlValidatorUtil;
 
 
 /**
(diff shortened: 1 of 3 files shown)


coral-spark/src/main/java/com/linkedin/coral/spark/SparkRelToSparkSqlConverter.java
```diff
@@ -203,6 +206,28 @@ public Result visit(Uncollect e) {
...
+   * equality check, where this method still generates the alias even if
+   * the alias is the same as the simple last part of the column name,
+   * this is to ensure the translated sparkSQL for view like the following
+   *   select foo.bar as bar from t
+   * will be correctly analyzed by spark3 sql analyzer.
+   */
+  @Override
+  public void addSelect(List<SqlNode> selectList, SqlNode node, RelDataType rowType) {
+    String name = rowType.getFieldNames().get(selectList.size());
+    String alias = SqlValidatorUtil.getAlias(node, -1);
```
ljfgem, line 220: `alias` is not used.

ljfgem (commented):
Thanks @rzhang10 !
Could we do the following things?
1. Update the unit tests to make sure it can be built successfully
2. Run i-test for Coral-Spark
3. Test this PR on gateway to make sure the affected views can be queried

rzhang10:
> I am okay with the overall approach of reconstructing the aliases, but I believe we should get them from Coral-Schema. Also, do we know if aliases get dropped if the only difference is casing?

By `get them from Coral-schema` I think you mean aligning the column names output by coral-schema with the alias output of `coral-spark`'s sql ?

If that's the case, I think my current approach already aligns the case of unspecified/auto-generated name like `EXPR_XX` to coral-schema, do you think there are other cases they are not aligned?  @funcheetah could you comment on this?
The current approach will not touch the existing logic/architecture of the translation, by which I mean the dependency structure is:

RelNode IR -> Coral-Schema
                  | -> Coral-Spark-SQL

If on the other hand, your `get them from Coral-schema` actually means to change the dependency to: 
RelNode IR -> Coral-Schema -> Coral-Spark-SQL
Then I feel that will unnecessarily make the architecture unclean and not friendly for open-source.

wmoustafa:
@rzhang10 The root problem here is the disagreement between the Coral-Spark SQL text and Coral-Schema. Aligning them correctly is the clean solution I would say. Keeping two copies of the logic for `EXPR` is an example where the code can break if either of them is changed. Different casing of the aliases is another issue, which leads to this next question: Let us say we produce some casing that is incompatible with Coral-Schema (which also feeds into Spark), would Spark complain? What do you think @shardulm94?

shardulm94:
> Let us say we produce some casing that is incompatible with Coral-Schema (which also feeds into Spark), would Spark complain?

My understanding is that Spark would complain iff `spark.sql.caseSensitive=true`. Default is `spark.sql.caseSensitive=false`.

wmoustafa:
Thanks! So this sounds to be another reason why we should stick to Coral-Schema. The challenge will be passing all the parameters currently passed to Coral-Schema to Coral-Spark, so Coral Spark can pass them to Coral-Schema.

rzhang10:
> Thanks! So this sounds to be another reason why we should stick to Coral-Schema. The challenge will be passing all the parameters currently passed to Coral-Schema to Coral-Spark, so Coral Spark can pass them to Coral-Schema.

Do you think there is a way to preserve casing in RelNode IR? so that Coral-schema and Coral-spark can utilize this same casing information in their transformations respectively?

rzhang10:
> Keeping two copies of the logic for `EXPR` is an example where the code can break

Yes I understand there's duplicate logic here, but I feel if this is the single inconsistency point it should be ok to manage.

wmoustafa:
> Do you think there is a way to preserve casing in RelNode IR? so that Coral-schema and Coral-spark can utilize this same casing information in their transformations respectively?

I am not sure how would that work. Coral-Schema is supposed to be the SOT for casing.

rzhang10:
> 

Whe

rzhang10:
> 

When deriving IR from Hive, Coral can also take into account the `avro.schema.literal` column names from the respective base tables? The `avro.schema.literal` is actually a Hive-specific property.

wmoustafa:
> When deriving IR from Hive, Coral can also take into account the `avro.schema.literal` column names from the respective base tables? The `avro.schema.literal` is actually a Hive-specific property.

That sounds like the definition of coral-schema which we decided to keep in a separate module. I think this design is appropriate so we separate the concerns of schema derivations from Hive to Coral IR. At the end of the day, adding a dependency on `coral-hive` and `coral-schema` is practically like depending on `coral-hive` plus this new functionality.

PR #248, opened 2022-03-30
Coral-Trino: Modify TrinoKeywordsConverter.RESERVED_KEYWORDS from Map to Set

ljfgem:
Previously, we used a map for `TrinoKeywordsConverter.RESERVED_KEYWORDS`, whose value is not used at all.
This PR is to modify it from map to set to clean the code.


funcheetah (approved):
Thanks @ljfgem

PR #250, opened 2022-04-03
Coral-Spark: Adds an API to add explicit aliases for translated SparkSql

rzhang10:
This patch adds a new main entry point API in Coral-Spark:
```
public static CoralSpark createWithAlias(RelNode irRelNode, List<String> aliases)
```
which let users feed an explicit list of aliases to the translated SQL string. This is implemented via an additional visitor `AddExplicitAlias`.

This API enables us to achieve things like aligning the SparkSql alias with Coral-Schema's output schema/field names. This will make the output SparkSql works more robustly with sparks, which expects strict alias matches.

I choose to use a `List<String>` as the interface to achieve a loose coupling of coral-schema and coral-spark, so that their dependency is bridged by the simple `List<String>`, and so that later we also have the flexibility to integrate/feed other potential schema/alias providers other than coral-schema.


coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
```diff
@@ -126,13 +126,20 @@ public Schema toAvroSchema(String dbName, String tableOrViewName) {
    * @return avro schema for a given sql query
    */
   // TODO: 2/3/22  to revisit whether we want to make it public
```
wmoustafa, line 128: Should this comment be fixed?

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
```diff
@@ -67,6 +67,23 @@ public static CoralSpark create(RelNode irRelNode) {
...
 
+  /**
+   * Users use this function as the main API for getting CoralSpark instance.
+   * This should be used when user need to explicitly add alias for the output SparkSQL string
+   *
+   * @param irRelNode An IR RelNode for which CoralSpark will be constructed.
+   * @param aliases A list of aliases for the translated SparkSQL select list
+   * @return [[CoralSparkInfo]]
+   */
+  public static CoralSpark createWithAlias(RelNode irRelNode, List<String> aliases) {
```
wmoustafa, line 78: I do not think Coral needs an external component to let it know how the alias is. This is like depending on Spark to get something from Coral-Schema. My suggestion is to preserve the old API and find from the current SparkSession whether we are in Spark 2.3 or Spark 3.1. If in Spark 3.1, return aliases from Coral schema.

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
```diff
@@ -89,6 +106,16 @@ private static String constructSparkSQL(RelNode sparkRelNode) {
     return rewritten.toSqlString(SparkSqlDialect.INSTANCE).getSql();
   }
 
+  private static String constructSparkSQLWithExplicitAlias(RelNode sparkRelNode, List<String> aliases) {
+    SparkRelToSparkSqlConverter rel2sql = new SparkRelToSparkSqlConverter();
+    // Create temporary objects r and rewritten to make debugging easier
+    SqlImplementor.Result r = rel2sql.visitChild(0, sparkRelNode);
+    SqlNode rewritten = r.asStatement().accept(new SparkSqlRewriter());
+    // Use a second pass visit to add explicit alias names
+    SqlNode aliasAdded = rewritten.accept(new AddExplicitAlias(aliases));
```
wmoustafa, line 115: Can we embed the alias addition in `SparkSqlRewriter` as we did originally?

rzhang10: `SparkSqlRewriter` also overrides `public SqlNode visit(SqlCall call)` and I feel tangling the logic there is not very clean, as the transformation there (manipulating the `CAST` might interfere with the alias manipulation) Also I only want this visitor to be applied with the new API code path.

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
rzhang10: My thought is that at the coral and spark integration site we are getting schema from coral via `ViewToAvroSchemaConverter`, so we can get the column names from coral schema there easily.

Also, I explicitly designed the API to achieve loose coupling and provide flexibility to the client.

wmoustafa: I understand that. I am saying it is not ideal. Let us discuss if needed. cc: @funcheetah.

rzhang10: Yes we can have a discussion on the pros/cons of each approach respectively.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
```diff
@@ -126,13 +126,20 @@ public Schema toAvroSchema(String dbName, String tableOrViewName) {
...
-  protected Schema toAvroSchema(String sql) {
+  public Schema toAvroSchema(String sql) {
```
funcheetah, line 129: Is this method only used in testing? If so, I'm not sure of making it public.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
```diff
@@ -126,13 +126,20 @@ public Schema toAvroSchema(String dbName, String tableOrViewName) {
...
     Preconditions.checkNotNull(sql);
 
     RelNode relNode = hiveToRelConverter.convertSql(sql);
     return relToAvroSchemaConverter.convert(relNode, false, false);
   }
 
+  public Schema toAvroSchema(String sql, boolean strictMode, boolean forceLowercase) {
```
funcheetah, line 136: Same here.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
yiqiangin: If the method is only used by the test in the same package, making it as package scope should be enough

rzhang10: The test code that uses this is in a different package so it won't work.

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
```diff
@@ -67,6 +68,29 @@ public static CoralSpark create(RelNode irRelNode) {
...
+  /**
+   * Users use this function as the main API for getting CoralSpark instance.
+   * This should be used when user need to align the Coral-spark translated SQL
+   * with Coral-schema output schema
+   *
+   * @param irRelNode An IR RelNode for which CoralSpark will be constructed.
+   * @param schema Coral schema that is represented by an Avro schema
+   * @return [[CoralSparkInfo]]
+   */
+  public static CoralSpark createWithCoralSchema(RelNode irRelNode, Schema schema) {
```
wmoustafa, line 80: Let us use the same name as the original `create` method, and consider this as an override.

rzhang10: I think you meant "overload" :)

wmoustafa (approved):

PR #251, opened 2022-04-13
Coral-as-a-Service Prototype

KevinGe00:
**Version 1 of Coral-as-a-Service:**

Main goal of this service is for users to be able to (quickly) clone the Coral repo and quickly spin up a REST API to test Coral translations with. 

Coral Service can be spun up using the production Hive Metastore Client (this requires keytab) or using a temporary local metastore (this comes with an additional endpoint for users to add tables/views to the local metastore).

**Endpoints:**
1. GET /translate 
    Query Params:
      - query: Query you want to have translated
      - fromLanguage: input SQL language of query
      - toLanguage: target SQL language you want query translated to
      
2. POST /create (note: this is only available if users choose to use Coral Service with local metastore)
    Query Params:
      - createQuery: SQL queries to create a DB/table/view (i.e "CREATE TABLE IF NOT EXISTS default.employees(e_id int, name string)")
      
**Basic instructions to start Coral Service:**
1. Clone Coral
2. `cd coral-service`
3. `./gradlew clean build`
4a. `./gradlew bootRun` to choose production Hive Metastore Client
4b. `./gradlew bootRun --args='--spring.profiles.active=localMetastore'` to choose local metastore


coral-service/src/main/java/com/linkedin/coral/coralservice/controller/TranslationController.java
```diff
@@ -0,0 +1,71 @@
...
+
+  @GetMapping("/translate")
+  public ResponseEntity translate(@RequestParam String query, @RequestParam String fromLanguage, @RequestParam String toLanguage) {
+    if (fromLanguage.equalsIgnoreCase(toLanguage)) {
+      return ResponseEntity
+          .status(HttpStatus.BAD_REQUEST)
+          .body("Please choose different languages to translate between.");
+    }
+
+    String translatedSql = "Translation not supported yet.";
```
ljfgem, line 44: May add the `fromLanguage` and `toLanguage` in the message.

coral-service/src/main/java/com/linkedin/coral/coralservice/controller/TranslationController.java
```diff
@@ -0,0 +1,71 @@
...
+      if (toLanguage.equalsIgnoreCase("spark")) {
+        translatedSql = translateHiveToSpark(query);
+      }
+      // To Trino
+      else if (toLanguage.equalsIgnoreCase("trino")) {
+        translatedSql = translateHiveToTrino(query);
+      }
+    }
+
+    String successMessage = String.format("Original query in " + fromLanguage + ": " + query + "   |   " + "Translated to " + toLanguage + ": " + translatedSql);
```
ljfgem, line 66: May use `"\n"` rather than `"   |   "` for separation.
And Redundant call to `'format()'`.

coral-service/src/main/java/com/linkedin/coral/coralservice/controller/TranslationControllerLocal.java
```diff
@@ -0,0 +1,58 @@
...
+        .body("Creation successful");
+  }
+
+  /** Method to check that query starts with "CREATE DATABASE|TABLE|VIEW" */
+  private boolean queryIsCreate(String[] splitQuery) {
+    if (!splitQuery[0].equalsIgnoreCase("create")
+        || (!splitQuery[1].equalsIgnoreCase("table") && !splitQuery[1].equalsIgnoreCase("database") && !splitQuery[1].equalsIgnoreCase("view"))) {
+      return false;
+    }
+    return true;
```
ljfgem, line 56: May just use
```
return splitQuery[0].equalsIgnoreCase("create") && (splitQuery[1].equalsIgnoreCase("table")
        || splitQuery[1].equalsIgnoreCase("database") || splitQuery[1].equalsIgnoreCase("view"));
```

coral-service/src/main/java/com/linkedin/coral/coralservice/metastore/MetastoreProvider.java
```diff
@@ -0,0 +1,125 @@
...
+  static {
+    DEFAULT_TMP_HIVE_DIR = Paths.get(System.getProperty("java.io.tmpdir"), "coral").toString();
+  }
+
+  public static HiveMetastoreClient getMetastoreClient() throws Exception {
+    final InputStream hiveConfStream =
+        CoralUtils.class.getClassLoader().getResourceAsStream("hive.properties");
+    final Properties props = new Properties();
+    props.load(hiveConfStream);
+    return new HiveMscAdapter(MetastoreProvider.getGridMetastoreClient(props));
```
ljfgem, line 57: At present, there is no `hive.properties` file, so it will fail with NPE with clean checkout.
I would recommend:
1. Add a dummy `hive.properties` file
2. If `hive.properties` doesn't exist or fails to instantiate grid HMS client (may use try catch), we can warn the user and fall back to the local HMS.

coral-service/src/main/java/com/linkedin/coral/coralservice/metastore/MetastoreProvider.java
```diff
@@ -0,0 +1,125 @@
...
+   * Get metastore client for the hive metastore on Grid. Supports kerberos based
+   * authentication. Using kerberos requires user to get kerberos ticket and setup
+   * keytab externally.
+   * @param props hive configuration to use for connecting to the metastore
+   * @return metastore client
+   * @throws IOException indicates failure to read keytab file
+   * @throws MetaException indicates failure to instantiate metastore client
+   * @throws ConfigurationException on missing required configuration parameter
+   */
+  public static IMetaStoreClient getGridMetastoreClient(Properties props)
```
ljfgem, line 70: Can this method be private?

coral-service/src/main/java/com/linkedin/coral/coralservice/metastore/MetastoreProvider.java
```diff
@@ -0,0 +1,125 @@
...
+
+  /**
+   * Get hive metastore client for local hive. Uses locally setup database
+   * and uses embedded derby to connect.
+   * @param props properties to connect to the local store
+   * @return Local hive metastore client
+   * @throws HiveException on failure to instantiate local hive
+   * @throws MetaException on failure to create local hive metastore
+   */
+  public static IMetaStoreClient getLocalMetastoreClient(Properties props) throws HiveException, MetaException {
```
ljfgem, line 111: Can this method be private?

ljfgem (commented):
Nice work! Thanks @KevinGe00 for this PR!
@wmoustafa @funcheetah PTAL

coral-service/src/main/java/com/linkedin/coral/coralservice/controller/TranslationController.java
```diff
@@ -0,0 +1,71 @@
...
+    // runs after the Spring context has been initialized
+    try {
+      initHiveMetastoreClient();
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+  }
+
+  @GetMapping("/translate")
+  public ResponseEntity translate(@RequestParam String query, @RequestParam String fromLanguage, @RequestParam String toLanguage) {
```
yiqiangin, line 37: It is better to annotate these parameters with @nonnull to proactively prevent null value passing into the function body.

coral-service/src/main/java/com/linkedin/coral/coralservice/controller/TranslationControllerLocal.java
```diff
@@ -0,0 +1,58 @@
...
+    // runs after the Spring context has been initialized
+    try {
+      CoralUtils.initLocalMetastore();
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+  }
+
+  @PostMapping("/create")
+  public ResponseEntity createInLocalMetastore(@RequestParam String createQuery) {
```
yiqiangin, line 34: same here

coral-service/src/main/java/com/linkedin/coral/coralservice/metastore/MetastoreProvider.java
```diff
@@ -0,0 +1,125 @@
...
+import static org.apache.hadoop.hive.conf.HiveConf.ConfVars.*;
+
+
+/**
+ * Utility class to get Hive metastore client for local hive or grid based hive using kerberos
+ */
+@SuppressWarnings("WeakerAccess")
+public class MetastoreProvider {
+
+  public static final String HIVE_METASTORE_URIS = "hive.metastore.uris";
```
yiqiangin, line 33: it is better to declare these constants as private, since they are not used outside this class.

coral-service/src/main/java/com/linkedin/coral/coralservice/utils/CoralUtils.java
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Utility class to provide Coral functionality to Coral Service.
+ */
+public class CoralUtils {
+
+  public static HiveMetastoreClient hiveMetastoreClient;
+
+  // Converters
+  public static HiveToRelConverter hiveToRelConverter;
```
yiqiangin, line 32: same here, public --> private

coral-service/src/main/java/com/linkedin/coral/coralservice/utils/CoralUtils.java
```diff
@@ -0,0 +1,87 @@
...
+    hiveMetastoreClient  = MetastoreProvider.getMetastoreClient();
+    hiveToRelConverter = new HiveToRelConverter(hiveMetastoreClient);
+    trinoToRelConverter  = new TrinoToRelConverter(hiveMetastoreClient);
+  }
+
+  public static void initLocalMetastore() throws IOException, HiveException, MetaException {
+    // Create a temporary local metastore
+    conf = loadResourceHiveConf();
+    String tempDir = conf.get(CORAL_SERVICE_DIR);
+    System.out.println("Temp Workspace: " + tempDir);
```
yiqiangin, line 52: Why not using log? it is not usual to use stdout to output message.

coral-service/src/main/java/com/linkedin/coral/coralservice/utils/CoralUtils.java
```diff
@@ -0,0 +1,87 @@
...
+    FileUtils.deleteDirectory(new File(tempDir));
```
yiqiangin, line 53: I suggest to catch the IOException here in case that the exception is caused by the non-existing directory, while the process can still go on in this case.

coral-service/src/main/java/com/linkedin/coral/coralservice/controller/TranslationController.java
```diff
@@ -0,0 +1,71 @@
...
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+
+/**
+ * Class defining the REST endpoints to translate queries using Coral.
+ */
+@RestController
+@Service
+@Profile({"hiveMetastoreClient", "default"})
```
wmoustafa, line 23: Could you rename to `remoteMetastore`?

wmoustafa (commented):
Thanks for this PR! Great work. Could you add the instructions as part of the `README` file? Also, could you please make it more detailed, with examples, for both remote and local HMS?

coral-service/src/main/java/com/linkedin/coral/coralservice/utils/CoralUtils.java
```diff
@@ -0,0 +1,87 @@
...
+
+  public static void run(Driver driver, String sql) {
+    while (true) {
+      try {
+        CommandProcessorResponse result = driver.run(sql);
+        if (result.getException() != null) {
+          throw new RuntimeException("Execution failed for: " + sql, result.getException());
+        }
+      } catch (CommandNeedRetryException e) {
+        continue;
```
yiqiangin, line 70: I suggest to use the library of FailSafe to implement the retry logic, which makes the code more concise. Also does it need to wait for sometime to retry?

coral-service/src/main/resources/application.properties
```diff
@@ -0,0 +1 @@
+
```
ljfgem, line 1: This file can be removed?

coral-service/build.gradle
```diff
@@ -0,0 +1,62 @@
+plugins {
+  id 'org.springframework.boot' version '2.1.2.RELEASE'
+  id 'io.spring.dependency-management' version '1.0.11.RELEASE'
+  id 'java'
+}
+
+group = 'com.linkedin.coral'
+version = '0.0.1-SNAPSHOT'
```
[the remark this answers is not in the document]
KevinGe00: It's specific to Springboot, it's used for "multi-module" Springboot apps which Coral Service technically is. https://spring.io/guides/gs/multi-module/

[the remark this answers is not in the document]
wmoustafa: Who updates the version number? On what occasion? Who consumes it?

coral-service/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+
+  // Coral
+  implementation 'com.linkedin.coral:coral-common:2.0.62'
+  implementation 'com.linkedin.coral:coral-trino:2.0.61'
+  implementation 'com.linkedin.coral:coral-hive:2.0.61'
+  implementation 'com.linkedin.coral:coral-pig:2.0.61'
+  implementation ('com.linkedin.coral:coral-spark:2.0.61') {
+    exclude group: 'org.slf4j'
+  }
+  compile('com.linkedin.coral:coral-trino-parser:2.0.61') {
```
HotSushi, line 38: we should stick to `Implementation` for uniformity.

coral-service/gradlew
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/env sh
```
HotSushi, line 1: include `:coral-service` in `ROOT_DIR/settings.gradle` file. You shouldnt need another set of gradlewrappers here.
If the reasoning behind this was a older gradle version, maybe update the gradle version for main project.

coral-service/settings.gradle
```diff
@@ -0,0 +1,9 @@
+pluginManagement {
+  repositories {
+    maven { url 'https://repo.spring.io/milestone' }
+    maven { url 'https://repo.spring.io/snapshot' }
+    gradlePluginPortal()
+  }
+}
```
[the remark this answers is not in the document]
KevinGe00: So the `settings.gradle` file is to define all included submodules, so you can only have one `settings.gradle` file per multi-module project (i.e only one in the entire Coral repo). I've discussed this with Lei but we had issues configuring the `coral-service` module as a submodule of `coral`, which would be ideal. Ultimately, this means in the meantime we need this `settings.gradle` file in `coral-service` because it's being treated as it's own multi-module project.

Integrating `coral-service` as a submodule of `coral` should be considered in future work.

coral-service/src/main/java/com/linkedin/coral/coralservice/utils/CoralUtils.java
```diff
@@ -0,0 +1,87 @@
...
+      }
+      break;
+    }
+  }
+
+  public static HiveConf loadResourceHiveConf() {
```
HotSushi, line 76: Provide this as a `bean` instead of a static function. Dependency injection is the most powerful feature of springboot.

coral-service/src/main/java/com/linkedin/coral/coralservice/utils/CoralUtils.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.hadoop.hive.ql.Driver;
+import org.apache.hadoop.hive.ql.metadata.Hive;
+import org.apache.hadoop.hive.ql.metadata.HiveException;
+import org.apache.hadoop.hive.ql.processors.CommandProcessorResponse;
+import org.apache.hadoop.hive.ql.session.SessionState;
+
+/**
+ * Utility class to provide Coral functionality to Coral Service.
+ */
+public class CoralUtils {
```
HotSushi, line 27: This class seems to be a provider for `HiveMetastoreClient`, `HiveToRelConverter`, `TrinoToRelConverter`. Maybe rename this to CoralProvider and provide `beans` for these objects.  Look into `@Configuration` tags here: https://spring.io/guides/gs/spring-boot/

coral-service/src/main/java/com/linkedin/coral/coralservice/metastore/MetastoreProvider.java
HotSushi, line 111: Similar feedback here, about using @beans.

Two objects here `GridMetastoreClient` and `LocalMetastoreClient`, one interface `IMetaStoreclient`. While booting the spring application, a decision can be made based on a configuration file (say coralconfigs.yaml) and appropriate object should be injected.

coral-service/src/main/java/com/linkedin/coral/coralservice/controller/TranslationControllerLocal.java
KevinGe00: > May just use
> 
> ```
> return splitQuery[0].equalsIgnoreCase("create") && (splitQuery[1].equalsIgnoreCase("table")
>         || splitQuery[1].equalsIgnoreCase("database") || splitQuery[1].equalsIgnoreCase("view"));
> ```

Much cleaner, thank you!

HotSushi (commented):
Excellent POC @KevinGe00 🎉, really excited to see this go forward.

First cut looks fantastic, some minor feedback for the next cut:
1) Layering the code in a MVC way will be helpful. controller does request validation, service does the actual translation.
2) Use dependency injection extensively. 
3) Provide a docker setup for people to play with.

coral-service/src/main/java/com/linkedin/coral/coralservice/metastore/MetastoreProvider.java
KevinGe00: Adding a dummy `hive.properties`. This seems a lot more user friendly.

coral-service/src/main/java/com/linkedin/coral/coralservice/controller/TranslationController.java
KevinGe00: Springboot makes RequestParams required by default!

I tested it out to make sure so if I send in a request with a param missing I get:
```
{
    "timestamp": "2022-04-14T19:42:45.115+0000",
    "status": 400,
    "error": "Bad Request",
    "message": "Required String parameter 'toLanguage' is not present",
    "path": "/translate"
}
```

coral-service/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+group = 'com.linkedin.coral'
+version = '0.0.1-SNAPSHOT'
+sourceCompatibility = '1.8'
+
+repositories {
+  mavenCentral()
+  maven { url 'https://repo.spring.io/milestone' }
+  maven { url 'https://repo.spring.io/snapshot' }
+  maven { url "http://conjars.org/repo" }
+}
```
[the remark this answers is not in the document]
KevinGe00: Turns out line 14 isn't required, but the purpose of this section is to direct gradle to the repositories to pull the required jars.

coral-service/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+  exclude group: 'javax.servlet'
+}
+
+dependencies {
+  // Spring Boot
+  implementation 'org.springframework.boot:spring-boot-starter-web'
+  testImplementation 'org.springframework.boot:spring-boot-starter-test:' + '2.6.1'
+
+  // Coral
+  implementation 'com.linkedin.coral:coral-common:2.0.62'
```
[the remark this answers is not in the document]
KevinGe00: That syntax is how you add dependencies of other submodules (i.e coral-trino depends on coral-hive and they are both submodules of the root coral repo). Once again, coral-service isn't a submodule of the root coral project and so we must depend on coral jars that are pulled from Maven repo.

coral-service/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+  implementation 'com.linkedin.coral:coral-pig:2.0.61'
+  implementation ('com.linkedin.coral:coral-spark:2.0.61') {
+    exclude group: 'org.slf4j'
+  }
+  compile('com.linkedin.coral:coral-trino-parser:2.0.61') {
+    artifact {
+      name = 'coral-trino-parser'
+      extension = 'jar'
+      type = 'jar'
+    }
```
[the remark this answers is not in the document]
KevinGe00: For some reason `coral-trino-parser` isn't pulled unless this artifact section is specified.

coral-service/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+  // Hadoop
+  implementation 'org.apache.hadoop:hadoop-common:2.7.0'
+  implementation 'org.apache.hadoop:hadoop-mapreduce-client-common:2.7.0'
+  implementation 'org.apache.hadoop:hadoop-mapreduce-client-core:2.7.0'
+
+}
+
+tasks.named('test') {
+  useJUnitPlatform()
+}
```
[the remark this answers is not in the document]
KevinGe00: It's not required since we've decided to take out the basic unit tests that come with Spring, I've deleted this part.

coral-service/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.8-bin.zip
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
```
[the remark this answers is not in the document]
KevinGe00: Same reason as here https://github.com/linkedin/coral/pull/251#discussion_r850689047

coral-service/gradlew
```diff
@@ -0,0 +1,185 @@
...
+
+#
+# Copyright 2015 the original author or authors.
```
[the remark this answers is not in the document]
KevinGe00: Same reason as here https://github.com/linkedin/coral/pull/251#discussion_r850689047

coral-service/gradlew.bat
```diff
@@ -0,0 +1,89 @@
+@rem
+@rem Copyright 2015 the original author or authors.
+@rem
+@rem Licensed under the Apache License, Version 2.0 (the "License");
+@rem you may not use this file except in compliance with the License.
```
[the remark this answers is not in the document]
KevinGe00: Same reason as here https://github.com/linkedin/coral/pull/251#discussion_r850689047

coral-service/settings.gradle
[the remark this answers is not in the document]
ljfgem: @wmoustafa I also think it's reasonable to treat `coral-service` as a separate module instead of a coral submodule, given `coral-service` is not like other coral modules which do translation.

coral-service/gradlew
KevinGe00: This will be a part of [this discussion](https://github.com/linkedin/coral/pull/251#discussion_r852461057), including `coral-service` will add it as a submodule.

coral-service/src/main/java/com/linkedin/coral/coralservice/utils/CoralUtils.java
KevinGe00: Thanks for the comment Yiqiang, I looked into Failsafe and it seems there's no way to set the retry condition be on an exception (which is the case where, we retry on CommandNeedRetryException. Please let me know if I'm miss understanding Failsafe, https://failsafe.dev/retry/ this is what I'm referring to.

coral-service/build.gradle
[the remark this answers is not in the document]
KevinGe00: Sorry turns out this is not Spring boot specific, https://stackoverflow.com/questions/23354243/what-is-group-property-for-in-gradle

It's simply just declaring the groupId:artifact:version, so we would update/consume these as such.

coral-service/src/main/java/com/linkedin/coral/coralservice/utils/CoralUtils.java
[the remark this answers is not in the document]
KevinGe00: I haven't updated this yet because I'll be following up on this https://github.com/linkedin/coral/pull/251#discussion_r850696632

coral-service/README.md
```diff
@@ -0,0 +1,42 @@
...
+2. From the root directory of Coral, access the coral-service module
+```
+cd coral-service
+```
+3. Build
+```
+./gradlew clean build
+```
+To run Coral Service using the **production metastore** (Hive Metastore Client):
+4. a) Add kerberos client keytab file `daliview.headless.keytab` to `coral-service/src/main/resources`
```
ljfgem, line 26: Add kerberos client keytab file to `coral-service/src/main/resources`

coral-service/src/main/java/com/linkedin/coral/coralservice/metastore/MetastoreProvider.java
```diff
@@ -0,0 +1,95 @@
...
+      String clientKeytab = props.getProperty(HIVE_METASTORE_CLIENT_KEYTAB);
+      if (Strings.isNullOrEmpty(clientKeytab)) {
+        throw new ConfigurationException(String.format("%s is required", HIVE_METASTORE_CLIENT_KEYTAB));
+      }
+      String servicePrincipal = props.getProperty(HIVE_METASTORE_SERVICE_PRINCIPAL);
+      if (Strings.isNullOrEmpty(servicePrincipal)) {
+        throw new ConfigurationException(String.format("%s is required", HIVE_METASTORE_SERVICE_PRINCIPAL));
+      }
+      // leave this commented...handy debugging option rather than searching
+      // System.setProperty("sun.security.krb5.debug", "true");
```
ljfgem, line 85: Remove these lines.

coral-service/src/main/java/com/linkedin/coral/coralservice/utils/CoralUtils.java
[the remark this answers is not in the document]
ljfgem: Let's rename it for now, https://github.com/linkedin/coral/pull/251#discussion_r850696632 and the other SpringBoot specific comments can be addressed in the following PRs if we still have time.

ljfgem (approved):
Thanks @KevinGe00!
Left several nits, and I am good with else.

yiqiangin (approved):

coral-service/README.md
```diff
@@ -0,0 +1,42 @@
...
+2. From the root directory of Coral, access the coral-service module
+```
+cd coral-service
+```
+3. Build
+```
+./gradlew clean build
+```
+To run Coral Service using the **production metastore** (Hive Metastore Client):
+4. a) Add your kerberos client keytab file to `coral-service/src/main/resources`
```
wmoustafa, line 26: Is there a better location than the source code? Where is the service deployed? Can we leverage that directory instead, or some conf directory that is relative to the service?

coral-service/README.md
```diff
@@ -0,0 +1,42 @@
+# Coral-as-a-Service
```
wmoustafa, line 1: How about we add this in the main README?

coral-service/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+**Coral-as-a-Service** or simply, **Coral Service** is a REST API allowing users to make SQL translations using Coral. The service can be spun up and connected to the production Hive Metastore Client or connected a temporary local metastore (this comes with an additional endpoints).
+
+## API Reference
+
+| Endpoint | Query Params |Example | Method | Semantics
+|--|--|--|--|--|
+| /translate |  <ul><li><strong>query:</strong> Query you want to have translated</li><li><strong>fromLanguage:</strong> input SQL language of query</li><li><strong> toLanguage:</strong> target SQL language you want query translated to</li></ul> | /translate?query=select%20strpos%28%27foobar%27%2C%20%27b%27%29%20as%20pos&fromLanguage=trino&toLanguage=spark </br> **Note:** 'select%20strpos%28%27foobar%27%2C%20%27b%27%29%20as%20pos' is URL decoded as 'select strpos('foobar', 'b') as pos' | GET |   READ - returns the translated query|
```
wmoustafa, line 9: Can we write the code using a monospaced font?

coral-service/README.md
```diff
@@ -0,0 +1,42 @@
+# Coral-as-a-Service
+
+**Coral-as-a-Service** or simply, **Coral Service** is a REST API allowing users to make SQL translations using Coral. The service can be spun up and connected to the production Hive Metastore Client or connected a temporary local metastore (this comes with an additional endpoints).
```
wmoustafa, line 3: ```**Coral-as-a-Service** or simply, **Coral Service** is a service that exposes REST APIs that allow users to interact with Coral without necessarily coming from a compute engine. Currently, the service supports an API for query translation between different dialects and another for interacting with a local Hive Metastore to create example databases, tables, and views so they can be referenced in the translation API. The service can be used in two modes: remote Hive Metastore mode, and local Hive Metastore mode. The remote mode uses an existing (already deployed) Hive Metastore to resolve tables and views, while the local one creates an empty embedded Hive Metastore so users can add their own table and view definitions```.

coral-service/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+| Endpoint | Query Params |Example | Method | Semantics
+|--|--|--|--|--|
+| /translate |  <ul><li><strong>query:</strong> Query you want to have translated</li><li><strong>fromLanguage:</strong> input SQL language of query</li><li><strong> toLanguage:</strong> target SQL language you want query translated to</li></ul> | /translate?query=select%20strpos%28%27foobar%27%2C%20%27b%27%29%20as%20pos&fromLanguage=trino&toLanguage=spark </br> **Note:** 'select%20strpos%28%27foobar%27%2C%20%27b%27%29%20as%20pos' is URL decoded as 'select strpos('foobar', 'b') as pos' | GET |   READ - returns the translated query|
+| /create  | <ul><li><strong>createQuery:</strong> SQL query to create a DB/table/view</li> | /create?createQuery=CREATE%20TABLE%20IF%20NOT%20EXISTS%20default.employees(e_id%20int%2C%20name%20string) **Note:** 'CREATE%20TABLE%20IF%20NOT%20EXISTS%20default.employees(e_id%20int%2C%20name%20string)' is URL decoded as 'CREATE TABLE IF NOT EXISTS default.employees(e_id int, name string)' | POST | CREATE - creates the metadata of the DB/table/view in the local metastore (**note:** this endpoint is only available if users choose to use Coral Service with local metastore) |
```
wmoustafa, line 10: Can we say that this API is only available in the local Hive Metastore mode? I understand it is mentioned below, but it may be good to mention here as well.

coral-service/README.md
wmoustafa: Please use remote as opposed to production and local as opposed to temporary.

coral-service/README.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+To run Coral Service using the temporary **local metastore**:
+4. a) Run
+```
+./gradlew bootRun --args='--spring.profiles.active=localMetastore'
+```
+
+## Currently Supported Translation Flows
+1. Trino to Spark
```
wmoustafa, line 40: I think we should point out that views referenced in queries are treated as defined in HiveQL, and hence cannot be used when translating a view from Trino. Only referencing base tables are supported in Trino queries.
We should also point out that this path is in POC stage (we may split it out from the list, or better, we may convert that section to a free-form paragraph).

coral-service/README.md
```diff
@@ -0,0 +1,42 @@
...
+4. b) Appropriately replace all instances of `SET_ME` in `coral-service/src/main/resources/hive.properties`
+4. c) Run
+```
+./gradlew bootRun
+```
+
+To run Coral Service using the temporary **local metastore**:
+4. a) Run
+```
+./gradlew bootRun --args='--spring.profiles.active=localMetastore'
```
wmoustafa, line 36: I feel we need to add more end to end example in the local HMS case. (like, create a database, create a table, issue a query translation API). You might also want to show the response.

coral-service/README.md
```diff
@@ -0,0 +1,42 @@
...
+```
+cd coral-service
+```
+3. Build
+```
+./gradlew clean build
+```
+To run Coral Service using the **production metastore** (Hive Metastore Client):
+4. a) Add your kerberos client keytab file to `coral-service/src/main/resources`
+4. b) Appropriately replace all instances of `SET_ME` in `coral-service/src/main/resources/hive.properties`
```
wmoustafa, line 27: It is better to describe the remote one after the local one since it is more constrained, so we can reference the local one when describing the remote one.

coral-service/build.gradle
```diff
@@ -0,0 +1,52 @@
...
+  exclude group: 'javax.servlet'
+}
+
+dependencies {
+  // Spring Boot
+  implementation 'org.springframework.boot:spring-boot-starter-web'
+  testImplementation 'org.springframework.boot:spring-boot-starter-test:2.6.1'
+
+  // Coral
+  implementation 'com.linkedin.coral:coral-common:+'
```
wmoustafa, line 29: Will this grab the last version published as opposed to the current local code? Is there a way to do the latter?

coral-service/build.gradle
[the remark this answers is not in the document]
wmoustafa: Can we add a comment stating this in the code?

coral-service/README.md
KevinGe00: Do you mean adding Coral Service's entire README in Coral's README, or do you mean simply adding "coral-service" under "Modules" in [Coral's README](https://github.com/linkedin/coral/blob/master/README.md)?

wmoustafa: The former.

ljfgem:
Closing it given #262 has been merged.

PR #253, opened 2022-04-14
Remove unnecessary files specific to LinkedIn

ljfgem:
Since it's an open-source project, these LI specific files should be removed.


funcheetah (approved):
Thanks @ljfgem !

PR #255, opened 2022-04-19
Manually register missing UDFs

aastha25:
Manually registering several more UDFs. 

Testing Done:
Built successfully. Spotless check successful. 
Ran with views that were previously failing. 

Also fixed the return type for previously added UDF IsValidKeyUDF.

### DIFF
--- a/coral-common/src/main/java/com/linkedin/coral/common/functions/FunctionReturnTypes.java
+++ b/coral-common/src/main/java/com/linkedin/coral/common/functions/FunctionReturnTypes.java
@@ -41,10 +41,12 @@ public final class FunctionReturnTypes {
     return opBinding.isOperandLiteral(1, false) ? type2 : type1;
   };
 
+  public static final SqlReturnTypeInference BOOLEAN = ReturnTypes.explicit(SqlTypeName.BOOLEAN);
   public static final SqlReturnTypeInference STRING = ReturnTypes.explicit(SqlTypeName.VARCHAR);
   public static final SqlReturnTypeInference BINARY = ReturnTypes.explicit(SqlTypeName.BINARY);
   public static final SqlReturnTypeInference BIGINT = ReturnTypes.explicit(SqlTypeName.BIGINT);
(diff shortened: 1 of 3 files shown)


coral-common/src/main/java/com/linkedin/coral/common/functions/FunctionReturnTypes.java
```diff
@@ -41,10 +41,12 @@ private FunctionReturnTypes() {
     return opBinding.isOperandLiteral(1, false) ? type2 : type1;
   };
 
+  public static final SqlReturnTypeInference BOOLEAN = ReturnTypes.explicit(SqlTypeName.BOOLEAN);
   public static final SqlReturnTypeInference STRING = ReturnTypes.explicit(SqlTypeName.VARCHAR);
   public static final SqlReturnTypeInference BINARY = ReturnTypes.explicit(SqlTypeName.BINARY);
   public static final SqlReturnTypeInference BIGINT = ReturnTypes.explicit(SqlTypeName.BIGINT);
   public static final SqlReturnTypeInference SMALLINT = ReturnTypes.explicit(SqlTypeName.SMALLINT);
+  public static final SqlReturnTypeInference INTEGER = ReturnTypes.explicit(SqlTypeName.INTEGER);
```
ljfgem, line 49: May just use `ReturnTypes.BOOLEAN` and `ReturnTypes.INTEGER` provided by calcite directly, so we don't need to add these 2 types here.
Actually I found `DATE` and `BIGINT` are also defined in `org.apache.calcite.sql.type.ReturnTypes`, could you help to remove them from this class too?

ljfgem (commented):
Thanks @aastha25 for this PR, left one nit.

coral-common/src/main/java/com/linkedin/coral/common/functions/FunctionReturnTypes.java
aastha25: Thanks for suggesting it. Updated the references. 
Also rebased.

ljfgem (approved):
LGTM, thanks @aastha25!

PR #256, opened 2022-04-19
Bump com.diffplug.spotless to 5.9.0 to fix a transient issue

ljfgem:
I met a transient issue similar to https://github.com/diffplug/spotless/issues/741 several times while building locally, bumping the version can fix it according to the discussion.


funcheetah (approved):
Thanks @ljfgem !

PR #257, opened 2022-04-19
Bump jobs-udf

rzhang10:
Bump jobs-udf to include the fix https://github.com/linkedin/transport/pull/115 for those transport udfs in spark.


ljfgem (approved):

funcheetah (approved):
Thanks @rzhang10 !

PR #260, opened 2022-04-28
Coral-Spark: Handle select star case for AddExplicitAlias

rzhang10:
in pr #250 I miss handling the special case of select star queries, for select star queries, we don't want to explicitly add any aliases and should just return select star as is.

The select star special case is distinguished by the `SqlSelect`'s `selectList` being `null`. If we don't explicitly handle this case, line https://github.com/linkedin/coral/blob/efc84d80b93844ba17715946eee65c87dc8e8cdb/coral-spark/src/main/java/com/linkedin/coral/spark/AddExplicitAlias.java#L42 will throw an NPE.

I also add a unit test case for this special case.

### DIFF
--- a/coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
+++ b/coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
@@ -120,14 +120,13 @@ public class CoralSpark {
     // Create temporary objects r and rewritten to make debugging easier
     SqlImplementor.Result r = rel2sql.visitChild(0, sparkRelNode);
     SqlNode rewritten = r.asStatement().accept(new SparkSqlRewriter());
-    SqlNode ret = rewritten;
     // Use a second pass visit to add explicit alias names,
     // only do this when it's not a select star case,
     // since for select star we don't need to add any explicit aliases
     if (rewritten.getKind() == SqlKind.SELECT && ((SqlSelect) rewritten).getSelectList() != null) {
-      ret = rewritten.accept(new AddExplicitAlias(aliases));
+      rewritten = rewritten.accept(new AddExplicitAlias(aliases));
     }
-    return ret.toSqlString(SparkSqlDialect.INSTANCE).getSql();
+    return rewritten.toSqlString(SparkSqlDialect.INSTANCE).getSql();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
```diff
@@ -118,9 +120,14 @@ private static String constructSparkSQLWithExplicitAlias(RelNode sparkRelNode, L
...
-    SqlNode aliasAdded = rewritten.accept(new AddExplicitAlias(aliases));
-    return aliasAdded.toSqlString(SparkSqlDialect.INSTANCE).getSql();
+    SqlNode ret = rewritten;
+    // Use a second pass visit to add explicit alias names,
+    // only do this when it's not a select star case,
+    // since for select star we don't need to add any explicit aliases
+    if (rewritten.getKind() == SqlKind.SELECT && ((SqlSelect) rewritten).getSelectList() != null) {
+      ret = rewritten.accept(new AddExplicitAlias(aliases));
+    }
+    return ret.toSqlString(SparkSqlDialect.INSTANCE).getSql();
```
ljfgem, line 130: Can we just use `rewritten` for simplicity? like:
```
    SqlNode rewritten = r.asStatement().accept(new SparkSqlRewriter());
    // Use a second pass visit to add explicit alias names,
    // only do this when it's not a select star case,
    // since for select star we don't need to add any explicit aliases
    if (rewritten.getKind() == SqlKind.SELECT && ((SqlSelect) rewritten).getSelectList() != null) {
      rewritten.accept(new AddExplicitAlias(aliases));
    }
    return rewritten.toSqlString(SparkSqlDialect.INSTANCE).getSql();
```

ljfgem (commented):
Thanks @rzhang10 for this PR, left 2 nits.

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
rzhang10: Sure, refactored.

coral-spark/src/test/java/com/linkedin/coral/spark/CoralSparkTest.java
```diff
@@ -815,6 +815,15 @@ public void tesCTEViewWithAlias() {
...
+  @Test
+  public void testSelectStar() {
+    String sourceSql = "SELECT * FROM default.basecomplex";
+    String expandedSql = getCoralSparkTranslatedSqlWithAliasFromCoralSchema(sourceSql);
+
+    String targetSql = "SELECT *\n" + "FROM default.basecomplex";
+    assertEquals(expandedSql, targetSql);
+  }
+
   private static String getCoralSparkTranslatedSqlWithAliasFromCoralSchema(String db, String view) {
```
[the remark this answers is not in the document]
rzhang10: I think given this is a useful test utility method it's fine to leave it there for future use.

funcheetah (approved):
Thanks @rzhang10 !

PR #264, opened 2022-05-03
Coral-Schema: Add useDaliRowSchema flag

ljfgem:
This PR is to add `useDaliRowSchema` flag to Coral-Schema given `dali.row.schema` property is a LinkedIn specific feature.
The default value of `useDaliRowSchema` is `false`, so we don't use schema from `dali.row.schema` by default for views, and we don't need to change the code in Spark side.

I know the concern of adding more and more flags to Coral-Schema, we may refactor it to a map like how we did for Coral-Trino in `CoralTrinoConfigKeys` in the future.

Tested on Spark gateway, worked for the affected view.


ljfgem:
Closing it given #265 has been merged.

Issue #266, opened 2022-05-09
Coral transform hive function “regexp_extract” to trino by introduce a new function "hive_pattern_to_trino"


gjhkael:
Coral transform the following hive sql to trino
```shell
select regexp_extract('1a 2b 14m', '\d+', 1); 
```

result:
```shell
select regexp_extract('1a 2b 14m', "hive_pattern_to_presto"('\d+'), 1); 
```

But trino not have the function name of "hive_pattern_to_presto", then the query will failed with exception:
```shell
Function 'hive_pattern_to_trino' not registered
```

So, why are you introduce the "hive_pattern_to_trino" function? Do this function is implement in linkedin internal?

PR #267, opened 2022-05-09
Translate to Trino properly `regexp_extract` function

findinpath:
Fixes: #266

Context: 

`regexp_extract` function in Hive https://www.revisitclass.com/hadoop/regexp_extract-function-in-hive-with-examples/
`regexp_extract` function in Trino https://trino.io/docs/current/functions/regexp.html#regexp_extract

Trino signature matches Hive's signature for `regexp_extract` function.

Remove from the code the unused reference `hive_pattern_to_trino`.

### DIFF
--- a/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/CalciteTrinoUDFMap.java
+++ b/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/CalciteTrinoUDFMap.java
@@ -62,8 +62,7 @@ public class CalciteTrinoUDFMap {
     createUDFMapEntry(UDF_MAP, hiveToCalciteOp("hex"), 1, "to_hex");
     createUDFMapEntry(UDF_MAP, hiveToCalciteOp("unhex"), 1, "from_hex");
     createUDFMapEntry(UDF_MAP, hiveToCalciteOp("array_contains"), 2, "contains");
-    createUDFMapEntry(UDF_MAP, hiveToCalciteOp("regexp_extract"), 3, "regexp_extract",
-        "[{\"input\": 1}, {\"op\": \"hive_pattern_to_trino\", \"operands\":[{\"input\": 2}]}, {\"input\": 3}]", null);
+    createUDFMapEntry(UDF_MAP, hiveToCalciteOp("regexp_extract"), 3, "regexp_extract");
     createUDFMapEntry(UDF_MAP, hiveToCalciteOp("instr"), 2, "strpos");
     createRuntimeUDFMapEntry(UDF_MAP, hiveToCalciteOp("decode"), 2,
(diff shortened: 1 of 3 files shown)


coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/CalciteTrinoUDFMap.java
```diff
@@ -62,8 +62,7 @@ private CalciteTrinoUDFMap() {
     createUDFMapEntry(UDF_MAP, hiveToCalciteOp("hex"), 1, "to_hex");
     createUDFMapEntry(UDF_MAP, hiveToCalciteOp("unhex"), 1, "from_hex");
     createUDFMapEntry(UDF_MAP, hiveToCalciteOp("array_contains"), 2, "contains");
-    createUDFMapEntry(UDF_MAP, hiveToCalciteOp("regexp_extract"), 3, "regexp_extract",
-        "[{\"input\": 1}, {\"op\": \"hive_pattern_to_trino\", \"operands\":[{\"input\": 2}]}, {\"input\": 3}]", null);
```
losipiuk, line 66: Curious what was the logic in `hive_pattern_to_trino` which we are missing. Should we add a hidden function `hive_pattern_to_trino` to Trino instead?

hackeryang: > Curious what was the logic in `hive_pattern_to_trino` which we are missing. Should we add a hidden function `hive_pattern_to_trino` to Trino instead?

Here is the issue we proposed, we think that maybe the udf logic is in the internal version of linkedin presto:  
https://github.com/trinodb/trino/issues/12288

ljfgem: Hi @findinpath @hackeryang @losipiuk, LinkedIn Trino has an internal udf `hive_pattern_to_trino` which converts the Hive regexp pattern to Trino regexp pattern. This function currently replaces consecutive backslashes with single backslash given Hive pattern uses double slash (\\\\) for escaping whereas trino requires single \\.

So I think it would be better to add this function to Trino.

hackeryang: > Hi @findinpath @hackeryang @losipiuk, LinkedIn Trino has an internal udf `hive_pattern_to_trino` which converts the Hive regexp pattern to Trino regexp pattern. This function currently replaces consecutive backslashes with single backslash given Hive pattern uses double slash (\\) for escaping whereas trino requires single \.
> 
> So I think it would be better to add this function to Trino.

Thank you for your reply, @ljfgem . Can you please contribute this udf to the Trino community? We will appreciate it if the community can have this feature.

findinpath:
Closing the PR because it looks like Coral has an actual purpose for `hive_pattern_to_trino` . Looking forward to the corresponding PR in Trino for adding the UDF to the Trino community.

PR #269, opened 2022-05-11
Coral-Common: Delete HiveSchemaUtils

ljfgem:
This class is not used anywhere (also checked internally).

Tested by building it.


aastha25:
Thanks for the PR! LGTM

PR #270, opened 2022-05-16
Modify Coral-Service endpoints to be RESTful style

ljfgem:
Given RESTful endpoints are usually nouns instead of verbs, this PR is to modify these endpoints.

@wmoustafa @aastha25 PTAL and let me know your thoughts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ curl -X POST "http://localhost:8080/api/catalog-ops/execute?statement=CREATE%20T
 ```bash
 # Non-URL encoded statement: SELECT * FROM db1.airport
 
-curl -X POST "http://localhost:8080//api/translations/translate?fromLanguage=hive&toLanguage=trino&query=SELECT%20%2A%20FROM%20db1.airport"
+curl -X POST "http://localhost:8080/api/translations/translate?fromLanguage=hive&toLanguage=trino&query=SELECT%20%2A%20FROM%20db1.airport"
 ```
 The translation result is:
 ```
--- a/README.md
+++ b/README.md
@@ -65,14 +65,14 @@ Please see the [Contribution Agreement](CONTRIBUTING.md).
 
 ### API Reference
 
-#### /translate
-A **GET** API which takes the following parameters and returns the translated query:
-- `query`: SQL query to translate between two dialects
+#### /api/translations/translate
+A **POST** API which takes the following parameters and returns the translated query:
 - `fromLanguage`: Input dialect (e.g., spark, trino, hive -- see below for supported inputs)
 - `toLanguage`: Output dialect (e.g., spark, trino, hive -- see below for supported outputs)
+- `query`: SQL query to translate between two dialects
 
-#### /create
-A **POST** API which takes a SQL query to create a database/table/view in the local metastore
+#### /api/catalog-ops/execute
+A **POST** API which takes a SQL statement to create a database/table/view in the local metastore
 (note: this endpoint is only available with Coral Service in local metastore mode).
 
 ### Instructions to use with examples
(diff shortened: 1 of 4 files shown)


wmoustafa:
Thanks @ljfgem! There are a few ways to describe inputs: in the request body, in the query string, and in the URL path. There are a few discussions on how to choose:
https://stackoverflow.com/questions/25385559/rest-api-best-practices-args-in-query-string-vs-in-request-body
https://stackoverflow.com/questions/4024271/rest-api-best-practices-where-to-put-parameters

Also there are a few examples of translation REST services from Microsoft and Google:
https://docs.microsoft.com/en-us/azure/cognitive-services/translator/reference/v3-0-translate
https://cloud.google.com/translate/docs/reference/rest/v3beta1/projects/translateText

Microsoft API uses query strings to model `fromLanguage` and `toLanguage` and request body to model `query` (or `text`). Google API uses the request body to model all three.

I am fine either way, but sounds the request body approach for all is simpler since it uses the same concept for all parameters. For the `query` the request body is particularly useful to avoid any URL length limits and avoid having to encode complex queries, which improves readability.

We should also use the `POST` method. Both the Microsoft and Google APIs do so. 

Let us come up with resource names as well. For example `/v1/translations/translate` and `/v1/tables/create`. We could also add `/v1/tables/drop`. We might model the version as a parameter to (e.g., `api-version=1`). Again, Google and Microsoft APIs made different choices here.

Finally, let us specify the response type/schema and document it. We might want to repeat the parameters (source language, target language and input query) in the response to be self-contained.

cc: @aastha25 @autumnust

aastha25:
I would recommend using the request and response body approach to model all the payloads. It'll keep the URLs short and clean and also allow for easier evolution in the future, if and when we add new arguments. Payloads also help with reusability. So for instance, we would have an enum schema called `Languages`. We could then create a request body for `/translations/v1/translate` and have `fromLanguage` and `toLanguage` arguments be enum values from `Languages`. 

Should we also come up with a top-level resource name for Coral, such as `coralCollections` to identify the project. So our URLs would look like `/coralCollections/v1/translations/translate`, `/coralCollections/v1/tables/create` etc.

autumnust:
> I would recommend using the request and response body approach to model all the payloads. It'll keep the URLs short and clean and also allow for easier evolution in the future, if and when we add new arguments. Payloads also help with reusability. So for instance, we would have an enum schema called `Languages`. We could then create a request body for `/translations/v1/translate` and have `fromLanguage` and `toLanguage` arguments be enum values from `Languages`.
> 
> Should we also come up with a top-level resource name for Coral, such as `coralCollections` to identify the project. So our URLs would look like `/coralCollections/v1/translations/translate`, `/coralCollections/v1/tables/create` etc.

I agree with you in general. But from my previous experience it is not clear if requestBody should be appearing in the GET requests at all. In fact popular framework like spring doesn't support such pattern and there's a post carrying 10+ yrs discussion about this topic: https://stackoverflow.com/questions/978061/http-get-with-request-body

autumnust:
For the POST (`creation`) endpoint I am totally with @aastha25

wmoustafa:
@autumnust, the method for both APIs could be `POST`. Both translation examples mentioned in my comment are `POST`.
@aastha25, I feel we could skip `coralCollections` since `Collections` is a bit generic. Also many best practice examples do not have a  common root for all resources except for the version number: https://www.freecodecamp.org/news/rest-api-best-practices-rest-endpoint-design-examples/.

aastha25:
+1 on both `/tables` and `/translate` being POST. 

Do we have intentions down the roadmap to introduce an endpoint like `https://mysite.com/`? In that case, a root resource might not be necessary. Otherwise, if most of our testing will be via `https://localhost:8080/`, would it look better to introduce a service name?

wmoustafa:
> Do we have intentions down the roadmap to introduce an endpoint like `https://mysite.com/`?

Yes, I think that should be the expectation.

Also, thinking again about `\tables\create`, I think for extensibility, we might just call it `catalog-ops/execute` (or `catalog-operations/execute`) so we can enable many useful statements like `SHOW CREATE TABLE`, `DESCRIBE` etc.

wmoustafa (approved):

PR #271, opened 2022-05-17
Coral-Schema: Preserve underlying table schema for union type fields 

aastha25:
Enhancing the schema generation logic to surface the same ordering of data-types in a union schema as the base table/view. 

The schema generated will now take into account the schema of the underlying field. If a base table has the following field definition - 

     {
        "name" : "tmpField",
        "type" : [ "string", "null" ],
        "doc" : "tmpField documentation",
      }

The avro schema generated for `tmpField` will have the same ordering for type. 

Testing done:
Existing tests + added unit test. 
Tested against views

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
@@ -684,7 +684,7 @@ class SchemaUtilities {
         case MAP:
         case UNION:
         case ARRAY:
-          Schema newFieldSchema = setupNestedNamespace(field.schema(), nestedNamespace);
+          Schema newFieldSchema = setupNestedNamespace(field.schema(), nestedNamespace, field.defaultValue());
           Schema.Field newField =
               new Schema.Field(field.name(), newFieldSchema, field.doc(), field.defaultValue(), field.order());
           appendField(newField, fieldAssembler);
@@ -706,7 +706,8 @@ class SchemaUtilities {
     return fieldAssembler.endRecord();
   }
 
-  private static Schema setupNestedNamespace(@Nonnull Schema schema, @Nonnull String namespace) {
+  private static Schema setupNestedNamespace(@Nonnull Schema schema, @Nonnull String namespace,
+      @Nullable JsonNode defaultValue) {
     Preconditions.checkNotNull(schema);
     Preconditions.checkNotNull(namespace);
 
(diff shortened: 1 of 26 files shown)


ljfgem (commented):
Thanks @aastha25 for this PR!
`setupNestedNamespace` may not be a suitable place for doing such a conversion, given this method should just be used for setting up the namespaces, and it's only called while `strictMode` is `false`.
Could you find another more suitable place to do it? Or maybe create a new schema visitor similar to `ToNullableSchemaVisitor` to do such a conversion.
Besides, I think we may also need to consider the union type like:
```
type: [int, string, null],
default: "aaa" 
```
for such a case, do we need to convert it to be 
```
type: [string, int, null],
default: "aaa" 
```
? 
@rzhang10 @wmoustafa FYI

wmoustafa:
Agreed with @ljfgem. I think if the underlying tables are defined correctly, then they should also have the correct ordering, but it seems the ordering is lost at some point. So I would identify where the ordering is lost first, and try to preserve it there, as opposed to try to reconstruct it after it was lost. It would be great to update the description with where it was being lost if it is the case.

aastha25:
@ljfgem , As per the existing logic, if the union type is - [int, string, null], it wont be considered nullable by AvroSerdeUtils.isNullableType(). That method checks if there's a schemaType null AND that the number of schemas are exactly 2. 
Hence, it would go to the else {} section and append the schemaType in the same order as the underlying table.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -724,41 +725,37 @@ private static Schema setupNestedNamespace(@Nonnull Schema schema, @Nonnull Stri
...
-
-        return enumSchemaWithNestedNamespace;
+        return Schema.createEnum(schema.getName(), schema.getDoc(), namespace, schema.getEnumSymbols());
       case RECORD:
-        Schema recordSchema = setupNestedNamespaceForRecord(schema, namespace);
-        return recordSchema;
+        return setupNestedNamespaceForRecord(schema, namespace);
       case UNION:
         List<Schema> types = new ArrayList<>();
         if (isNullableType(schema)) {
```
wmoustafa, line 740: Does `schema` have the correct ordering? Can we preserve it regardless of what the default value is?

aastha25: schema will have the same ordering as the underlying table/view. It may or may not be correct. 
We can simply preserve it I think. But In that implementation choice, we won't be able to 'fix' it for corrupted tables.

wmoustafa: I would preserve it. It is good to flag corrupted tables.

aastha25: I have preserved the underlying ordering for union type irrespective of the nullability/default value check. 
Now, this has led to another change. When we generate a new field on the fly, for e.g., apply a transformation like count() / max() / unnest() on a nullable column, the schema of this new field appears to be `["long", "null"]` instead of `["null", "long"]` as before. Technically, both representations are correct. But we would be changing the semantics. 

Is it possible that it might lead to regression for some usecases?

wmoustafa: Thanks @aastha25! As long as these are not assigned a default value (which should be the case since they are derived fields), these options should be equivalent.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -725,40 +725,23 @@ private static Schema setupNestedNamespace(@Nonnull Schema schema, @Nonnull Stri
...
-            types.add(typeWithNestNamespace);
-          }
-        }
-        Schema unionSchema = Schema.createUnion(types);
 
-        return unionSchema;
+        for (Schema type : schema.getTypes()) {
+          Schema typeWithNestNamespace = setupNestedNamespace(type, namespace);
+          types.add(typeWithNestNamespace);
+        }
```
wmoustafa, line 743: Nice simplification! @funcheetah do you remember any reason why we did not go it this way in the first place?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
wmoustafa: @funcheetah, a large number of test cases has been impacted by this. Could you take a look?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
funcheetah: I think it is to be consistent with recommended order of optional field in Avro
`(Note that when a [default value](https://avro.apache.org/docs/current/spec.html#schema_record) is specified for a record field whose type is a union, the type of the default value must match the first element of the union. Thus, for unions containing "null", the "null" is usually listed first, since the default value of such unions is typically null.)` as specified in Avro spec: https://avro.apache.org/docs/current/spec.html#Unions

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
funcheetah: Yeah, I agree that if there is no default value set, they should be equivalent. Though a little against Avro optional fields recommendation (null as the first option).

aastha25: So we can either go with the simplified implementation. In this case, we will always generate schema `["long", "null"]` for derived optional columns. It is technically correct but not recommended.

Or we can go with the earlier implementation, where we take into consideration the `defaultValue` and return the recommmend schema type. 

I'm fine with either. If we choose the simplified implementation, this issue might come back to us again? If Avro makes their recommendation mandatory at some point or if some user has an issue due to the inversion?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
wmoustafa: The way I am reading this is that the whole paragraph is qualified by having a default value. If not, I am not sure if they recommendation applies to begin with.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
wmoustafa: @aastha25, for backward compatibility, are we able to maintain the nulls first for derived fields? Which part is responsible for generating `["long", "null"]` (and not the opposite) in the first place?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
funcheetah: Yeah, I agree. However, consider this case, underlying file schema has [a, b], table schema evolves to [a, b, c]. c has to be `[type, null] default: value of type` or `[null, type]` to work. By putting `null` first in optional schema makes it more robust.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
aastha25: When traversing the RelNodePlan, we land upon `visit(LogicalAggregate logicalAggregate)` 

and then this line https://github.com/linkedin/coral/blob/b2ef3ef62de1d6769a6de0183526ea531f2ce16e/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java#L217 
here, you can see we are appending null later.

funcheetah: I do not recall specific reason. Actually, I think it is better to put null as first option. We can use integration tests to catch potential regressions after this change.

wmoustafa: +1. That sounds like the best option here.

aastha25: Tried out this change and it seems to be backward compatible now.

wmoustafa (approved):
Thanks @aastha25 and @funcheetah for the productive discussions.

Issue #272, opened 2022-05-20
Address gson CVE-2022-25647


alexjo2144:
The gson dependency version 2.8.1 is impacted by a CVE https://github.com/advisories/GHSA-4jrv-ppp4-jm57

wmoustafa:
Thanks for filing. We will take a look.

ljfgem:
Closing it given #273 has been merged, thanks @alexjo2144

PR #273, opened 2022-05-20
Update gson to version 2.9.0

ljfgem:
This PR is to solve #272, 2.9.0 is the newest version of gson.

Tested by building it.


wmoustafa (approved):
Thanks @ljfgem!

PR #274, opened 2022-05-23
Clean dependencies and simply build.gradle files

ljfgem:
This PR is to clean all the unused dependencies and simplify build.gradle files.
Tested by building it.

### DIFF
--- a/coral-hive/build.gradle
+++ b/coral-hive/build.gradle
@@ -8,7 +8,6 @@ dependencies {
   // required for hiveSQL parser code
   compile deps.'antlr-runtime'
 
-  testCompile project(path: ':coral-common')
   testCompile(deps.'hive'.'hive-exec-core') {
     exclude group: 'org.apache.avro', module: 'avro-tools'
     // These exclusions are required to prevent duplicate classes since we include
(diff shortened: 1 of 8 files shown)


coral-hive/build.gradle
```diff
@@ -28,12 +27,3 @@ generateGrammarSource {
     '-lib',
     'src/main/antlr/imports'
   ]}
-
-task customFatJar(type: Jar) {
```
aastha25, line 32: does this mean we dont need a fat-jar for coral-hive? 
how did we verify that there are no consumers of this jar?

gradle/dependencies.gradle
```diff
@@ -1,40 +1,28 @@
...
   'testing': '6.4',
-  'coral-calcite': '0.0.3',
   'hive': '1.2.2',
   'hadoop': '2.7.0',
   'pig': '0.15.0',
   'ivy': '2.4.0',
   'spark': '2.4.0',
   'spark3': '3.1.1',
-  'presto-parser': '0.262'
+  'linkedin-calcite-core': '1.21.0.151'
```
aastha25, line 12: nit: could you please add it in alphabetical order?

coral-hive/build.gradle
ljfgem: Thanks! I don't think it's used in Coral itself, and I don't know any consumers of it.
@wmoustafa do you remember the reason why you added this part of code?

wmoustafa: I think it helped testing/debugging on the cluster. I am okay removing it if preferred.

gradle/dependencies.gradle
ljfgem: Thanks for point it out, modified.

aastha25:
Thanks for the PR @ljfgem! 
Changes look good to be checked in.

Issue #249, opened 2022-04-02
Coral-Schema alias output upper-cased schema even if I explicitly alias to lower case


rzhang10:
To reproduce this issue:

add this test in `ViewToAvroSchemaConverterTests` and run it:

```
  @Test
  public void foo() {
    String viewSql = "CREATE VIEW v AS SELECT id as id FROM basecomplex";
    TestUtils.executeCreateViewQuery("default", "v", viewSql);

    ViewToAvroSchemaConverter viewToAvroSchemaConverter = ViewToAvroSchemaConverter.create(hiveMetastoreClient);
    Schema actualSchema = viewToAvroSchemaConverter.toAvroSchema("default", "v", false, false);

    System.out.println(actualSchema);
  }
```

the output will be:
```
{"type":"record","name":"v","namespace":"default.v","fields":[{"name":"Id","type":"int"}]}
```

Which shows the column name is `Id` rather than the explicit lower-cased alias: `id`.

PR #275, opened 2022-05-26
Fix force lowercase is ignored for alias in coral schema conversion

770120041:
This is a similar issue to https://github.com/linkedin/coral/issues/249, in this case, the scheme alias is upper case, even if `forcelowercase` flag is true, the output schema is still upper case.

For the test case below
```
  @Test
  public void testLowercase() {
    String viewSql = "CREATE VIEW v AS SELECT id as ID FROM basecomplex";
    TestUtils.executeCreateViewQuery("default", "v", viewSql);

    ViewToAvroSchemaConverter viewToAvroSchemaConverter = ViewToAvroSchemaConverter.create(hiveMetastoreClient);

    // the last flag is forceLowercase, it is set to true
    Schema actualSchema = viewToAvroSchemaConverter.toAvroSchema("default", "v", false, true); 

    System.out.println(actualSchema);
  }
```
the output will be 
```
{"type":"record","name":"v","namespace":"default.v","fields":[{"name":"Id","type":"int"}]}
```
after this PR, the output will be (the column name is lowercase, as expected)
```
{"type":"record","name":"v","namespace":"default.v","fields":[{"name":"id","type":"int"}]}
```

This PR is removing the `forcelowercase` being used inside [RelToAvroSchemaConverter.java](https://github.com/770120041/coral/commit/425cc2faa552451923dce8b428fb78df9a588635#diff-a2e05f91f268adfd9edbe9dda395197fad075217aaac529dbee5d794a8f6866c), and cast the schema to lowercase in [ViewToAvroSchemaConverter.java](https://github.com/770120041/coral/commit/425cc2faa552451923dce8b428fb78df9a588635#diff-29d8b6a6932537360d80a64e0114c5da3e86b43a3f7f30c3e2b70cfc22dc581b) for both table and views. The test case is added in [ViewToAvroSchemaConverterTests.java](https://github.com/linkedin/coral/compare/master...770120041:lowercase-schema#diff-e1845c8c382ec1780c233a2385d55e8ae52a31699084c02a2a8216ce43b16322)

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
@@ -273,7 +273,8 @@ public class RelToAvroSchemaConverter {
       Schema inputSchema1 = schemaMap.get(logicalUnion.getInput(0));
       Schema inputSchema2 = schemaMap.get(logicalUnion.getInput(1));
 
-      Schema mergedSchema = SchemaUtilities.mergeUnionRecordSchema(inputSchema1, inputSchema2, strictMode);
+      Schema mergedSchema =
+          SchemaUtilities.mergeUnionRecordSchema(inputSchema1, inputSchema2, strictMode, forceLowercase);
 
       schemaMap.put(logicalUnion, mergedSchema);
 
@@ -374,7 +375,7 @@ public class RelToAvroSchemaConverter {
         throw new RuntimeException("Cannot find table " + dbName + "." + tableName + " in Hive metastore");
       }
 
-      Schema tableSchema = SchemaUtilities.getAvroSchemaForTable(baseTable, strictMode, forceLowercase);
+      Schema tableSchema = SchemaUtilities.getAvroSchemaForTable(baseTable, strictMode);
 
       return tableSchema;
     }
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
@@ -192,10 +192,14 @@ public class ViewToAvroSchemaConverter {
 
     if (!tableOrView.getTableType().equals("VIRTUAL_VIEW")) {
       // It's base table, just retrieve the avro schema from Hive metastore
-      return SchemaUtilities.getAvroSchemaForTable(tableOrView, strictMode, forceLowercase);
+      Schema schema = SchemaUtilities.getAvroSchemaForTable(tableOrView, strictMode);
+      return forceLowercase ? ToLowercaseSchemaVisitor.visit(schema) : schema;
     } else {
       RelNode relNode = hiveToRelConverter.convertView(dbName, tableOrViewName);
       Schema schema = relToAvroSchemaConverter.convert(relNode, strictMode, forceLowercase);
+      if (forceLowercase) {
+        schema = ToLowercaseSchemaVisitor.visit(schema);
+      }
       Schema avroSchema = schema;
 
       // In flex mode, we assign a new set of namespace
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
@@ -88,10 +88,9 @@ class SchemaUtilities {
    *
    * @param table
    * @param strictMode if set to true, we do not fall back to Hive schema
-   * @param forceLowercase if set to true, the schema returned is converted to lowercase
    * @return Avro schema for table including partition columns
    */
-  static Schema getAvroSchemaForTable(@Nonnull final Table table, boolean strictMode, boolean forceLowercase) {
+  static Schema getAvroSchemaForTable(@Nonnull final Table table, boolean strictMode) {
     Preconditions.checkNotNull(table);
     Schema resultTableSchema;
     Schema originalTableSchema = SchemaUtilities.getCasePreservedSchemaForTable(table);
@@ -125,8 +124,7 @@ class SchemaUtilities {
       }
     }
 
-    // Return lowercased schema if forceLowercase is set to True
-    return forceLowercase ? ToLowercaseSchemaVisitor.visit(resultTableSchema) : resultTableSchema;
+    return resultTableSchema;
   }
 
   static Schema convertHiveSchemaToAvro(@Nonnull final Table table) {
@@ -451,12 +449,19 @@ class SchemaUtilities {
    * @param rightSchema Right schema to be merged
    * @param strictMode If set to true, namespaces are required to be same.
    *                   If set to false, we don't check namespaces.
-   *
+   * @param forceLowercase If set to true, cast schema to lowercase
    * @return Merged schema if the input schemas can be merged
    */
-  static Schema mergeUnionRecordSchema(@Nonnull Schema leftSchema, @Nonnull Schema rightSchema, boolean strictMode) {
+  static Schema mergeUnionRecordSchema(@Nonnull Schema leftSchema, @Nonnull Schema rightSchema, boolean strictMode,
+      boolean forceLowercase) {
     Preconditions.checkNotNull(leftSchema);
     Preconditions.checkNotNull(rightSchema);
+    // TODO: we should investigate simplify casing transformations
+    if (forceLowercase) {
+      leftSchema = ToLowercaseSchemaVisitor.visit(leftSchema);
+      rightSchema = ToLowercaseSchemaVisitor.visit(rightSchema);
+    }
+
     if (leftSchema.toString(true).equals(rightSchema.toString(true))) {
       return leftSchema;
     }
@@ -558,7 +563,7 @@ class SchemaUtilities {
           }
           break;
         case RECORD:
-          return mergeUnionRecordSchema(leftSchema, rightSchema, strictMode);
+          return mergeUnionRecordSchema(leftSchema, rightSchema, strictMode, false);
         case MAP:
           Schema valueType = getUnionFieldSchema(leftSchema.getValueType(), rightSchema.getValueType(), strictMode);
           return Schema.createMap(valueType);
--- a/coral-schema/src/test/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverterTests.java
+++ b/coral-schema/src/test/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverterTests.java
@@ -976,5 +976,16 @@ public class ViewToAvroSchemaConverterTests {
         TestUtils.loadSchema("testProjectUdfReturnedStruct-expected.avsc"));
   }
 
+  @Test
+  public void testLowercaseSchema() {
+    String viewSql = "CREATE VIEW v AS SELECT id as Id FROM basecomplex";
+    TestUtils.executeCreateViewQuery("default", "v", viewSql);
+
+    ViewToAvroSchemaConverter viewToAvroSchemaConverter = ViewToAvroSchemaConverter.create(hiveMetastoreClient);
+    Schema actualSchema = viewToAvroSchemaConverter.toAvroSchema("default", "v", false, true);
+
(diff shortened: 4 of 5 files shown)


coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
```diff
@@ -272,7 +272,14 @@ public RelNode visit(LogicalUnion logicalUnion) {
       RelNode relNode = super.visit(logicalUnion);
       Schema inputSchema1 = schemaMap.get(logicalUnion.getInput(0));
       Schema inputSchema2 = schemaMap.get(logicalUnion.getInput(1));
-
+      /* forceLowercase is only used for the union of schemas so column name will match
```
ljfgem, line 275: I am not following, looks like this PR converts the final schema to lowercase if `forceLowercase =true` for all the schemas, why do we say it's only used for union of schemas?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
```diff
@@ -272,7 +272,14 @@ public RelNode visit(LogicalUnion logicalUnion) {
...
+       * if forcelowercase is true and user want to cast the schema to lowercase,
```
ljfgem, line 276: `forceLowercase`

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
```diff
@@ -374,7 +381,7 @@ private Schema getTableScanSchema(TableScan tableScan) {
         throw new RuntimeException("Cannot find table " + dbName + "." + tableName + " in Hive metastore");
       }
 
-      Schema tableSchema = SchemaUtilities.getAvroSchemaForTable(baseTable, strictMode, forceLowercase);
+      Schema tableSchema = SchemaUtilities.getAvroSchemaForTable(baseTable, strictMode, false);
```
ljfgem, line 384: I think we can remove `forceLowercase` parameter directly from `SchemaUtilities.getAvroSchemaForTable` given we don't convert the schema to lowercase there anymore with this PR.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
```diff
@@ -91,7 +91,7 @@ public Schema toAvroSchema(String dbName, String tableOrViewName, boolean strict
    *                        the resulting schema. The rule is as follows:
    *                                1. Top level namespace is dbName.tableOrViewName
    *                                2. Nested namespace is parentNamespace.parentFieldName
-   * @param forceLowercase if forceLowercase is set to True, we will return schema with lowercased field names
+   * @param forceLowercase if forceLowercase is set to True, we use lowercase schema for union
```
ljfgem, line 94: Why do we say `we use lowercase schema for union`?
Looks like the newly added unit test is testing a conversion for schema without union.

ljfgem (commented):
Thanks Zhe for this PR!
Please make sure the affected views can be queried on gateway with this PR.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
770120041: This make sense, thanks for finding this. I will try to remove this parameter for this function

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
770120041: Actually I find out the issue in this unit test `testUnionForceLowercase`, please check the comment above. sorry I should include this test in the PR description.

ljfgem: Yeah I saw your description, I don't think it's related to union in particular. 
The issue you mentioned in the description happened because previously, we only used `forceLowercase` while getting the base table's avro schema [here](https://github.com/linkedin/coral/blob/bcfb470a0bbc6ccf8f409363ebe3bc88d2464d45/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java#L129), so that if we set the alias to be non-lowercased (`Id` in your test case), the derived schema is also non-lowercased given we only set lowercase while getting base table's schema, which will be overwritten to be non-lowercased afterwards while visiting the alias.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
ljfgem: Yeah the implementation makes sense, I was just concerned about the comment, I meant `forceLowercase` is not just used for `union`.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
770120041: I see, make sense, I will change this comment to "we will return schema with lowercased field names **(but it won't affect upper cased alias)**", does this works?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
770120041: Sounds good, I will change this comment. By the way you mentioned `Please make sure the affected views can be queried on gateway with this PR.`, is there a document for that?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
ljfgem: I think it will affect upper cased alias because it overwrites the derived schema to be lowercased almost as the last step, I think just keeping the existing comment is fine.

770120041: Sure, sounds good.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
```diff
@@ -196,6 +196,9 @@ private Schema inferAvroSchema(String dbName, String tableOrViewName, boolean st
     } else {
       RelNode relNode = hiveToRelConverter.convertView(dbName, tableOrViewName);
       Schema schema = relToAvroSchemaConverter.convert(relNode, strictMode, forceLowercase);
+      if (forceLowercase) {
```
aastha25, line 199: is it a bit confusing that for views, we force LowerCase at this point, but for tables, that transformation is inside schemaUtilities?
For consistency and readability, can we visit `ToLowercaseSchemaVisitor` for both tables and views in this layer when `forceLowercase` is true? then probably `schemaUtilites` doesn't need to worry about lowercase flag?

770120041: thanks, I have removed the flag `forceLowercase` for schema utility as well, lowercase is handed in  [ViewToAvroSchemaConverter.java](https://github.com/linkedin/coral/pull/275/files/818f6ce19615b6221d875a2474a8c96dc1f34003#diff-29d8b6a6932537360d80a64e0114c5da3e86b43a3f7f30c3e2b70cfc22dc581b) for both tables and views

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
```diff
@@ -272,7 +272,13 @@ public RelNode visit(LogicalUnion logicalUnion) {
       RelNode relNode = super.visit(logicalUnion);
       Schema inputSchema1 = schemaMap.get(logicalUnion.getInput(0));
       Schema inputSchema2 = schemaMap.get(logicalUnion.getInput(1));
-
+      /* if forceLowercase is true, we should cast the schemas to lowercase
+       * before union to make sure column name matches
+       */
+      if (forceLowercase) {
```
aastha25, line 278: I understand that this logic is required. Although I'm not sure if this is the best place for it. Here it seems a custom transformation for merging only. On the surface, it's hard to understand "why".
 
Before reaching LogicalUnion, I think we'll always visit TableScan for the underlying individual tables. Can we visit `ToLowercaseSchemaVisitor` as part of `visit(TableScan tableScan)` instead? Then we won't need this code here

770120041: could you please take a look at the thread I have with Jiefan? there is a test case involved and I find out this is needed

770120041: And I think what you are suggesting here is what we are doing previously. We used to call in `getAvroSchemaForTable`, and this is being used in `getTableScanSchema`, my concern is, when we are calling `ToLowercaseSchemaVisitor`, it is casting the schema fields to lowercase, but it won't cast the final generated schema to lowercase and we have to call `ToLowercaseSchemaVisitor` after `getAvroSchemaForTable` to generated the lower cased schema. 
If we are more concerned about the readability, we can restore everything and just add
```
      if (forceLowercase) {
        schema = ToLowercaseSchemaVisitor.visit(schema);
      }
```
to the converter.

aastha25: Let's not intentionally degrade performance. 
I just noticed that `LogicalUnion` has `LogicalProject` as its child nodes. That probably means that inputSchema1 and inputSchema2 are generated by first visiting `visit(LogicalProject logicalProject)` 

Could you check that if we force lower case in `visit(LogicalProject logicalProject)` , would it be backward compatible and also resolve the issue at hand?

Else, I'm good with the above changes.

770120041: Thanks for the suggestion!

Do you mean the function `RelNode visit(LogicalProject project)` in this class `HiveRelConverter`? 
I didn't find that we have the flag `forcelowercase` available in this class. I don't think we can move `ToLowercaseSchemaVisitor` unless the flag is accessible there.

770120041: I just tested it by adding 
```
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
@@ -214,6 +214,9 @@ public class RelToAvroSchemaConverter {
     public RelNode visit(LogicalProject logicalProject) {
       RelNode relNode = super.visit(logicalProject);
       Schema inputSchema = schemaMap.get(logicalProject.getInput());
+      if (forceLowercase) {
+        ToLowercaseSchemaVisitor.visit(inputSchema);
+      }
```
And this  can't pass the test case we just added. Is this what you are suggesting?

I think this is also caused by the same issue we are trying to solve, which is even if we cast the schema to upper case, it was overwritten to lowercase and we have to call `ToLowercaseSchemaVisitor` out side of the schema converter

rzhang10: Maybe consider factoring the `forceLowercase` flag into the `SchemaUtilities.mergeUnionRecordSchema` utility method to make the method tolerant of field name case sensitivity and always return the lowercase unioned schema.

FYI @aastha25  do you think this is better?

aastha25: Thanks for the suggestion. I think that's a good middle-ground for now. It'll ensure that `SchemaUtilities .mergeUnionRecordSchema` handles casing as per merging requirements.

770120041: thanks Raymond! fixed in the latest patch.

Issue #235, opened 2022-02-28
Extract shared classes between coral-hive and coral-trino to coral-common


KevinGe00:
After [adding to trino2rel](https://github.com/linkedin/coral/pull/224), there are a few classes that both coral-hive and coral-trino use that can now be put in coral-common instead such as `HiveConvertletTable`. Class name changes will also be needed, i.e `HiveConvertletTable` to just `ConvertletTable`. 

 These changes will be also helpful when adding future support for other SqlLanguage2rel conversions.

PR #276, opened 2022-05-31
Move `DaliOperatorTable` to Coral-Common and rename it

ljfgem:
This PR is for #235.
`DaliOperatorTable` registers legal operators for Calcite to validate and convert the input SqlNode.
Given we convert other types of operators (like Trino operators) to Hive (Coral) operators in SqlNode phase, we use `DaliOperatorTable` for both Coral-Hive hive2rel, Coral-Trino trino2rel and all the future inputs, therefore, it makes more sense to move `DaliOperatorTable` to Coral-Common. 
To move it, we need to create an abstract class `FunctionResolver` to be used in `CoralOperatorTable`.
cc: @wmoustafa 

Tested by building and i-itest.

### DIFF
--- a/coral-common/src/main/java/com/linkedin/coral/common/functions/CoralOperatorTable.java
+++ b/coral-common/src/main/java/com/linkedin/coral/common/functions/CoralOperatorTable.java
@@ -3,7 +3,7 @@
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
-package com.linkedin.coral.hive.hive2rel;
+package com.linkedin.coral.common.functions;
 
 import java.util.Collection;
 import java.util.List;
@@ -19,26 +19,23 @@ import org.apache.calcite.sql.SqlSyntax;
 import org.apache.calcite.sql.validate.SqlNameMatcher;
 import org.apache.calcite.util.Util;
 
-import com.linkedin.coral.common.functions.Function;
-import com.linkedin.coral.hive.hive2rel.functions.HiveFunctionResolver;
-
 
 /**
- * Class to resolve Dali function names in SQL definition based on
+ * Class to resolve Coral operator names in SQL definition based on
  * the mapping stored in table parameters in the metastore.
  */
-public class DaliOperatorTable implements SqlOperatorTable {
+public class CoralOperatorTable implements SqlOperatorTable {
   // TODO: support injection framework to inject same function resolver here and ParseTreeBuilder.
   // For now, we create another instance since the function registry is simple.
-  private final HiveFunctionResolver funcResolver;
+  private final FunctionResolver funcResolver;
 
-  public DaliOperatorTable(HiveFunctionResolver funcResolver) {
+  public CoralOperatorTable(FunctionResolver funcResolver) {
     this.funcResolver = funcResolver;
   }
 
   /**
-   * Resolves functions names to corresponding Calcite UDF. HiveFunctionResolver ensures that
-   * {@code sqlIdentifier} has function name or corresponding class name for Dali functions. All function registry
+   * Resolves operator names to corresponding Calcite operator. FunctionResolver ensures that
+   * {@code sqlIdentifier} has operator name or corresponding class name for Coral operator. All registry
    * lookups performed by this class are case-insensitive.
    *
    * Calcite invokes this function multiple times during analysis phase to validate SqlCall operators. This is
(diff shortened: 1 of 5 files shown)


coral-common/src/main/java/com/linkedin/coral/common/functions/CoralOperatorTable.java
```diff
@@ -3,7 +3,7 @@
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
-package com.linkedin.coral.hive.hive2rel;
+package com.linkedin.coral.common.functions;
```
wmoustafa, line 6: I would skip `functions` since this is a core class.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveFunctionResolver.java
```diff
@@ -43,14 +44,13 @@
 /**
  * Class to resolve hive function names in SQL to Function.
  */
-public class HiveFunctionResolver {
+public class HiveFunctionResolver extends FunctionResolver {
```
wmoustafa, line 47: Why could not not this be moved to coral-common?

ljfgem: `HiveFunctionResolver` is only used for Coral-Hive and contains many Coral-Hive specific methods like Generic UDF dynamic registration. Therefore, I don't think we need this class for all the other xxx2rel modules, we need to have `xxxFunctionResolver` extending `FunctionResolver` for each xxx2rel module instead.

coral-trino/src/main/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverter.java
```diff
@@ -47,8 +48,12 @@
 public class TrinoToRelConverter extends ToRelConverter {
   private final ParseTreeBuilder parseTreeBuilder = new ParseTreeBuilder();
   private final ParserVisitorContext parserVisitorContext = new ParserVisitorContext();
-  private final HiveFunctionResolver functionResolver =
-      new HiveFunctionResolver(new StaticHiveFunctionRegistry(), new ConcurrentHashMap<>());
+  private final FunctionResolver functionResolver = new FunctionResolver(new StaticHiveFunctionRegistry()) {
```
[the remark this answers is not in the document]
ljfgem: Thanks for suggestion, added.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveFunctionResolver.java
aastha25: @wmoustafa do you have a preference between common function resolver vs language-specific function resolvers?

ljfgem: Not sure IIUC, I don't think we can use a common function resolver in `coral-common` to solve all the `language2rel` functions, given difference input languages may need difference methods in their function resolvers, for example, hive2rel needs dynamic Hive generic UDF registration while trino2rel doesn't need.

wmoustafa: I think aside from the dynamic resolution, rest should be language independent and go to coral-common?

> And in practice, given `HiveFunctionResolver` uses many classes in Coral-Hive like `VersionedSqlUserDefinedFunction`, `HiveRLikeOperator` and `HiveGenericUDFReturnTypeInference`, we need to move all of them for `HiveFunctionResolver`, which is not suitable in my opinion.

I think those are ultimately going to be Coral IR specific classes, and not Hive?

ljfgem: @wmoustafa Rests are `resolveUnaryOperator` and `resolveBinaryOperator` which are only used in Hive's `ParseTreeBuilder`, not used in `trino2rel` because `trino2rel` has its own `ParseTreeBuilder`. So I would prefer to just keep `resolve` in `FunctionResolver` for now, we may add more common methods if there is in the future.

PR #279, opened 2022-06-07
Coral-Hive: Make field access case-insensitively

ljfgem:
This PR creates `HiveRexBuilder` which makes field access case-insensitively, because if the base table `t` contains non-lowercase struct field like `s struct(A:string)`, the schema of the view `v` based on the base table would become `s struct(a:string)`, translation for SQL `SELECT * FROM v WHERE v.s.A='xxx'` will fail with the following exception if caseSensitive=true, given Calcite would convert `v.s.A` to `v.s.a` to be aligned with the schema of view `v` during the validation phase:
```
java.lang.AssertionError: Type 'RecordType(VARCHAR(2147483647) A)' has no field 'a'
```
Setting caseSensitive=false would not cause regression because Calcite doesn't allow two struct fields which only differ in casing like `struct(a:string,A:string)`, check `org.apache.calcite.sql.validate.DelegatingScope.fullyQualify` for more info.

Tests:
1. Tested on the affected production views, which could be translated well with this patch
2. Integration test for Trino, Spark and Avro, no regression.

### DIFF
--- a/coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/HiveToRelConverterTest.java
+++ b/coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/HiveToRelConverterTest.java
@@ -633,6 +633,17 @@ public class HiveToRelConverterTest {
     assertEquals(generated, expected);
   }
 
+  @Test
+  public void testCaseInsensitiveFieldAccess() {
+    final String expected = "LogicalProject(s=[$0])\n" + "  LogicalFilter(condition=[=($0.name, 'Jason')])\n"
+        + "    LogicalProject(s=[$0])\n"
+        + "      LogicalTableScan(table=[[hive, test, table_with_uppercase_struct_field]])\n";
+    final String sql =
+        "SELECT * FROM test.view_with_uppercase_struct_field WHERE view_with_uppercase_struct_field.s.Name='Jason'";
+    String generated = relToString(sql);
+    assertEquals(generated, expected);
+  }
+
   private String relToString(String sql) {
     return RelOptUtil.toString(converter.convertSql(sql));
   }
--- a/coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/HiveToRelConverterTest.java
+++ b/coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/HiveToRelConverterTest.java
@@ -633,17 +633,6 @@ public class HiveToRelConverterTest {
     assertEquals(generated, expected);
   }
 
-  @Test
-  public void testCaseInsensitiveFieldAccess() {
-    final String expected = "LogicalProject(s=[$0])\n" + "  LogicalFilter(condition=[=($0.name, 'Jason')])\n"
-        + "    LogicalProject(s=[$0])\n"
-        + "      LogicalTableScan(table=[[hive, test, table_with_uppercase_struct_field]])\n";
-    final String sql =
-        "SELECT * FROM test.view_with_uppercase_struct_field WHERE view_with_uppercase_struct_field.s.Name='Jason'";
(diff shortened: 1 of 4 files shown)


coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/HiveToRelConverterTest.java
```diff
@@ -633,6 +633,17 @@ public void testNameSakeColumnNamesShouldGetUniqueIdentifiers() {
     assertEquals(generated, expected);
   }
 
+  @Test
+  public void testCaseInsensitiveFieldAccess() {
```
rzhang10, line 637: I'm not sure if I'm doing anything wrong, but it seems this test passes even without your coral-hive main source code changes. Does this imply this unit test is not well designed?

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/HiveToRelConverterTest.java
```diff
@@ -633,6 +633,17 @@ public void testNameSakeColumnNamesShouldGetUniqueIdentifiers() {
...
+    final String expected = "LogicalProject(s=[$0])\n" + "  LogicalFilter(condition=[=($0.name, 'Jason')])\n"
```
aastha25: This relNode has `condition=[=($0.name, 'Jason')]`. 

what will the hive / spark / trino SQL look like?
Will the generated SQL would have where small-casing for fieldnames in where clause, like - `WHERE view_with_uppercase_struct_field.s.name='Jason'`? or will the fieldName will upper-case?

ljfgem: I just removed this unit test as we can't reproduce this issue by unit test because Coral uses Hive 1.2.2 while LI uses 1.1 in production.
But according to my test on the cluster, it would preserve the upper-case field names in the result SQL, so that it's aligned with the base table's schema and the translation could succeed.

wmoustafa:
We need to think through this and ensure it does not have backward compatibility implications for coral-schema.

ljfgem:
> We need to think through this and ensure it does not have backward compatibility implications for coral-schema.

@wmoustafa @rzhang10 
This overwritten method is called [here](https://github.com/linkedin/linkedin-calcite/blob/c13ba44c6974574d07a47766e32e695e5c5e042f/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java#L3731), which is set to be `true` by default, because `Calcite would convert v.s.A to v.s.a to be aligned with the schema of view v during the validation phase`, so Calcite assumes the name already fully-qualified. However, the base table's schema (`s struct(A:string)`) is not the same as the intermediate view `v`'s schema (`s struct(a:string)`) while creating the view, which causes the issue. Therefore, the root cause is that our Hive preserves the field name casing info when we create a table, but loses the field name casing info when we create a view based on that table. I don't know why it happens at present, need to do more investigations.

As for your concern, this fix wouldn't cause backward compatibility issue for Coral-Schema because it wouldn't convert the fields to lowercase, it would just help to loose the case-sensitive restriction while retrieving a field from a struct, so the retrieved field name's casing info would still be preserved. And enhanced i-test for Coral-Schema also passed.
And as I mentioned in the comment, setting `caseSensitive=false` would not cause regression because Calcite doesn't allow base table with two struct fields which only differ in casing like `struct(a:string,A:string)` according to my experiment, check [here](https://github.com/linkedin/linkedin-calcite/blob/c13ba44c6974574d07a47766e32e695e5c5e042f/core/src/main/java/org/apache/calcite/sql/validate/DelegatingScope.java#L496) for more info.

wmoustafa (approved):

PR #280, opened 2022-06-14
Support schema converting of union for fixed and bytes

770120041:
From [Avro Spec](https://avro.apache.org/docs/current/spec.html),  bytes is a sequence of 8-bit unsigned bytes and fixed is a containing a fixed-size number of bytes, so we can support the union of Bytes and Fixed column and the return type will be bytes.
 
Without this PR, we will see the following error message during union of fixed and bytes column

`Exception in thread "main" java.lang.RuntimeException: Found two incompatible schemas for LogicalUnion operator. Left schema is: "bytes". Right schema is: fixed`

PR #282, opened 2022-06-15
Support schema enum union with different number of symbols and enum u…

770120041:
This is similar to this [PR](https://github.com/linkedin/coral/pull/280). We should support the union of `enum` and `string` type, and the union of  enums with different number of symbols but the symbols of one enum is a strict subset of the other.

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
@@ -554,19 +554,11 @@ class SchemaUtilities {
           }
           break;
         case ENUM:
-          if (leftSchema.getEnumSymbols().size() == rightSchema.getEnumSymbols().size()) {
-            return leftSchema;
-          }
-          // Check if the symbols of one Enum is a strict subset of the other
-          final ImmutableSet<String> leftSchemaSymbols = ImmutableSet.copyOf(leftSchema.getEnumSymbols());
-          final ImmutableSet<String> rightSchemaSymbols = ImmutableSet.copyOf(rightSchema.getEnumSymbols());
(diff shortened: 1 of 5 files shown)


coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -556,6 +557,15 @@ private static Schema getUnionFieldSchema(@Nonnull Schema leftSchema, @Nonnull S
...
+          // Check if the symbols of one Enum is a strict subset of the other
+          final ImmutableSet<String> leftSchemaSymbols = ImmutableSet.copyOf(leftSchema.getEnumSymbols());
+          final ImmutableSet<String> rightSchemaSymbols = ImmutableSet.copyOf(rightSchema.getEnumSymbols());
+          if (leftSchemaSymbols.containsAll(rightSchemaSymbols)) {
+            return leftSchema;
+          }
+          if (rightSchemaSymbols.containsAll(leftSchemaSymbols)) {
+            return rightSchema;
+          }
           break;
```
ljfgem, line 569: Could you verify if we can union enum types like `[a, b]` and `[c, d]` in Hive?
If that is allowed, I think we may need to return the union of these enums (`[a, b, c, d]` for the above example)
If only subset-union is allowed, I think we need to remove 
```
          if (leftSchema.getEnumSymbols().size() == rightSchema.getEnumSymbols().size()) {
            return leftSchema;
          }
```

770120041: Sounds good, will do. Thanks for pointing this out.

ljfgem (approved):
LGTM, thanks @770120041
Let's merge after the view can be queried by Spark3 on the gateway.

PR #281, opened 2022-06-14
Register some UDFs

ljfgem:
`ArtifactsResolver` takes very long to resolve these UDFs dynamically due to the issues like:
```
	impossible to acquire lock for com.fasterxml.jackson.core#jackson-annotations;2.7.8

22/06/14 19:44:09 WARN functions.ArtifactsResolver: Unable to fetch dependencies: [unresolved dependency: com.fasterxml.jackson.core#jackson-annotations;2.7.8: configuration not found in com.fasterxml.jackson.core#jackson-annotations;2.7.8: 'master'. It was required from com.fasterxml.jackson.core#jackson-databind;2.7.8 compile]
```
This PR registers these UDFs manually to mitigate this issue.

Tested on the affected production views, which can be translated well in normal time with this PR.

PR #288, opened 2022-07-13
Register some UDFs

ljfgem:
Similar to #281 
Tested on the affected production views, which can be translated well in normal time with this PR


aastha25:
Thanks for the PR! LGTM

PR #292, opened 2022-07-25
Correct nullability information for certain functions to unblock a vi…

rzhang10:
…ew for spark3

Previously a production view failed in translation in spark3, due to the fact that the view defines a `named_struct` logic: 
```
NAMED_STRUCT(
    'objectUrn', CONCAT('xxx', company_id),
    ...
```

which the `company_id` itself is from another projection:

```
    mapLookup(
      trackinginfo,
      'xxx',
      'xxx''
    ) AS company_id
```

Coral derives the `objectUrn` to be non-nullable, which is incorrect against spark3's derivation, and it makes the view fails in spark3.

I found out that the `named_struct` relevant code logic in `HiveNamedStructFunction` , and it uses the return type inference logics defined in `StaticHiveFunctionRegistry` to derive type information (including nullability). So this patch will fix the nullability to be correct.


ljfgem (approved):
Let's merge after verifying the view can be queried by Spark3 with this PR.

rzhang10:
Tested on gateway the view can be corrected accessed.

PR #295, opened 2022-08-04
Coral-Spark: Register generic_project UDF

ljfgem:
We need to register `generic_project` UDF if it's used in Spark SQL, otherwise Spark will throw the following exception because it can't recognize `generic_project`:
```
org.apache.spark.sql.AnalysisException: Undefined function: 'generic_project'. This function is neither a registered temporary function nor a permanent function registered in the database 'default'.; line 1 pos 7
```

Tested on gateway with Spark3, the affected view can be queried with this PR.


aastha25:
Thanks for fixing this issue, @ljfgem 
LGTM!

PR #297, opened 2022-08-16
[Sample PR|WIP] Spark View Catalog with HMS and Coral

ljfgem:
This PR is almost based on the [upstream Spark view catalog PR](https://github.com/apache/spark/pull/28147) with some modifications on that PR:
1. Added `tableCatalog` property in [View](https://github.com/apache/spark/pull/28147/files#diff-15aa34af438db0a59b1c7ca13bfda59768ac66957d4233de41e4f8977323aad4) interface so that the view `coral.db.view` in `CoralSparkViewCatalog` can be based on the tables in other catalogs like default Hive catalog. Without such a property, `CoralSparkViewCatalog` must implement `TableCatalog` interface and implement all the table operation methods because `coral.db.view` will be assumed to be based on `coral.db.tbl` table in the same catalog. With this property, we will use the `tableCatalog` while [qualifying base table identifiers](https://github.com/apache/spark/pull/28147/files#diff-ed19f376a63eba52eea59ca71f3355d4495fad4fad4db9a3324aade0d4986a47R375) during the view resolution.

The unit tests worked well.


wmoustafa:
Could you make the design doc public and link it from here?

ljfgem:
> Could you make the design doc public and link it from here?

@wmoustafa Added in the description, but due to the org restriction, I can only share a copied doc (without previous comments) from my personal google account.

PR #301, opened 2022-09-08
Add return type inference for isb.GetProfileSections

rzhang10:
Fix the return type inference of `isb.GetProfileSections` by explicitly define its `SqlReturnTypeInference` logic in the `StaticHiveFunctionRegistry`.

Verified the fix works by validating the output avro schema of such a UDF transformation result in the correct output avro schema.


ljfgem (approved):
LGTM, please ensure that the view can be queried on gateway with this pr.

PR #303, opened 2022-09-08
Fix a bug in basicSqlTypeToAvroType

rzhang10:
The Calcite `ANY` type shouldn't correspond to any Avro primitive types, if we get a `ANY` type in this function, it certainly means we are doing something wrong, instead of returning a Avro `BYTES` type.


ljfgem (approved):
LGTM, i-test passed.
Thanks @rzhang10 for this PR!

PR #304, opened 2022-09-09
Coral-Spark: Use wildcard for GenericProject and GenericUDFCoalesceStruct dependencies

ljfgem:
This PR is to use the latest published `GenericUDFCoalesceStruct` with a fix. Wildcard matching helps to avoid version bumping everytime we publish a new version of these UDFs.

Tested on the gateway, the latest published UDF was picked up and the affected view can be queried well with this PR.


aastha25:
Thanks for the PR @ljfgem . LGTM!

PR #306, opened 2022-09-15
UDF: Add additional UDF mapping to handle Trino naming discrepancies

autumnust:
Add a hive udf registration, as well as the mapping rule to a Trino udf to fill in the gap of default rules of UDF naming conversion, which is from camel use to snake case.


ljfgem (approved):
LGTM, thanks @autumnust

PR #308, opened 2022-09-19
Register some UDFs

ljfgem:
Similar to https://github.com/linkedin/coral/pull/281
Tested on the affected production view, its translation can be finished in normal time with this PR


aastha25:
thanks for the PR @ljfgem! LGTM

PR #309, opened 2022-09-21
Modify return types of 2 UDFs

ljfgem:
The return types of these 2 UDFs were not correct, this PR is to modify them.

Tested on the affected production views.


aastha25:
Thanks for fixing the return types @ljfgem! 
LGTM.

PR #310, opened 2022-09-23
Coral-spark: Remove explicit casting for nullability difference

rzhang10:
Previously, for a CASE expression: `CASE WHEN TRUE THEN NULL ELSE split(b, ' ') END` , Coral spark sql will generate `CASE WHEN TRUE IS CAST(NULL AS ARRAY<STRING>) THEN NULL ELSE CAST(split(b, ' ')) AS ARRAY<STRING>) END`.

In spark3, it treats target SQL type names in the CAST as all nullable, it means, for the above nested type of `ARRAY<STRING>` , its nullable at all levels: `ARRAY<STRING:nullable>:nullable`. This is actually different from the Calcite RelNode nullability: `ARRAY<STRING:non-nullable>:nullable`, which is entailed in RelNode and Coral-schema, but lost when the RelNode is converted to SparkSql (because spark sql type names can't contain nullable info).

With this fix, we basically remove those CAST wrappers when generating the sparkSQL, and the resulting SQL will always  be compatible with both Spark sql analyzer and coral-schema with regard to nullability.

The affected view is tested with the patch in spark3 and it succeeds.

### DIFF
--- a/coral-spark/src/main/java/com/linkedin/coral/spark/IRRelToSparkRelTransformer.java
+++ b/coral-spark/src/main/java/com/linkedin/coral/spark/IRRelToSparkRelTransformer.java
@@ -34,11 +34,7 @@ import org.apache.calcite.rel.logical.LogicalValues;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.rel.type.RelRecordType;
-import org.apache.calcite.rex.RexBuilder;
-import org.apache.calcite.rex.RexCall;
-import org.apache.calcite.rex.RexLiteral;
-import org.apache.calcite.rex.RexNode;
-import org.apache.calcite.rex.RexShuttle;
+import org.apache.calcite.rex.*;
 import org.apache.calcite.sql.SqlIdentifier;
 import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlOperator;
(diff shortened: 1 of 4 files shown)


coral-spark/src/main/java/com/linkedin/coral/spark/IRRelToSparkRelTransformer.java
```diff
@@ -368,6 +364,29 @@ else if (call.getOperands().size() == 2) {
       return Optional.empty();
     }
 
+    // Calcite entails the nullability of an expression by casting it to the correct nullable type.
+    // However, for complex types like ARRAY<STRING NOT NULL> (element non-nullable, but top-level nullable),
+    // the translated SQL will still be `CAST(XXX) AS ARRAY<STRING>`. Since Spark treats a Cast target type
```
ljfgem, line 369: nit: `CAST(XXX AS ARRAY<STRING>)`

coral-spark/src/main/java/com/linkedin/coral/spark/IRRelToSparkRelTransformer.java
```diff
@@ -368,6 +364,29 @@ else if (call.getOperands().size() == 2) {
...
+    // as always nullable (both inner and outer), it will treat the SQL type as ARRAY<STRING>, this deviates
+    // from the nullability represented in RelNode/Coral-Schema.
```
ljfgem, line 371: Could you clarify the nullability of RelNode/Coral-Schema here? Maybe point to the unit test you added in Coral-Schema.

ljfgem (commented):
Thanks @rzhang10 for this PR!
Left some nits, and kicked off the regression test.

coral-spark/src/main/java/com/linkedin/coral/spark/IRRelToSparkRelTransformer.java
rzhang10: added

coral-spark/src/main/java/com/linkedin/coral/spark/IRRelToSparkRelTransformer.java
rzhang10: fixed

ljfgem (approved):
LGTM, and the i-test result looks good, thanks @rzhang10 !

PR #311, opened 2022-09-27
Coral-Trino: Handle casting from varbinary/binary to varchar

ljfgem:
Trino doesn't allow casting varbinary/binary to varchar, we need to use the built-in function `from_utf8` to replace the cast, i.e. `CAST(binary AS VARCHAR) -> from_utf8(binary)`

Tests:
1. Unit test
2. Tested on the affected production view
3. i-test, no regression after verification

### DIFF
--- a/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
+++ b/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
@@ -406,10 +406,6 @@ public class Calcite2TrinoUDFConverter {
       return Optional.of(rexBuilder.makeCast(typeFactory.createSqlType(TIMESTAMP, 3), rexBuilder.makeCall(op)));
     }
 
-    // Hive allows for casting of TIMESTAMP to DECIMAL, which converts it to unix time if the decimal format is valid
-    // Example: "SELECT cast(current_timestamp() AS decimal(10,0));" -> 1633112585
-    // Trino does not allow for such conversion, but we can achieve the same behavior by first calling "to_unixtime"
-    // on the TIMESTAMP and then casting it to DECIMAL after.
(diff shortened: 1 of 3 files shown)


coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java
```diff
@@ -419,13 +415,25 @@ private Optional<RexNode> visitCast(RexCall call) {
...
       if (call.getType().getSqlTypeName() == DECIMAL && leftOperand.getType().getSqlTypeName() == TIMESTAMP) {
         SqlOperator trinoToUnixTime = createUDF("to_unixtime", explicit(DOUBLE));
         SqlOperator trinoWithTimeZone = createUDF("with_timezone", explicit(TIMESTAMP /* should be WITH TIME ZONE */));
         return Optional.of(rexBuilder.makeCast(call.getType(), rexBuilder.makeCall(trinoToUnixTime,
             rexBuilder.makeCall(trinoWithTimeZone, leftOperand, rexBuilder.makeLiteral("UTC")))));
       }
 
+      // Trino doesn't allow casting varbinary/binary to varchar, we need to use the built-in function `from_utf8`
+      // to replace the cast, i.e. CAST(binary AS VARCHAR) -> from_utf8(binary)
+      if (call.getType().getSqlTypeName() == VARCHAR && (leftOperand.getType().getSqlTypeName() == VARBINARY
```
rzhang10, line 431: Can this also be `CHAR` instead of `VARCHAR`?

ljfgem: Thanks for reminding, added this case. Since there is no special function returning `char` type, I still replace it with `from_utf8`.

rzhang10 (approved):

PR #313, opened 2022-10-06
[Reference PR] Coral-Schema: Upgrade Avro to 1.10 and modify makeNullable logic

ljfgem:
This PR is based on #307 by @i-ony with the following modifications:
1. Combine the commits and rebase for simplicity
2. Remove unnecessary changes to [build.gradle of other modules](https://github.com/linkedin/coral/pull/307/files#diff-a44c8ceda67698445c043896bee418bce68a0d9a23eb62f2373233c5307201ce) and [dependencies.gradle](https://github.com/linkedin/coral/pull/307/files#diff-02b139e755ab0e27e0b0a6f9845843ef189116955cc340c49a4022587dbb2b52), we just need to add `avroCompatHelper` dependency and force Coral-Schema to use Avro 1.10 by the following config in Coral-Schema's build.gradle:
```
configurations.all {
  resolutionStrategy {
    force 'org.apache.avro:avro:1.10.2'
  }
}
```
3. Modify `makeNullable` method to take a boolean parameter `nullAsSecond`, if it's true, will construct schemas like `[long, null]`. Previously, `makeNullable` can only construct `[null, long]` with `null` as the first option.
4. Modify `MergeHiveSchemaWithAvro.struct`, `MergeHiveSchemaWithAvro.list`, `MergeHiveSchemaWithAvro.map` and `MergeHiveSchemaWithAvro.primitive` methods which call `SchemaUtilities.makeNullable` to construct nullable schema with the same order as `partner` by the following code:
```
SchemaUtilities.makeNullable(result, SchemaUtilities.isNullSecond(partner));
```
so that if `partner` is `[long, null]`, the return result is also `[long, null]` rather than `[null, long]`

5. Similar to 4, in `SchemaUtilities.getUnionFieldSchema` method, if leftSchema and rightSchema are both nullable, we construct final schema with the same order as left schema (it's just one of the options, we need further discussion to determine the behavior):
```
return makeNullable(getUnionFieldSchema(makeNonNullable(leftSchema), makeNonNullable(rightSchema), strictMode),
          isNullSecond(leftSchema));
```

Tested on the affected production view, which can be translated with change 4 and 5.

FYI @wmoustafa @aastha25

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
@@ -16,6 +16,7 @@ import javax.annotation.Nullable;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableSet;
+import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
 
 import org.apache.avro.Schema;
 import org.apache.avro.SchemaBuilder;
@@ -36,7 +37,6 @@ import org.apache.hadoop.hive.serde2.typeinfo.StructTypeInfo;
 import org.apache.hadoop.hive.serde2.typeinfo.TypeInfo;
 import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoFactory;
 import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoUtils;
-import org.codehaus.jackson.JsonNode;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -164,10 +164,10 @@ class SchemaUtilities {
       if (!Strings.isNullOrEmpty(schemaStr)) {
         schemaStr = schemaStr.replaceAll("\n", "\\\\n");
         // Given schemas stored in `dali.row.schema` are all non-nullable, we need to convert them to be nullable to be compatible with Spark
-        schema = ToNullableSchemaVisitor.visit(new Schema.Parser().parse(schemaStr));
+        schema = ToNullableSchemaVisitor.visit(AvroCompatibilityHelper.parse(schemaStr));
       }
     } else {
-      schema = new Schema.Parser().parse(schemaStr);
+      schema = AvroCompatibilityHelper.parse(schemaStr);
     }
 
     if (schema != null) {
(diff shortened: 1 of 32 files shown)


coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -540,7 +553,8 @@ private static Schema getUnionFieldSchema(@Nonnull Schema leftSchema, @Nonnull S
       return makeNullable(leftSchema);
     }
     if (isNullableType(leftSchema) || isNullableType(rightSchema)) {
-      return makeNullable(getUnionFieldSchema(makeNonNullable(leftSchema), makeNonNullable(rightSchema), strictMode));
+      return makeNullable(getUnionFieldSchema(makeNonNullable(leftSchema), makeNonNullable(rightSchema), strictMode),
+          isNullSecond(leftSchema));
```
aastha25, line 557: could you add a comment here suggesting that if leftSchema and rightSchema are nullable union types with different ordering, we are choosing the ordering from the leftSchema. It could serve useful during future refactoring, if and when needed.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -618,6 +632,10 @@ static Schema makeNonNullable(Schema schema) {
   }
 
   static Schema makeNullable(Schema schema) {
```
aastha25, line 634: why not remove this method and change all the calling references to `makeNullable(schema, false)` ? I feel that makes the code easier to follow.

coral-schema/src/test/resources/base-complex-nullable-with-defaults.avsc
```diff
@@ -7,7 +7,7 @@
     "type" : "int"
   }, {
     "name" : "Int_Field",
-    "type" : [ "null", "int" ],
+    "type" : [ "int", "null" ],
```
aastha25, line 10: why do we still need these changes? Arent we surfacing whatever is the schema in the base avro/orc table?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -618,6 +632,10 @@ static Schema makeNonNullable(Schema schema) {
...
+    return makeNullable(schema, false);
+  }
+
+  static Schema makeNullable(Schema schema, boolean nullAsSecond) {
```
aastha25, line 638: could you also please add a unit test in `ViewToAvroSchemaConverterTests.java` where we generate the schema from an orc table? Something similar to `testSchemaWithNullableFieldsAndDefaults` but from an orc table.
All the test tables we currently create are avro as per this method - https://github.com/linkedin/coral/blob/master/coral-schema/src/test/java/com/linkedin/coral/schema/avro/TestUtils.java#L166

aastha25 (commented):
Thanks for the PR @ljfgem ! 
Reviewed changes related to ordering in nullable union types and left comments.

ljfgem:
@aastha25 thanks for your review, this PR is a reference PR for next i-test, so I just focused on the logic and didn't spend much time on the coding style/unit tests, which should be handled by the formal Avro upgrade PR. If we are fine with the logic, we can proceed with the i-test.

aastha25:
Thanks for the clarification @ljfgem . The logic looks good to me and we can use it for reference at this stage. We can come back to reviews before actually checking in. 
could you just help answer this comment https://github.com/linkedin/coral/pull/313#discussion_r989323304

coral-schema/src/test/resources/base-complex-nullable-with-defaults.avsc
ljfgem: We need to change the order here because this is not a valid schema in Avro 1.10, and the table with this wrong schema can't be created.

ljfgem:
Closing it given the formal PR #314 has been opened.

PR #314, opened 2022-10-11
Coral-Schema: Make Coral-Schema compatible with both Avro 1.7 and 1.10

ljfgem:
This PR is to make Coral-Schema compatible with both Avro 1.7 and Avro 1.10.

Main changes:
1. Use `AvroCompatibilityHelper` to create schema field, get default value, parse schema and add props to field/schema.
2. Modify `makeNullable` method to take a boolean parameter `nullAsSecond`, if it's true, will construct schemas like `[long, null]`. Previously, `makeNullable` can only construct `[null, long]` with `null` as the first option.
3. Modify `MergeHiveSchemaWithAvro.struct`, `MergeHiveSchemaWithAvro.list`, `MergeHiveSchemaWithAvro.map` and `MergeHiveSchemaWithAvro.primitive` methods which call `SchemaUtilities.makeNullable` to construct nullable schema with the same order as `partner`.
4. Similar to 3, in `SchemaUtilities.getUnionFieldSchema` method, if `leftSchema` and `rightSchema` are both nullable, we construct final schema with the same order as `leftSchema`.
5. Move `reorderOptionIfRequired` from `MergeHiveSchemaWithAvro` to `SchemaUtilities` and use it in `SchemaUtilities.mergeUnionRecordSchema` to reorder the union option if necessary.
6. Disable unit tests for join given Avro 1.10 doesn't allow two schema fields to share the same name, we may need to investigate if there is another way to handle it. It doesn't affect our production views for now.
7. Modify expected unit test results by adding `default : null` or removing `default : null` to let the UTs pass, it shouldn't affect production use because default value will be dropped while converting Avro schema to Spark schema. Reasons of adding/removing `default : null` are:
a. Reason of adding: Now we create schema field by `AvroCompatibilityHelper.createSchemaField`, if the input default value is `null`, `default : null` will be explicitly added in the final schema. Previously with Avro 1.7, we created schema field by `new Schema.field`, if the input default value is `null`, `default : null` would not appear in the final schema. So we need to add `default : null` to some result fields
b. Reason of removing: If the created field's default value is `null`, while calling [appendField](https://github.com/linkedin/coral/blob/d0a47081ced98c6aaa9db62eff5160189bad88b0/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java#L183) afterwards, the default value will be [removed](https://github.com/linkedin/coral/blob/d0a47081ced98c6aaa9db62eff5160189bad88b0/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java#L194). So all the UTs calling [appendField](https://github.com/linkedin/coral/blob/d0a47081ced98c6aaa9db62eff5160189bad88b0/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java#L183) would drop `default : null` in the final result

Tests:
1. Unit tests
2. i-test on production views, regressions are caused by some issues of `AvroCompatibilityHelper` and faulty schemas of base tables, all the other issues have been fixed in this PR.


coral-common/src/main/java/com/linkedin/coral/common/HiveTable.java
```diff
@@ -186,7 +185,7 @@ private Deserializer getDeserializer() {
   private Deserializer getDeserializerFromMetaStore() {
     try {
       return MetaStoreUtils.getDeserializer(new Configuration(false), hiveTable, false);
-    } catch (MetaException e) {
+    } catch (Throwable e) { // Catch Throwable here because it may throw Exception or Error
```
aastha25, line 188: why do we catch and rethrow an RTE here at all when the exception will anyway be caught here https://github.com/linkedin/coral/blob/master/coral-common/src/main/java/com/linkedin/coral/common/HiveTable.java#L166 and logged? can we remove the try catch block?

coral-schema/src/test/java/com/linkedin/coral/schema/avro/TestUtils.java
```diff
@@ -184,10 +184,8 @@ private static void executeCreateTableWithPartitionQuery(String dbName, String t
   private static void executeCreateTableWithPartitionFieldSchemaQuery(String dbName, String tableName,
       String fieldSchema) {
     executeQuery("DROP TABLE IF EXISTS " + dbName + "." + tableName);
-    executeQuery("CREATE EXTERNAL TABLE " + tableName + " (" + fieldSchema + ") "
-        + "PARTITIONED BY (datepartition string) " + "ROW FORMAT SERDE 'org.apache.hadoop.hive.serde2.avro.AvroSerDe' "
-        + "STORED AS " + "INPUTFORMAT 'org.apache.hadoop.hive.ql.io.avro.AvroContainerInputFormat' "
-        + "OUTPUTFORMAT 'org.apache.hadoop.hive.ql.io.avro.AvroContainerOutputFormat'");
+    executeQuery(
```
aastha25, line 187: what does this simplification achieve? will the resulting schema format be avro / orc?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
```diff
@@ -232,6 +232,7 @@ public RelNode visit(LogicalProject logicalProject) {
 
     @Override
     public RelNode visit(LogicalJoin logicalJoin) {
+      // TODO: Modify this method to avoid that 2 schema fields share the same name in Avro 1.10, and enable the tests for join
```
aastha25, line 235: 1. does this scenario mean using the same fieldNames, but referring to different columns, with different data types, across 2 schemas?
2. is this a non-blocking to-do? I see that some UTs have been disabled. Will there be a regression in production views if we don't implement it in this PR?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeHiveSchemaWithAvro.java
```diff
@@ -58,7 +58,10 @@ public Schema struct(StructTypeInfo struct, Schema partner, List<Schema.Field> f
     } else {
       result = SchemaUtilities.copyRecord(SchemaUtilities.extractIfOption(partner), fieldResults);
     }
-    return shouldResultBeOptional ? SchemaUtilities.makeNullable(result) : result;
+    // While calling `makeNullable`, we should respect the option order of `partner`
```
aastha25, line 61: this comment is not added is every method call for `makeNullable(..)`.
Could you either add it everywhere or just add it in the class description and the method's javadoc?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -975,4 +1001,18 @@ static StructTypeInfo structTypeInfoFromCols(List<FieldSchema> cols) {
...
   }
+
+  /**
+   * Reorders an option schema so that the type of the provided default value is the first type in the option schema
+   *
+   * e.g. If the schema is [null, int] and the default value is 1, the returned schema is [int, null]
+   * If the schema is not an option schema or if there is no default value, schema is returned as-is
+   */
+  static Schema reorderOptionIfRequired(Schema schema, Object defaultValue) {
+    if (isNullableType(schema) && defaultValue != null && schema.getTypes().get(0).getType() == Schema.Type.NULL) {
```
aastha25, line 1012: should we also check if `defaultValue.toString != null`?

coral-schema/src/test/resources/base-complex-nullable-with-defaults.avsc
```diff
@@ -7,7 +7,7 @@
     "type" : "int"
   }, {
     "name" : "Int_Field",
-    "type" : [ "null", "int" ],
+    "type" : [ "int", "null" ],
```
aastha25, line 10: can you add a UT where we union `basecomplexnullablewithdefaults` with itself and some of the underlying fields have incorrect ordering for nullable union types? (basically, retaining the origin basecomplexnullablewithdefaults and union with self)?

(1) I see that during the merge of LHS and RHS schemas, if the datatype ordering in the nullable union is not correct, we fix it. it'll be good to have a UT validating it. 
(2) this seems like a design choice where if we are union-ing 2 incorrect base tables, we fix the datatype ordering of the merged schema on the fly (by calling `reorderOptionIfRequired`), but we don't fix the schema if we are generating schema of a single incorrect base table.  Should we be consistent in our approach?

aastha25 (commented):
Thanks for the comprehensive PR @ljfgem ! 
Could you also elaborate on point 7 of your PR description?  As per avro 1.10, when is `default:null` expected? That would be helpful in reviewing the test file changes

coral-common/src/main/java/com/linkedin/coral/common/HiveTable.java
ljfgem: https://github.com/linkedin/coral/blob/master/coral-common/src/main/java/com/linkedin/coral/common/HiveTable.java#L166 also only catches `Exception`, not `Error`.
Compared to throw a Throwable and modify [here](https://github.com/linkedin/coral/blob/master/coral-common/src/main/java/com/linkedin/coral/common/HiveTable.java#L166) to be `Throwable`, I prefer the current approach.

coral-schema/src/test/java/com/linkedin/coral/schema/avro/TestUtils.java
ljfgem: Removing it because with Avro 1.10, we can't create a table just with [this kind of schema](https://github.com/linkedin/coral/blob/master/coral-schema/src/test/resources/base-complex-fieldschema)
the created table format is `inputFormat = "org.apache.hadoop.mapred.TextInputFormat",
outputFormat = "org.apache.hadoop.hive.ql.io.HiveIgnoreKeyTextOutputFormat"`

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
ljfgem: 1. Yes, and the data types may also be same
2. It's a non-blocking to-do, there is no regression on all the production views.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
ljfgem: If `defaultValue` is not `null`, I think `defaultValue.toString != null` is always true?

coral-schema/src/test/resources/base-complex-nullable-with-defaults.avsc
ljfgem: (1) We can't create a base table with incorrect ordering for nullable union types in UT given Avro 1.10 doesn't allow it
(2) We call `reorderOptionIfRequired` for `unionFieldSchema` because it is generated by our code, we need to ensure its correctness. But for a single base table, it should be user's responsibility to ensure the correctness of its schema.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeHiveSchemaWithAvro.java
ljfgem: Thanks for reminding, added.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
aastha25: (obj = null) is not a safe null check. https://stackoverflow.com/questions/2369226/object-null-or-null-object mentions that if obj datatype is Boolean, this check could be problematic. 
Do we use bool/boolean? 

Its probably safe to replace with `Objects.toString(object, "isNull") != "isNull"` as mentioned https://stackoverflow.com/a/13684257 . What do you think?

coral-schema/src/test/java/com/linkedin/coral/schema/avro/TestUtils.java
aastha25: (1) so all schema generation will go in the else case here https://github.com/linkedin/coral/blob/master/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java#L118  ?

(2) can you add this input format as an inline comment? it doesn't seem intuitive. 
(3) Also, how can we create and test avro format tables? or is that code path only tested on production views via integ tests?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
ljfgem: I still don't think it's necessary for this method, I have tried to reorder the following schema:
```
{
    "name" : "Boolean_field",
    "type" : ["boolean", "null"],
    "default": false
}
```
the default value's type is Boolean only if it's `true` or `false`, which can't be null. The current `reorderOptionIfRequired` method handles the above test case well.

coral-schema/src/test/java/com/linkedin/coral/schema/avro/TestUtils.java
ljfgem: (1) (3) No, the other tables are created with Avro format by these 2 methods: [1](https://github.com/linkedin/coral/blob/d0a47081ced98c6aaa9db62eff5160189bad88b0/coral-schema/src/test/java/com/linkedin/coral/schema/avro/TestUtils.java#L166), [2](https://github.com/linkedin/coral/blob/d0a47081ced98c6aaa9db62eff5160189bad88b0/coral-schema/src/test/java/com/linkedin/coral/schema/avro/TestUtils.java#L175)
(2) Will add

aastha25:
thanks for the changes! 
LGTM!

ljfgem:
Thanks @aastha25 for your review!
@wmoustafa: do you want to take another look given it's a big change?

PR #323, opened 2022-10-28
Introduce select * optimization in CoralIR to trino translations

aastha25:
When an input query is projecting all the columns from an underlying un-evolved table, the translated trino SQL does not need to list all the columns explicitly in the projected list. The translated SQL can instead use "*". 

Ex: 
for a table: `foo<a int, b string, c boolean>`
input hive SQL: `select a,b,c from foo`
old translated trino SQL: `select a,b,c from foo`
new translated SQL: `select * from foo`


coral-trino/src/test/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverterTest.java
```diff
@@ -235,7 +234,7 @@ public void testSupport(String trinoSql, String expectedRelString, String expect
     RelToTrinoConverter relToTrinoConverter = new RelToTrinoConverter();
     // Convert rel node back to Sql
     String expandedSql = relToTrinoConverter.convert(relNode);
-    assertEquals(expandedSql, expectedSql);
+    assertEquals(expectedSql, expandedSql);
```
ljfgem, line 237: Could you also modify the `assertEquals` before it? (line 232)

ljfgem (commented):
Thanks @aastha25, it looks good to me. And I can't come up with a scenario/view query which could fail due to this change.

coral-trino/src/test/java/com/linkedin/coral/trino/trino2rel/TrinoToRelConverterTest.java
aastha25: done. thanks for pointing out.

ljfgem (approved):
LGTM, @wmoustafa do you want to take another look? Or could you come up with a case that would cause regression?

PR #324, opened 2022-10-28
[WIP] Supporting CTAS queries for Hive to Spark query translations

nimesh1601:
The approach and details are mentioned in this Doc for reference: [Supporting CTAS queries for Hive to Spark](https://docs.google.com/document/d/1Q8UIpUjDoJ6tM7JYLsKdIyBjOcZ_GbF-wnssKgANAwk/edit#heading=h.1b49jd39z58x)

Unit tests are WIP


coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilder.java
```diff
@@ -293,34 +270,34 @@ LATERAL VIEW json_tuple(json, p1, p2) jt AS a, b
 
       // '$["jsonKey"]'
       SqlCall jsonPath = SqlStdOperatorTable.CONCAT.createCall(ZERO,
-          SqlStdOperatorTable.CONCAT.createCall(ZERO, SqlLiteral.createCharString("$[\"", ZERO), jsonKey),
-          SqlLiteral.createCharString("\"]", ZERO));
+              SqlStdOperatorTable.CONCAT.createCall(ZERO, SqlLiteral.createCharString("$[\"", ZERO), jsonKey),
```
wmoustafa: Could you check why we have all these indentation changes?

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
```diff
@@ -115,6 +118,43 @@ private static String constructSparkSQL(RelNode sparkRelNode) {
...
+   * @param sqlNode CoralNode which will be translated to SparkSQL.
+   * @param convertor Functional Interface to convert SqlNode to appropriate RelNode
+   *
+   *
+   * @return [[CoralSparkInfo]]
+   */
+  public static CoralSpark create(SqlNode sqlNode, Function<SqlNode, RelNode> convertor){
+    SparkRelInfo sparkRelInfo;
+    //apply RelNode transformations for sqlNode eligible for transformation.
+    if(sqlNode instanceof SqlCreate) {
```
wmoustafa, line 141: How about adding an interface `SqlCommand`, and have `SqlCreateTable` extend `SqlCreate` and implement `SqlCommand`? `SqlCommand` defines `getSelectQueries()` and `setSelectQueries()`. Those are just list versions of what you have now in `SqlCreateTable`.

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
```diff
@@ -115,6 +118,43 @@ private static String constructSparkSQL(RelNode sparkRelNode) {
...
+   *  2) Base tables
+   *  3) Spark UDF information objects, ie. List of {@link SparkUDFInfo}
+   *
+   * @param sqlNode CoralNode which will be translated to SparkSQL.
+   * @param convertor Functional Interface to convert SqlNode to appropriate RelNode
+   *
+   *
+   * @return [[CoralSparkInfo]]
+   */
+  public static CoralSpark create(SqlNode sqlNode, Function<SqlNode, RelNode> convertor){
```
wmoustafa, line 138: Note that we are assuming this is the `Coral IR SqlNode`. Assuming that `Coral IR SqlNode` has array indexes start from 1 (see `standardizeRel()`, I am pretty sure the array index in this tree will start from 0. For consistency, somehow we need a `SELECT` node here that:
1-  is a`SqlNode`
2- has array indexes starting from 1.
I think this implies that the caller to this method must have taken the query already, converted the `SELECT SqlNode` to `RelNode` then back to `SELECT SqlNode`. The former `SELECT` will have array indexes start from 0, and the latter will have them start from 1. What remains after the caller has done this step is that the latter `Coral IR SqlNode` is simply converted to Spark SQL string (using the Spark dialect and after applying the`SparkSqlRewriter` (on the latter `SqlNode`), without having to go through `RelNode` here.
The caller can call the `RelNode` based `create` on the obtained `RelNode` in the caller. It is best to code this logic in a test case. @aastha25 @ljfgem FYI.

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
nimesh1601: yeah, this sounds good

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
nimesh1601: Yes, currently SqlNode given as input to this function has `0-based indexing. Now as you have mentioned that we need a `SELECT`node here should have array indexes starting from 1, so it means all `RelNode` based transformations, would have been done by the caller of this function. It means the caller should be able to execute the below transformations on the `SELECT `node before calling this new create function.
```
    if(sqlNode instanceof SqlCommand) {
      SqlNode selectNode = ((SqlCreateTable) sqlNode).getSelectQuery();
      sparkRelInfo = IRRelToSparkRelTransformer.transform(convertor.apply(selectNode));
      selectNode = new CoralRelToSqlNodeConverter().convert(sparkRelInfo.getSparkRelNode());
      ((SqlCreateTable) sqlNode).setSelectQuery(selectNode);
    } else {
      sparkRelInfo = IRRelToSparkRelTransformer.transform(convertor.apply(sqlNode));
      sqlNode = new CoralRelToSqlNodeConverter().convert(sparkRelInfo.getSparkRelNode());
    }
```
 Should we create a separate function in `CoralSpark` to handle this `SqlNode`->`RelNode`->`SqlNode` transformation? 
 I don't think, we can call `RelNode`-based create as currently, it will convert `RelNode` to `SQL` directly.

coral-common/src/main/java/com/linkedin/coral/common/calcite/sql/SqlCreateTable.java
```diff
@@ -0,0 +1,91 @@
...
+            }
+            writer.endList(frame);
+        }
+        if(tableSerializer != null){
+            writer.keyword("ROW FORMAT SERDE");
+            tableSerializer.unparse(writer, 0, 0);
+            writer.newlineAndIndent();
+        }
+        if(tableRowFormat != null){
+            writer.keyword("ROW FORMAT DELIMITED FIELDS TERMINATED BY");
```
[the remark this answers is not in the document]
nimesh1601: Are you suggesting having a conditional check if entered values for the row format are correct or something else?

coral-common/src/main/java/com/linkedin/coral/common/calcite/sql/SqlCreateTable.java
```diff
@@ -0,0 +1,91 @@
...
+            tableRowFormat.unparse(writer, 0, 0);
+            writer.newlineAndIndent();
+        }
+        if(tableFileFormat != null){
+            if(tableFileFormat.size() == 1){
+                writer.keyword("STORED AS");
+                tableFileFormat.get(0).unparse(writer, 0, 0);
+                writer.newlineAndIndent();
+            } else {
```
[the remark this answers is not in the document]
nimesh1601: Similarly, are you suggesting to have a check, if the file format is valid, like ORC, Parquet, etc? I think unparsing logic shouldn't worry about these.

coral-common/src/main/java/com/linkedin/coral/common/calcite/sql/SqlCreateTable.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.List;
+import java.util.Objects;
+
+public class SqlCreateTable extends SqlCreate {
+    private final SqlIdentifier name;
+    private final @Nullable SqlNodeList columnList;
+    private @Nullable SqlNode query;
+    private final @Nullable SqlNode tableSerializer;
+    private final @Nullable SqlNodeList tableFileFormat;
+    private final @Nullable SqlCharStringLiteral tableRowFormat;
```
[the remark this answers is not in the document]
nimesh1601: Are you suggesting using a separate custom SqlNode for RowFormat which will contain information like serDe, fieldDelim, colDelim, etc? I think having a custom SqlNode will later help in supporting other syntaxes of row formats in a better way.

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
aastha25, line 138: If the argument `sqlNode` is of type `SqlCreate`, the query's semantics are not getting validated. For example: 
for a table `foo` with multiple columns, if there are input queries like `create table t1 as select * as col0 from foo` or `create table t1 (a int) as select * from foo`, I'm not sure if Coral will throw an exception right now. 

The validation work can be incremental, but maybe we can setup a class like `ddlSqlValidator` in coral-common / coral-hive modules which can validate the different types of ddl statements and invoke it here. 

We should validate that the rowtype is compatible for `SqlCreateTable`.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilder.java
```diff
@@ -608,6 +611,96 @@ protected SqlNode visitSelect(ASTNode node, ParseContext ctx) {
...
+          ctOptions.tableFileFormat = (SqlNodeList) visitFileFormatGeneric(ast, ctx);
+          break;
+        case HiveParser.TOK_TABLEROWFORMAT:
+          ctOptions.tableRowFormat = (SqlCharStringLiteral) visitTableRowFormat(ast, ctx);
+          break;
+        default:
+          break;
+      }
+    }
+    return new SqlCreateTable(ZERO, false, ctOptions.ifNotExists != null ? ctOptions.ifNotExists : false, ctOptions.name,
```
aastha25, line 648: Can we introduce a utility class like calcite does here https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql/ddl/SqlDdlNodes.java to manage all ddl sqlNode creation?

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
aastha25: I feel that `getSelectQueries()` and `setSelectQueries()` will be used for `CREATE TABLE/VIEW....AS`. If there's no other use case, can we rename `SqlCommand` -> `SqlCreateCommand`?

coral-common/src/main/java/com/linkedin/coral/common/calcite/sql/SqlCreateTable.java
```diff
@@ -0,0 +1,91 @@
...
+        this.tableRowFormat = tableRowFormat;
+    }
+
+    @SuppressWarnings("nullness")
+    @Override public List<SqlNode> getOperandList() {
+        return ImmutableNullableList.of(name, columnList, query);
+    }
+
+    @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
+        writer.keyword("CREATE");
```
aastha25, line 40: nit: line 40 & 41 can be combined

aastha25 (commented):
Thanks @nimesh1601  for the PR!

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
nimesh1601: Queries like `create table t1 as select * as col0 from foo` will fail query parsing, but yes, we would need validation for cases like `create table t1 (a int) as select * from foo`. Do we have anything similar for select queries?

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
nimesh1601: Not sure, but I think commands like `INSERT OVERWRITE TABLE tmp  select * FROM foo`, might also leverage this

coral-common/src/main/java/com/linkedin/coral/common/calcite/sql/SqlCreateTable.java
nimesh1601: better to keep it separate, later we will have support for temporary and external table, which will be prefix to `TABLE` keyword

coral-common/src/main/java/com/linkedin/coral/common/calcite/sql/SqlCreateTable.java
```diff
@@ -0,0 +1,91 @@
...
+        this.tableFileFormat = tableFileFormat;
+        this.tableRowFormat = tableRowFormat;
+    }
+
+    @SuppressWarnings("nullness")
+    @Override public List<SqlNode> getOperandList() {
+        return ImmutableNullableList.of(name, columnList, query);
+    }
+
+    @Override public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {
```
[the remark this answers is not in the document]
nimesh1601: Added in ParseTreeBuilderTest.java

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilder.java
```diff
@@ -1059,4 +1152,14 @@ Optional<Table> getHiveTable() {
...
+
+  class CreateTableOptions {
+    SqlIdentifier name;
+    SqlNodeList columnList;
+    SqlNode query;
+    Boolean ifNotExists;
+    SqlNode tableSerializer;
+    SqlNodeList tableFileFormat;
+    SqlCharStringLiteral tableRowFormat;
+  }
```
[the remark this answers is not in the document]
nimesh1601: yes, we can, but then we would need to expose appropriate setter methods

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
aastha25: for select queries, we use the [hiveSqlValidator](https://github.com/linkedin/coral/blob/master/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveSqlValidator.java#L25) to validate the CoralSqlNode and then generate the relNode representation. This is done [here](https://github.com/linkedin/coral/blob/master/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveSqlToRelConverter.java#L59).

if create DDL (and other DDL) are not flowing through this translation path, we should validate them separately.

aastha25: in the translation path:
`Language1 (hiveSQL) --> CoralSqlNode (LHS) --> CoralRelNode --> CoralSqlNode (RHS) --> Language 2 (spark SQL) `

The `coral-spark` class is responsible for generating sparkSQL from CoralIR. It currently accepts the `CoralRelNode` and applies the RHS side of transformations to generate a spark-compatible SQL. In your current implementation of `create(..)`, you are passing CoralSqlNode (LHS) and using `convertor.apply(selectNode)` to generate `CoralRelNode`. Could you please refactor to apply the LHS side of transformations outside this class and then invoke existing method `create(RelNode irRelNode){..}` to generate the updated `selectNode`?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilder.java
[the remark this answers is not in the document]
aastha25: Can you add it to the existing ParseContext? ParseContext is being used for the same purpose for select queries.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilder.java
```diff
@@ -1059,4 +1152,14 @@ Optional<Table> getHiveTable() {
       return hiveTable;
     }
   }
-}
+
+  class CreateTableOptions {
```
ljfgem, line 1156: It may be static.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilder.java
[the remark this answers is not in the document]
ljfgem: I think it's fine to keep this class, adding these fields to `ParseContext` may make it confusing, i.e. `selects` and `columnList`.

ljfgem (commented):
Thanks @nimesh1601 for this PR!
Please run `./gradlew spotlessApply` to fix the format issues.

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
```diff
@@ -70,6 +70,22 @@ public static CoralSpark create(RelNode irRelNode) {
...
 
+  /**
+   * Users use this function to get CoralSqlNode from CoralRelNode
+   * This should be used when user need to get CoralSqlNode from CoralRelNode by applying
+   * spark specific transformations on CoralRelNode
+   * with Coral-schema output schema
+   *
+   * @return [[SqlNode]]
+   */
+  public static SqlNode getCoralSqlNode(RelNode irRelNode) {
```
aastha25, line 81: I'm unclear why this is needed. 

CoralSpark API should take Coral intermediate representation as input and return Spark SQL / SparkSQLNode. 

You could create a 
```
 String convert(SqlCommand sqlCommand, RelNode coralRelNode) {
// generate CoralSqlNode (RHS) for coralRelNode
// set the updated CoralSqlNode (RHS) in sqlCommand
// send this (combined) SqlNode to CoralSqlNodeToSparkSqlNodeConverter and then SparkSqlRewriter
}
```

coral-service/src/main/java/com/linkedin/coral/coralservice/utils/TranslationUtils.java
```diff
@@ -29,8 +31,21 @@ public static String translateHiveToTrino(String query) {
   }
 
   public static String translateHiveToSpark(String query) {
-    RelNode relNode = new HiveToRelConverter(hiveMetastoreClient).convertSql(query);
-    CoralSpark coralSpark = CoralSpark.create(relNode);
-    return coralSpark.getSparkSql();
+    HiveToRelConverter hiveToRelConverter = new HiveToRelConverter(hiveMetastoreClient);
```
aastha25, line 34: (Pl read the comment on CoralSpark.getCoralSqlNode(..) before this comment) 

I think this class could use a bit of refactoring.
For `SqlNode sqlNode = hiveToRelConverter.toSqlNode(query)` : hiveToRelConverter could return a coralSqlNode. Currently, on the LHS side, hiveSqlNode is equivalent CoralSqlNode (LHS).

Inside the if condition, you could extract CoralSelectSqlNode from coralSqlNode, and hand it over first to  `hiveToRelConverter` to generate CoralRelNode. And then hand over sqlCommand + CoralRelNode to CoralSpark.

aastha25 (commented):
Thanks for the changes!! Left some comments about the structuring, will continue reviewing. 

Also, could you please add javadoc for all public classes and methods :)

coral-common/src/main/java/com/linkedin/coral/common/calcite/sql/SqlCreateTable.java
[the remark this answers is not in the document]
wmoustafa: I am saying ROW FORMAT implies a few things (e.g., either SERDE or DELIMITED). Somehow you take `tableRowFormat` to reflect a lot of things together: `ROW FORMAT DELIMITED FIELDS TERMINATED BY`.

coral-common/src/main/java/com/linkedin/coral/common/calcite/sql/SqlCreateTable.java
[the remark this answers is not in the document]
wmoustafa: I am saying `ROW FORMAT` concept encompasses both `SERDE` and `DELIMITED` concepts/keywords. The code structure should reflect the syntax structure.

coral-common/src/main/java/com/linkedin/coral/common/calcite/sql/SqlCreateTable.java
```diff
@@ -0,0 +1,91 @@
...
+
+import com.linkedin.coral.javax.annotation.Nullable;
+import org.apache.calcite.sql.*;
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.util.ImmutableNullableList;
+
+import java.util.List;
+import java.util.Objects;
+
+public class SqlCreateTable extends SqlCreate {
```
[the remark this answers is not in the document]
wmoustafa: Let us get back to this, but it might depend on the discussion above.

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
wmoustafa: @nimesh1601 could you summarize the offline discussion/agreement here?

wmoustafa: @aastha25 I do not think we should create two parallel paths. Can we unify them using common APIs?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilder.java
wmoustafa: @aastha25 Why does DDL SqlNode creation warrant a separate path? Why cannot it be done through generic SqlNode generation?

coral-common/src/main/java/com/linkedin/coral/common/calcite/DdlSqlValidator.java
```diff
@@ -0,0 +1,14 @@
...
+ */
+package com.linkedin.coral.common.calcite;
+
+import org.apache.calcite.sql.SqlNode;
+
+
+public interface DdlSqlValidator {
+
+  void validate(SqlNode ddlSqlNode);
+}
```
wmoustafa, line 14: Who is supposed to call this (external calling code or internally inside some Coral step)? How can we unify it with the existing pipeline for validation? @aastha25 please let me know your thoughts.

coral-common/src/main/java/com/linkedin/coral/common/calcite/sql/util/SqlDdlNodes.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class SqlDdlNodes {
+
+  /** Creates a CREATE TABLE. */
+  public static SqlCreateTable createTable(SqlParserPos pos, boolean replace, boolean ifNotExists, SqlIdentifier name,
+      SqlNodeList columnList, SqlNode query, SqlNode tableSerializer, SqlNodeList tableFileFormat,
+      SqlCharStringLiteral tableRowFormat) {
+    return new SqlCreateTable(pos, replace, ifNotExists, name, columnList, query, tableSerializer, tableFileFormat,
+        tableRowFormat);
+  }
```
wmoustafa, line 25: Sounds redundant? @aastha25

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
```diff
@@ -70,6 +70,22 @@ public static CoralSpark create(RelNode irRelNode) {
...
   }
 
+  /**
+   * Users use this function to get CoralSqlNode from CoralRelNode
+   * This should be used when user need to get CoralSqlNode from CoralRelNode by applying
+   * spark specific transformations on CoralRelNode
+   * with Coral-schema output schema
+   *
+   * @return [[SqlNode]]
+   */
```
wmoustafa, line 80: Java doc needs to be more elaborate, and reflect the relationship with existing methods. @aastha25 could help with that.

PR #315, opened 2022-10-19
Integrate CoralRelToSqlNodeConverter in CoralRelNode to trino SQL translation path. 

aastha25:
Update the CoralRelNode -> Trino SQL translation path to use `CoralRelToSqlNodeConverter`.

Added `CoralSqlNodeToTrinoSqlNodeConverter` to enable conversion from CoralSqlNode to TrinoSqlNode. 

Testing:
(1) Updated / Added tests in HiveToTrinoConverterTest to validate hive -> trino SQL translations via new code path

Pending:
Migrate RelToTrinoConverterTest to use hive SQL as input.
Update expected trino SQLs in TrinoToRelConverterTest

PR #329, opened 2022-11-03
Coral-trino: Migrate Alias appending transformer from RelNode To SqlNode layer

aastha25:
Currently, `RelToTrinoConverter` hosts multiple transformations required to convert a CoralRelNode to a Trino-compatible SQL statement. This PR intends to relocate one such transformation from RelNode layer to SqlNode layer. 

Currently, AS operator is inserted while accessing nested fields. For ex: `SELECT structField.intField FROM foo` is converted to `SELECT structField.intField AS intField FROM foo`. Previously, this transformation was done at the Rel->SqlNode layer. This PR moves the change to SqlNode rewritten layer. 

Testing:
Integ tested with all the production data sources. No regression was observed. 
Also refactored `RelToTrinoConverter` to utilize `HiveToRelConverter`  

**Note:** This PR captures a more minor change from a larger effort here https://github.com/linkedin/coral/pull/315  to test backward compatibility in smaller chunks. The difference in translations due to PR#315 is large. The changes have been split over multiple PRs to reduce the effort of testing and be able to monitor the impact on production views more closely.


coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/IntermediateTrinoSqlNodeRewriter.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.sql.util.SqlShuttle;
+import org.apache.calcite.sql.validate.SqlValidatorUtil;
+
+import static org.apache.calcite.rel.rel2sql.SqlImplementor.*;
+import static org.apache.calcite.sql.parser.SqlParserPos.*;
+
+
+/**
+ * IntermediateTrinoSqlNodeRewriter rewrites the TrinoSqlNode representation. It transforms the sqlCalls
```
wmoustafa, line 30: `SqlNodes`

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/IntermediateTrinoSqlNodeRewriter.java
```diff
@@ -0,0 +1,104 @@
...
+  public IntermediateTrinoSqlNodeRewriter() {
+  }
+
+  @Override
+  public SqlNode visit(final SqlCall call) {
+    SqlCall transformedSqlCall = getTransformedSqlCall(call);
+    return super.visit(transformedSqlCall);
+  }
+
+  public static SqlCall getTransformedSqlCall(SqlCall sqlCall) {
```
yiqiangin, line 51: declare as `private`?

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/IntermediateTrinoSqlNodeRewriter.java
```diff
@@ -0,0 +1,104 @@
...
+
+import static org.apache.calcite.rel.rel2sql.SqlImplementor.*;
+import static org.apache.calcite.sql.parser.SqlParserPos.*;
+
+
+/**
+ * IntermediateTrinoSqlNodeRewriter rewrites the TrinoSqlNode representation. It transforms the sqlCalls
+ * in the TrinoSqlNode to be compatible with Trino engine in a backward compatible manner.
+ *
+ * This is achieved by visiting the TrinoSqlNode AST in a pre-order traversal manner and
```
wmoustafa, line 33: Please do not use `TrinoSqlNode` since it is not a class or well defined entity. In fact, if anything it should be the final output, not the input.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/IntermediateTrinoSqlNodeRewriter.java
```diff
@@ -0,0 +1,104 @@
...
+ * transforming each SqlNode (SqlCall), wherever required.
```
wmoustafa, line 34: Stick to one in the entire comment (SqlNode or SqlCall).

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/IntermediateTrinoSqlNodeRewriter.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  /**
+   * Input SqlCall to this method will access nested fields as is, such as SELECT a.b, a.c.
+   * The below transformations appends an AS operator on nested fields and updated the SqlCall to: SELECT a.b AS b, a.c AS c
+   * @param sqlCall input SqlCall
+   * @return transformed SqlCall
+   */
+  private static SqlCall getTransformedSqlSelectSqlCall(SqlCall sqlCall) {
+    if (((SqlSelect) sqlCall).getSelectList() != null && ((SqlSelect) sqlCall).getSelectList().size() != 0) {
```
yiqiangin, line 85: Does Coral use Apache Commons Collection? If yes, I think it is better to call `Collections.isEmpty()` to check if the list is null or empty().

yiqiangin (approved):
LGTM

ljfgem:
@aastha25 could you elaborate the reason why we need to separate this PR from #315 ? Would be great if we don't need to introduce a temp class `IntermediateTrinoSqlNodeRewriter` which will be modified soon.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/IntermediateTrinoSqlNodeRewriter.java
```diff
@@ -0,0 +1,104 @@
...
+      final List<SqlNode> modifiedSelectList = new ArrayList<>();
+
+      for (SqlNode selectNode : ((SqlSelect) sqlCall).getSelectList().getList()) {
+        final String name = SqlValidatorUtil.getAlias(selectNode, -1);
+        final boolean nestedFieldAccess =
+            selectNode instanceof SqlIdentifier && ((SqlIdentifier) selectNode).names.size() > 1;
+
+        // always add "AS" when accessing nested fields.
+        // In parent class "as" is skipped for "select a.b as b", here we will keep the "a.b as b"
```
wmoustafa, line 94: `Always add`
Convert all SQL to uppercase.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/RelToTrinoConverter.java
```diff
@@ -98,6 +99,7 @@ public Result visit(Window window) {
     return null;
   }
 
+  // override is required to prevent select nodes such as SELECT CAST(NULL AS NULL)
```
wmoustafa, line 102: Comment is not clear. Please contrast the behavior before and after override clearly.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/IntermediateTrinoSqlNodeRewriter.java
aastha25: Tried changing to:
`if (CollectionUtils.isNotEmpty(((SqlSelect) sqlCall).getSelectList().getList()))`

it throws NPE since `getSelectList()` might be null. `getSelectList()` returns type `SqlNodeList` which would also need to be checked.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/TrinoSqlDialect.java
```diff
@@ -71,10 +71,6 @@ public void unparseIdentifier(SqlWriter writer, SqlIdentifier identifier) {
     writer.endList(frame);
   }
 
-  public boolean requireCastOnString() {
-    return true;
-  }
-
```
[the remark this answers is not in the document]
aastha25: this public method was an override for https://github.com/linkedin/linkedin-calcite/blob/li-1.21.0/core/src/main/java/org/apache/calcite/sql/SqlDialect.java#L612 without the annotation. 

It should be safe to delete.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/IntermediateTrinoSqlNodeRewriter.java
aastha25: done

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/IntermediateTrinoSqlNodeRewriter.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.calcite.sql.util.SqlShuttle;
+import org.apache.calcite.sql.validate.SqlValidatorUtil;
+
+import static org.apache.calcite.rel.rel2sql.SqlImplementor.*;
+import static org.apache.calcite.sql.parser.SqlParserPos.*;
+
+
+/**
+ * IntermediateTrinoSqlNodeRewriter rewrites the TrinoSqlNode representation. It transforms the sqlCalls
+ * in the TrinoSqlNode to be compatible with Trino engine in a backward compatible manner.
```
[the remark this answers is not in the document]
aastha25: updated the documentation. Please review it again when you get a chance.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/IntermediateTrinoSqlNodeRewriter.java
```diff
@@ -0,0 +1,104 @@
...
+      case EQUALS:
+        return getTransformedEqualsOperatorSqlCall(sqlCall);
+      case SELECT:
+        return getTransformedSqlSelectSqlCall(sqlCall);
+      default:
+        return sqlCall;
+    }
+  }
+
+  // Append TryCast operator to both operands to cast each operand's data type to VARCHAR
```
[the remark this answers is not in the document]
aastha25: updated the documentation. Please review it again when you get a chance.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/RelToTrinoConverter.java
aastha25: updated the documentation. Please review it again when you get a chance.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/SqlNodeConverter.java
```diff
@@ -0,0 +1,111 @@
...
+        return castOperandsToVarchar(sqlCall);
+      case SELECT:
+        return getTransformedSqlSelectSqlCall(sqlCall);
+      default:
+        return sqlCall;
+    }
+  }
+
+  /**
+   * Coral IR generates a SqlNode representation where operands of a relational operator may not be compatible.
```
wmoustafa, line 65: Coral IR allows implicit casting to VARCHAR from other data types such as INTs, etc. However, Trino requires explicit casting.

wmoustafa (commented):
LGTM after addressing one comment and ensuring regression tests pass.

aastha25:
@wmoustafa the integ tests succeeded and I have addressed the open comment. Could you give it a final pass?

wmoustafa (approved):

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
```diff
@@ -359,6 +359,29 @@ public void testIfWithNullAsThirdParameter() {
...
+    String expandedSql = relToTrinoConverter.convert(relNode);
+    assertEquals(expandedSql, targetSql);
+  }
+
+  @Test
+  public void testRelationalOperators() {
+    RelNode relNode = hiveToRelConverter.convertSql("SELECT 1 > 2 AS col1, 2 <= 3 AS col2 WHERE 3=3");
+    String targetSql =
+        "SELECT TRY_CAST(1 AS VARCHAR) > TRY_CAST(2 AS VARCHAR) AS \"col1\", TRY_CAST(2 AS VARCHAR) <= TRY_CAST(3 AS VARCHAR) AS \"col2\"\n"
+            + "FROM (VALUES  (0)) AS \"t\" (\"ZERO\")\n" + "WHERE TRY_CAST(3 AS VARCHAR) = TRY_CAST(3 AS VARCHAR)";
```
ljfgem, line 378: I am worried about the query result regression if we add `try_cast` for some operators like `>`/`<`.
For example:
HQL:
```
SELECT * FROM t WHERE 10 > 6; // always true
```
Translated Trino SQL:
```
SELECT * FROM t WHERE try_cast(10, varchar) > try_cast(6, varchar); // always false because '10' < '6'
```
This kind of regression can't be caught by regression test, but will affect the query result.
cc: @wmoustafa

ljfgem (commented):
LGTM except one concern.

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
wmoustafa: Good catch. The agreement was to fix this behavior once the type inference work is checked in. Now that it is checked in, maybe we can fix it as part of this PR.

aastha25: @ljfgem @wmoustafa  I have updated the PR to address this concern. PTAL

ljfgem: Thanks @aastha25, have you run the regression test again for this update? Let's ensure there is no translation failure.

ljfgem:
@wmoustafa @aastha25 After investigation and tests, I found that we may not need `RelationalOperatorTransformer` (original [adjustInconsistentTypesToEqualityOperator](https://github.com/linkedin/coral/blob/master/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java#L426)) to begin with.

Reason:
For relational operation like `'1' = 1` where there is a type mismatch in CoralSqlNode1, Calcite will add explicit `CAST` in Coral IR automatically during `CoralSqlNode1 -> Coral IR` and the corresponding converted RexCall becomes 
`=(CAST('1'):INTEGER NOT NULL, 1)`, then CoralSqlNode2 becomes `CAST('1' AS INTEGER) = 1`. So we don't need to add `TRY_CAST` anymore (Actually this PR already rules it out [here](https://github.com/linkedin/coral/pull/329/files#diff-7f8a049f1695fd79362914170997ba47b067eb80d098f5ae01cedeb1dd88a5e3R76)).
Regression test also proved it - there is no added `TRY_CAST` in any production view with this PR.
I have also verified that there is no regression if we just remove [adjustInconsistentTypesToEqualityOperator](https://github.com/linkedin/coral/blob/master/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java#L426) without introducing `RelationalOperatorTransformer` with Trino local server.

In conclusion, I think we can  remove [adjustInconsistentTypesToEqualityOperator](https://github.com/linkedin/coral/blob/master/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java#L426) directly without introducing `RelationalOperatorTransformer` in SqlNode layer.

ljfgem:
@aastha25 As we discussed before, we need to remove the unnecessary `RelationalOperatorTransformer`, so this PR doesn't contain any transformation migrations, then I think it would be great if you can split this PR to 2 PRs to keep PRs smaller and more specific:
1. PR that includes the work in `SqlNodeConverter`, which is related to the work in #315 
2. PR that includes the API refactoring and type derivation enhancement. Since these parts are for transformation migration with where type derivation is needed, I would suggest to add another transformation migration with type derivation (i.e. `visitConcat`) together with them, so that we can add the minimum code and test API refactoring & type derivation enhancement parts.

cc: @wmoustafa

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/SqlSelectListTransformer.java
```diff
@@ -0,0 +1,55 @@
...
+
+import com.linkedin.coral.common.transformers.SqlCallTransformer;
+
+import static org.apache.calcite.rel.rel2sql.SqlImplementor.*;
+
+
+/**
+ * This transformer operates on SqlSelect type sqlCalls. It appends an alias to all the projected fields with fully qualified names.
+ */
+public class SqlSelectListTransformer extends SqlCallTransformer {
```
wmoustafa, line 28: Could you make the name descriptive what the transformation does?

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/SqlSelectListTransformer.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.validate.SqlValidatorUtil;
+
+import com.linkedin.coral.common.transformers.SqlCallTransformer;
+
+import static org.apache.calcite.rel.rel2sql.SqlImplementor.*;
+
+
+/**
+ * This transformer operates on SqlSelect type sqlCalls. It appends an alias to all the projected fields with fully qualified names.
```
wmoustafa, line 26: Could you add an example and a justification for the transformation in the comment?

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/RelToTrinoConverter.java
```diff
@@ -105,6 +104,14 @@ public Result visit(Window window) {
...
   }
 
+  /**
+   * A Project RelNode represents a relational operator that projects a subset of columns from an input relational expression.
+   * When projecting a NULL column, super's implementation casts this column to the derived data type to ensure that the column can be interpreted correctly.
+   * If the derived data type is NULL, super's implementation casts the NULL column to a NULL data type.
+   * This override prevents appending NULL cast on a NULL column.
+   * @param e Project RelNode.
+   * @return Result of converting the RelNode to a SqlNode.
+   */
```
wmoustafa, line 114: What makes this PR include this change? Could you explain the need for it in the description?

wmoustafa: Also, should we assume transformation at this layer will be converted later to SqlNode level?

wmoustafa:
Let us move the tranformation in #373 to SqlNode layer as well.

aastha25:
> Let us move the tranformation in #373 to SqlNode layer as well.
@wmoustafa  we can include the corresponding SqlNode transformer - `RelationalOperatorTransformer` but it'll be a no-op given the existing translation path.

wmoustafa:
> @wmoustafa we can include the corresponding SqlNode transformer - `RelationalOperatorTransformer` but it'll be a no-op given the existing translation path.

I think it is fine. At some point we should add tests that do not go through RelNodes.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/RelToTrinoConverter.java
aastha25: I'll add an example to clarify the need for the override. 
At this point, we should expect this transformation to stay here in the Rel -> SqlNode layer. This transformation helps achieve parity between sqlNode1 & sqlNode2.

wmoustafa: Will it then move to Coral RelNode to Coral SqlNode converter?

ljfgem:
> Let us move the tranformation in #373 to SqlNode layer as well.

@wmoustafa If we want to ensure CoralSqlNode1 == CoralSqlNode2, I would prefer to add a transformation on `HiveSqlNode -> CoralSqlNode1` conversion path to add explicit `CAST` for type mismatch in HQL because:
1. Even if we move the transformation in #373 to SqlNode layer (`CoralSqlNode2 -> TrinoSqlNode`), CoralSqlNode1 != CoralSqlNode2 because Calcite adds the explicit `CAST`
2. Transformation in #373 just replaced `CAST` by `TRY_CAST`, which is not necessary for Trino SQL as I mentioned before:
```
If CAST succeeds, then CAST is equivalent to TRY_CAST.
If CAST fails, a query using CAST fails, while a query using TRY_CAST succeeds. However, since TRY_CAST returns null in this case, it may lead to query result data quality issues, which is undesirable.
```
And I would prefer to add/migrate transformations in another PR because #329 focuses on adding `SqlSelectListTransformer`, it would be great to keep PRs small and specific.

wmoustafa:
@ljfgem It is still equivalent since the Coral IR spec in that case is that we are just saying it the sides of comparison operators do not necessarily match. If through the RelNode path the CAST was introduced, it is still acceptable. It is optional in the first place.
For `CAST` vs `TRY_CAST`, yes I prefer `CAST`.
I will let @aastha25 decide on which PR.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/SqlSelectListTransformer.java
aastha25: done.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/SqlSelectListTransformer.java
aastha25: I have added an example in the java doc. I have also updated the PR description to explain the motivation. Tldr: This is a migration from RelNode -> SqlNode

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/RelToTrinoConverter.java
aastha25: yes.

aastha25:
> @ljfgem SqlNode1 and SqlNode2 are still equivalent since the Coral IR spec in that case is that we are just saying that the sides of comparison operators do not necessarily match. If through the RelNode path the CAST was introduced, it is still acceptable. It is optional in the first place. For `CAST` vs `TRY_CAST`, yes I prefer `CAST`. I will let @aastha25 decide on which PR.

I prefer to add `RelationalOperatorTransformer` it in a separate PR after #378 is merged. This is a transformer that needs data type derivation and will rely on the new API for type derivation.

ljfgem (commented):
Thanks @aastha25 for this PR!

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/RelToTrinoConverter.java
```diff
@@ -338,41 +331,6 @@ public SqlNode toSql(RexProgram program, RexNode rex) {
         }
         return super.toSql(program, rex);
       }
-
-      @Override
-      public SqlNode field(int ordinal) {
```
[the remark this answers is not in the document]
aastha25: this seems to be a part of the legacy code, that is not used. Removing it has no impact on translations.

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
```diff
@@ -731,12 +756,25 @@ public void testCastVarbinaryToVarchar() {
...
+    RelNode relNode =
+        TestUtils.getHiveToRelConverter().convertSql("SELECT CAST(b AS CHAR(255)) FROM test.table_with_binary_column");
     String targetSql = "SELECT \"from_utf8\"(\"table_with_binary_column\".\"b\")\n"
         + "FROM \"test\".\"table_with_binary_column\" AS \"table_with_binary_column\"";
     String expandedSql = relToTrinoConverter.convert(relNode);
     assertEquals(expandedSql, targetSql);
   }
+
+  @Test
+  public void testNew() {
```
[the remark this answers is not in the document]
aastha25: renamed the function.

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
```diff
@@ -215,242 +214,264 @@ public void testLateralViewArray() {
...
-        new RelToTrinoConverter(ImmutableMap.of(SUPPORT_LEGACY_UNNEST_ARRAY_OF_STRUCT, true));
+        TestUtils.getRelToTrinoConverter(ImmutableMap.of(SUPPORT_LEGACY_UNNEST_ARRAY_OF_STRUCT, true));
     String expandedSql = relToTrinoConverter.convert(relNode);
     assertEquals(expandedSql, targetSql);
   }
 
   @Test
-  public void testAvoidTransformToDate() {
-    RelNode relNode = hiveToRelConverter
+  public void testAvoidTransformToDate() throws MetaException {
```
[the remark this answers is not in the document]
aastha25: removed the exception

ljfgem (commented):
LGTM except [a remaining comment](https://github.com/linkedin/coral/pull/329/files#r1156435615)
@wmoustafa @yiqiangin please also take a look if you get a chance.

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
```diff
@@ -754,9 +780,20 @@ public void testRlikeTransformation() {
...
 
-    RelNode relNode = hiveToRelConverter.convertSql("SELECT '1' NOT REGEXP '^1$'");
+    RelNode relNode = TestUtils.getHiveToRelConverter().convertSql("SELECT '1' NOT REGEXP '^1$'");
     String targetSql = "SELECT NOT \"REGEXP_LIKE\"('1', '^1$')\n" + "FROM (VALUES  (0)) AS \"t\" (\"ZERO\")";
     String expandedSql = relToTrinoConverter.convert(relNode);
     assertEquals(expandedSql, targetSql);
   }
+
+  public void testSqlSelectAliasAppenderTransformer() {
+    //test.tableA(a int, b struct<b1:string>
```
ljfgem, line 790: nit: a space after `//`, or we can remove this comment since there is no similar comment for other UTs.

aastha25: I can fix it in the next PR if its not a blocking concern. Thanks

PR #331, opened 2022-11-16
Fix the schema field namespace before converting to lowercase

aastha25:
`relToAvroSchemaConverter` converts a relNode representation into an Avro schema. During this transformation, the fields are assigned a default namespace: `rel_avro.<fieldName>` . If a schema has nested fields with the same fieldName and this schema is converted to lowercase by `ToLowercaseSchemaVisitor`, the conversion fails a check [here](https://github.com/linkedin/coral/blob/master/coral-schema/src/main/java/com/linkedin/coral/schema/avro/AvroSchemaVisitor.java#L23) as the visitor detects two nested fields with the same namespace.  

This PR reorders "setupNameAndNamespace" & "convertToLowerCase" transformations so that `ToLowercaseSchemaVisitor` takes a correct schema as an input. 

Since the transformations are additive, the reordering should be safe. 

Testing done:
unit tests
tested the affected view on production


ljfgem (approved):
LGTM, thanks @aastha25!
Let's merge once the i-test passes.

aastha25:
Thanks for helping out with the integ tests @ljfgem ! looks like they passed. 
Lets merge the PR now.

PR #332, opened 2022-11-29
Remove 2 wrongly registered UDFs

ljfgem:
These 2 UDFs are registered with wrong return types (`SqlTypeName.ANY` should never appear in the UDF return type), removing them given they can be handled by the dynamic Hive generic UDF registration. Didn't choose to modify registered return type given the return types are too complex to register manually, and might be modified afterwards. We should only register UDFs that can't be handled by the dynamic Hive generic UDF registration.

Note: We need to be careful about the return type for UDF registration, because it would affect the RelNode data type [consumed by Trino](https://github.com/trinodb/trino/blob/master/plugin/trino-hive/src/main/java/io/trino/plugin/hive/ViewReaderUtil.java#L218) for return type validation.

### DIFF
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
@@ -429,11 +429,6 @@ public class StaticHiveFunctionRegistry implements FunctionRegistry {
         STRING_STRING);
     createAddUserDefinedFunction("com.linkedin.dali.view.udf.entityhandles.PhoneNumberNormalizer",
         FunctionReturnTypes.STRING, STRING_STRING_STRING);
-    createAddUserDefinedFunction("com.linkedin.dali.views.feed.udf.GenerateDecoratedFeedUpdateData",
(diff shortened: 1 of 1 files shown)


aastha25:
Thanks for the PR @ljfgem !
Did this get surfaced as part of some bug report?

ljfgem:
> Thanks for the PR @ljfgem ! Did this get surfaced as part of some bug report?

No, the views can be translated well to Trino SQL, but the type of RelNode is not correct, which would cause issue like:
```
Caused by: io.trino.spi.type.TypeNotFoundException: Unknown type: any
  at io.trino.metadata.TypeRegistry.instantiateParametricType(TypeRegistry.java:180)
  at io.trino.metadata.TypeRegistry.lambda$getType$0(TypeRegistry.java:148)
  at com.google.common.cache.LocalCache$LocalManualCache$1.load(LocalCache.java:4853)
  at com.google.common.cache.LocalCache$LoadingValueReference.loadFuture(LocalCache.java:3529)
  at com.google.common.cache.LocalCache$Segment.loadSync(LocalCache.java:2278)
  at com.google.common.cache.LocalCache$Segment.lockedGetOrLoad(LocalCache.java:2155)
```
because Trino would validate the query return type against the RelNode type.

It was caught by the test Trino server approach that I proposed.

aastha25:
I see, thanks for elaborating. 
LGTM!

PR #335, opened 2023-01-04
Coral-Hive: Cache UDF after dynamic resolution

ljfgem:
After resolving an UDF dynamically, we can cache it to avoid duplicated dynamic resolution. This change will shorten the translation time for some prod views because UDF dynamic resolution might take long.

Tested on the prod views, no regression.


aastha25:
LGTM! thanks @ljfgem

PR #336, opened 2023-01-04
Create SqlCallTransformer for generic transformations on SqlCall

ljfgem:
This PR creates an abstract class `SqlCallTransformer` for generic SqlCall transformations. `SqlCallTransformer` can be utilized to facilitate interconversions between languageSqlNode and CoralSqlNode, particularly when information about the operand and operator datatype is necessary for the transformations. It provides following APIs:

1. `protected abstract boolean predicate(SqlCall sqlCall)`: predicate of the transformer, it’s used to determine if the SqlCall should be transformed or not
2. `protected abstract SqlCall transform(SqlCall sqlCall)`: implementation of the transformation, returns the transformed SqlCall
3. `public SqlCall apply(SqlCall sqlCall)`: public entry of the transformer, it returns the result of 2 (transformed SqlCall) if condition(SqlCall sqlCall) returns true, otherwise returns the input SqlCall without any transformation
4. `protected RelDataType getRelDataType(SqlNode sqlNode)`: helper method to derive the datatype of the input SqlNode with SqlValidator, it’s only used when sqlValidator is provided in the constructor

To use `SqlCallTransformer`, this PR creates a container class `SqlCallTransformers`, with which we can create a `SqlCallTransformers` with `of` method and call `apply` to apply all the SqlCallTransformers to the SqlCall.
After defining a set of transformers, we need a SqlShuttle to traverse the SqlNode tree and apply the transformers to the SqlCalls, like [HiveSqlNodeToCoralSqlNodeConverter](https://github.com/linkedin/coral/pull/336/files#diff-836bb2593923651887a5d9c065ff1fa04cdafc5af872d39de01ee2e54e004bfd). For Coral-Spark and Coral-Trino, we can add this part in [CoralSqlNodeToSparkSqlNodeConverter](https://github.com/linkedin/coral/blob/master/coral-spark/src/main/java/com/linkedin/coral/spark/CoralSqlNodeToSparkSqlNodeConverter.java#L38) and CoralSqlNodeToTrinoSqlNodeConverter.

Tests:
1. Existing UTs, which already contains the tests for array index transformation
2. i-test on all the prod views, no regression


coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveToRelConverter.java
```diff
@@ -92,12 +93,16 @@ protected SqlToRelConverter getSqlToRelConverter() {
...
-    return parseTreeBuilder.process(trimParenthesis(sql), hiveView);
+    final SqlNode sqlNode = parseTreeBuilder.process(trimParenthesis(sql), hiveView);
+    if (hiveView != null) {
+      sqlNode.accept(new FuzzyUnionSqlRewriter(hiveView.getTableName(), this));
+    }
+    return sqlNode.accept(new HiveToCoralOperatorConverter(getSqlValidator()));
   }
 
   @Override
   protected RelNode standardizeRel(RelNode relNode) {
```
aastha25, line 104: can we clean up this method as well? or keep it and restructure it to do HiveToCoral conversions here?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveToRelConverter.java
```diff
@@ -92,12 +93,16 @@ protected SqlToRelConverter getSqlToRelConverter() {
 
   @Override
   protected SqlNode toSqlNode(String sql, Table hiveView) {
-    return parseTreeBuilder.process(trimParenthesis(sql), hiveView);
+    final SqlNode sqlNode = parseTreeBuilder.process(trimParenthesis(sql), hiveView);
+    if (hiveView != null) {
+      sqlNode.accept(new FuzzyUnionSqlRewriter(hiveView.getTableName(), this));
+    }
+    return sqlNode.accept(new HiveToCoralOperatorConverter(getSqlValidator()));
```
aastha25, line 100: the sqlValidator has not validated the node yet. The validation process populates the SqlValidator with the context of the RelNode here https://github.com/linkedin/coral/blob/master/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveSqlToRelConverter.java#L59 . Before this step, the sqlValidator is empty.

If we are going to use an empty sqlValidator and populate it inside each "OperatorTransformer", then why do we pass it at all? Should we consider manufacturing it inside HiveToCoralOperatorConverter or OperatorTransformer?

coral-hive/src/main/java/com/linkedin/coral/transformers/OneBasedArrayIndexTransformer.java
```diff
@@ -0,0 +1,51 @@
...
+
+
+/**
+ * Transformer to convert SqlCall from array[i] to array[i+1] to ensure array indexes start at 1.
+ */
+public class OneBasedArrayIndexTransformer extends OperatorTransformer {
+  @Override
+  public boolean condition() {
+    if (inputSqlNode instanceof SqlBasicCall
+        && "ITEM".equalsIgnoreCase(((SqlBasicCall) inputSqlNode).getOperator().getName())) {
```
aastha25, line 29: nit: Please declare "ITEM" as a static constant

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilderTest.java
```diff
@@ -85,7 +85,7 @@ public Iterator<Object[]> getConvertSql() {
         // ORDER BY
         "SELECT a, b from foo order by a DESC",
         // complex types
-        "SELECT c[0], s.name from complex",
```
aastha25, line 88: why is this test excluded?

coral-common/src/main/java/com/linkedin/coral/common/transformers/OperatorTransformer.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.calcite.sql.validate.SqlValidator;
+
+
+/**
+ * Abstract class for generic transformations on SqlNodes
+ */
+public abstract class OperatorTransformer {
+  protected SqlNode inputSqlNode;
+  protected SqlValidator sqlValidator;
+  protected List<SqlSelect> topSelectNodes = new ArrayList<>();
```
aastha25, line 24: why is topSelectnodes a list? 
For a given SQL, we should just have 1 top level SelectNode, ie, the top level query?

coral-common/src/main/java/com/linkedin/coral/common/transformers/OperatorTransformer.java
```diff
@@ -0,0 +1,75 @@
...
+   */
+  protected RelDataType getRelDataType(SqlNode sqlNode) {
+    if (sqlValidator == null) {
+      throw new RuntimeException("Please provide sqlValidator to get the RelDataType of a SqlNode!");
+    }
+    // Traverse the stored top select SqlNodes from new to old to construct the dummy SqlSelect,
+    // return the SqlNode's RelDataType directly if SqlValidator can validate the dummy SqlSelect
+    for (int i = topSelectNodes.size() - 1; i >= 0; --i) {
+      final SqlSelect topSelectNode = topSelectNodes.get(i);
+      final SqlSelect dummySqlSelect = new SqlSelect(topSelectNode.getParserPosition(), null, SqlNodeList.of(sqlNode),
```
aastha25, line 65: (1) What are the criteria for manufacturing this dummySelectNode? Why do we selectively use "topSelectNode.getFrom()" but not the other clauses? The sqlValidator constructed like this may not be the right SOT.
instead of creating the dummySelectNode, can we do the following - 

//take the input SqlNode (SqlCall) and construct a dummyRelNode
RelNode dummyRelNode = hiveToRelConverter.toRel(sqlCall);

//extract the dummy validator (will need to make the method public)
SqlValidator validator = hiveToRelConverter.getSqlValidator();

//get the top-level scope from the top level selectNode - 
SqlValidatorScope topLevelScope = validator.getSelectScope((SqlSelect) sqlCall);

//derive the type
final RelDataType argType = validator.deriveType(topLevelScope, sqlNode);

(2) topSelectNodes is a list and we want to iterate over the entire list to identify the relDataType. However, we return the first relDataType discovered. That is always going to be derived using topSelectNodes[0]? Then why is  topSelectNodes a list?
Could you explain the process of deriving the data type here?

coral-common/src/main/java/com/linkedin/coral/common/transformers/OperatorTransformer.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlNodeList;
+import org.apache.calcite.sql.SqlSelect;
+import org.apache.calcite.sql.validate.SqlValidator;
+
+
+/**
+ * Abstract class for generic transformations on SqlNodes
+ */
+public abstract class OperatorTransformer {
```
aastha25, line 21: nit: Not sure about the naming here. This is a SqlCallTranformer, which can modify any aspect of a SqlCall, not just the operator.
Will get back with some naming ideas.

aastha25 (commented):
Thanks for the PR @ljfgem! Lets discuss offline how we can best introduce SqlValidator in the path: hive SQL -> hive SqlNode -> Coral SqlNode -> CoralRelNode. 

We can also consider absorbing the logic of OneBasedArrayIndexTransformer in the parseTreeBuilder.

coral-common/src/main/java/com/linkedin/coral/common/transformers/StandardOperatorTransformer.java
```diff
@@ -0,0 +1,277 @@
...
+import org.apache.calcite.sql.type.SqlReturnTypeInference;
+import org.apache.calcite.sql.validate.SqlUserDefinedFunction;
+
+import static com.linkedin.coral.common.calcite.CalciteUtil.*;
+
+
+/**
+ * Standard operator transformer which can be applied to the input SqlNode directly without requiring the type derivation
+ */
+public class StandardOperatorTransformer extends OperatorTransformer {
```
aastha25, line 40: where is this class integrated?
Is this going to be accessed in ParseTreeBuilder?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveToRelConverter.java
ljfgem: I think we can remove this method given we are migrating from RelNode transformation to SqlNode transformation. Will remove it.

coral-common/src/main/java/com/linkedin/coral/common/transformers/OperatorTransformer.java
ljfgem: (1) I used `topSelectNode.getFrom()` only because that is the only information required to construct the minimum dummy SqlSelect: `SELECT <SqlNode to get the datatype> FROM xxx`. Other parts (orderby/groupby/...) are not necessary to derive the datatype.

And as for `RelNode dummyRelNode = hiveToRelConverter.toRel(sqlCall);`, I think `SqlCall` should be the top level selectNode instead of the [inputSqlNode](https://github.com/linkedin/coral/pull/336/files#diff-2fb1eb21bc5936ee6b8406a4c9e238c018ac121a14cdc928e8e05a7677fb9887R22) ? Otherwise `toRel` will fail.

Comparing these 2 approaches of getting datatype, I think both of them work for deriving the datatype, but using SqlValidator + dummy Select would require less resources, because constructing a `ToRelConverter` requires more things apart from `SqlValidator`, and converting the whole top level selectNode would take more resources than validating the minimum dummy SqlSelect.

(2) To begin with, I only stored the last visited topSelectNode to construct the dummy SqlNode, but I found a corner case in the regression test. For example, in the following SQL:
```
SELECT *
FROM (
   SELECT * FROM db.tbl
)
LATERAL JOIN EXPLODE(col1) t AS a
WHERE t.a > 0
```
If I want to derive the datatype of `t.a`, the last visited top level SqlNode is `SELECT * FROM db.tbl`, however, `t.a` doesn't exist in `db.tbl` so it would throw exception. 
So we need to store all the top level SqlSelects (both inner `SELECT * FROM db.tbl` and the whole SQL) in the list and traverse them from end to beginning, return the datatype directly once the datatype can be derived without exception.
I will add more contexts in the comment of this method to elaborate its logic.

coral-common/src/main/java/com/linkedin/coral/common/transformers/StandardOperatorTransformer.java
ljfgem: It's not integrated yet, @yiqiangin is working to replace [CalciteTrinoUDFMap](https://github.com/linkedin/coral/blob/93ae233c63872f2f98c2a69adc1145c446a7fcd9/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/CalciteTrinoUDFMap.java#L26) with it.

coral-hive/src/main/java/com/linkedin/coral/transformers/OneBasedArrayIndexTransformer.java
ljfgem: Thanks for suggestion, modified.

coral-common/src/main/java/com/linkedin/coral/common/transformers/OperatorTransformer.java
ljfgem: (1) I used `topSelectNode.getFrom()` only because that is the only necessary information required to construct the minimum dummy SqlSelect: `SELECT <SqlNode to get the datatype> FROM <topSelectNode.getFrom()>`. Other parts (orderby/groupby/...) are not needed to derive the datatype.

I have implemented the approach that you mentioned to get the datatype, but there are some regressions on the prod views. Besides, using SqlValidator + dummy sqlSelect would require less resources, because constructing a `ToRelConverter` requires more things apart from SqlValidator, and converting the whole top level selectNode would take more resources than validating the minimum dummy SqlSelect.

(2) I have added the [detailed comment](https://github.com/linkedin/coral/pull/336/files#diff-2fb1eb21bc5936ee6b8406a4c9e238c018ac121a14cdc928e8e05a7677fb9887R50) to the method, PTAL.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveToRelConverter.java
ljfgem: In order to create a new SqlValidator in `OperatorTransformer`/`HiveSqlNodeToCoralSqlNodeConverter`, we need to provide/instantiate lots of required parameters, because [the instantiation of sqlValidator is pretty complex](https://github.com/linkedin/coral/blob/39d4ba362cf4cd2fd83874ed2ef2f607225b328f/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveToRelConverter.java#L53). 
Therefore, I would prefer to use the one in `HiveToRelConverter` directly for simplification, and it doesn't cause regression.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveSqlNodeToCoralSqlNodeConverter.java
```diff
@@ -0,0 +1,31 @@
...
+public class HiveSqlNodeToCoralSqlNodeConverter extends SqlShuttle {
+  private final OperatorTransformerList operatorTransformerList;
+
+  public HiveSqlNodeToCoralSqlNodeConverter(SqlValidator sqlValidator) {
+    operatorTransformerList = OperatorTransformerList.of(new OneBasedArrayIndexTransformer(sqlValidator));
+  }
+
+  @Override
+  public SqlNode visit(SqlCall call) {
+    return super.visit(operatorTransformerList.apply(call));
```
wmoustafa, line 29: `OperatorTransfomers` is clearer.

coral-common/src/main/java/com/linkedin/coral/common/transformers/OperatorTransformer.java
wmoustafa: +1 to the name change. I was waiting for the investigation on `SqlBasicCall` as well to suggest names.

coral-common/src/main/java/com/linkedin/coral/common/transformers/OperatorTransformer.java
```diff
@@ -0,0 +1,81 @@
...
+
+  public OperatorTransformer() {
+
+  }
+
+  public OperatorTransformer(SqlValidator sqlValidator) {
+    this.sqlValidator = sqlValidator;
+  }
+
+  protected abstract boolean condition(SqlCall sqlCall);
```
wmoustafa, line 34: `predicate`?

coral-common/src/main/java/com/linkedin/coral/common/transformers/OperatorTransformer.java
```diff
@@ -0,0 +1,81 @@
...
+
+  protected abstract SqlCall transform(SqlCall sqlCall);
+
+  public SqlCall apply(SqlCall sqlCall) {
```
wmoustafa, line 38: Consider `SqlBasicaCall`?

coral-common/src/main/java/com/linkedin/coral/common/transformers/OperatorTransformer.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  /**
+   * To get the RelDatatype of a SqlNode, we iterate through `topSelectNodes` in reverse chronological order,
+   * for each `topSelectNode`, we create a minimum dummy SqlSelect: SELECT `sqlNode` FROM `topSelectNode.getFrom()`
+   * If the SqlValidator is able to validate the dummy SqlSelect, return the SqlNode's RelDataType directly.
+   *
+   * We can't just use the latest visited `topSelectNode` to construct the dummy SqlSelect because of
+   * the following corner case:
+   * SELECT * FROM (
```
wmoustafa, line 56: Please list the tables and columns before the example.

coral-common/src/main/java/com/linkedin/coral/common/transformers/OperatorTransformer.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+    if (condition(sqlCall)) {
+      return transform(sqlCall);
+    } else {
+      return sqlCall;
+    }
+  }
+
+  /**
+   * To get the RelDatatype of a SqlNode, we iterate through `topSelectNodes` in reverse chronological order,
```
wmoustafa, line 50: I feel `reverse chronological` is not capturing the intended ordering. Maybe it will be clearer after we finish reviewing the example below.

coral-common/src/main/java/com/linkedin/coral/common/transformers/OperatorTransformer.java
```diff
@@ -0,0 +1,81 @@
...
+   * for each `topSelectNode`, we create a minimum dummy SqlSelect: SELECT `sqlNode` FROM `topSelectNode.getFrom()`
+   * If the SqlValidator is able to validate the dummy SqlSelect, return the SqlNode's RelDataType directly.
+   *
+   * We can't just use the latest visited `topSelectNode` to construct the dummy SqlSelect because of
+   * the following corner case:
+   * SELECT * FROM (
+   *   SELECT col1 FROM db.tbl
+   * ) LATERAL JOIN EXPLODE(col1) t AS a WHERE t.a = 0
+   * If we want to derive the datatype of `t.a`, the latest visited `topSelectNode` will be `SELECT * FROM db.tbl`,
```
wmoustafa, line 59: Is it `SELECT col1`?

coral-common/src/main/java/com/linkedin/coral/common/transformers/OperatorTransformer.java
```diff
@@ -0,0 +1,81 @@
...
+    this.sqlValidator = sqlValidator;
+  }
+
+  protected abstract boolean condition(SqlCall sqlCall);
+
+  protected abstract SqlCall transform(SqlCall sqlCall);
+
+  public SqlCall apply(SqlCall sqlCall) {
+    if (sqlCall instanceof SqlSelect) {
+      this.topSelectNodes.add((SqlSelect) sqlCall);
```
wmoustafa, line 40: I think this conflates order derivation with transformation, i.e., we are assuming a transformation, and specific ordering of transformation to take place first in order to be able to derive the types. Should not type derivation be a standalone class? cc: @aastha25.

coral-hive/src/main/java/com/linkedin/coral/transformers/OneBasedArrayIndexTransformer.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.sql.validate.SqlValidator;
+
+import com.linkedin.coral.common.transformers.OperatorTransformer;
+
+
+/**
+ * Transformer to convert SqlCall from array[i] to array[i+1] to ensure array indexes start at 1.
+ */
+public class OneBasedArrayIndexTransformer extends OperatorTransformer {
```
wmoustafa, line 25: `ShiftArrayIndexTransformer`? Since we need to highlight the operation?

coral-common/src/main/java/com/linkedin/coral/common/transformers/StandardOperatorTransformer.java
```diff
@@ -0,0 +1,277 @@
...
+import org.apache.calcite.sql.validate.SqlUserDefinedFunction;
+
+import static com.linkedin.coral.common.calcite.CalciteUtil.*;
+
+
+/**
+ * Class to instantiate standard transformers, whose condition(SqlCall) method only checks operator name and number of operands
+ * and transform(SqlCall) method is achieved by JSON-format transformers
+ */
+public class StandardOperatorTransformer extends OperatorTransformer {
```
wmoustafa, line 41: Is this going to be committed as part of this patch or will @yiqiangin apply it in his patch? If the latter let us remove it. If the former, I think the name could be more expressive.

coral-hive/src/main/java/com/linkedin/coral/transformers/OneBasedArrayIndexTransformer.java
```diff
@@ -0,0 +1,57 @@
...
+
+  private static final String ITEM_OPERATOR = "ITEM";
+
+  public OneBasedArrayIndexTransformer(SqlValidator sqlValidator) {
```
wmoustafa, line 29: Why does the caller need to provide the SqlValidator? Is there a way to set it automatically?

coral-hive/src/main/java/com/linkedin/coral/transformers/OneBasedArrayIndexTransformer.java
```diff
@@ -0,0 +1,57 @@
...
+    super(sqlValidator);
+  }
+
+  @Override
+  public boolean condition(SqlCall sqlCall) {
+    if (ITEM_OPERATOR.equalsIgnoreCase(sqlCall.getOperator().getName())) {
+      final SqlNode columnNode = sqlCall.getOperandList().get(0);
+      return getRelDataType(columnNode) instanceof ArraySqlType;
```
wmoustafa, line 37: Do you agree this has an assumption that the caller is coming from a SqlShuttle? What are the conditions on the SqlNode that can be passed here? Seems it is restricted (I think must be at the same SELECT as the condition's sqlCall). Is there an additional requirement that it must be a child of it?

ljfgem: Yes, the caller is coming from a SqlShuttle, but that is the case for all the SqlNode converters like `CoralSqlNodeToSparkSqlNodeConverter`. 
The condition is that the parent SqlSelect should be visited before we call this method on a child sqlNode, but it's still the case for the SqlNode traversal pattern. I don't think we need to derive the type of a SqlNode without visiting its parent SqlSelect in advance. (If we don't visit the parent SqlSelect first, we can't even transform the child SqlNode of it)

coral-hive/src/main/java/com/linkedin/coral/transformers/OneBasedArrayIndexTransformer.java
ljfgem: In order to derive the datatype, we need to validate the SqlNode first, and SqlValidator contains necessary information (like table schema in HMS, and legal operators in `DaliOperatorTable`) for type derivation.

coral-common/src/main/java/com/linkedin/coral/common/transformers/SqlCallTransformer.java
```diff
@@ -0,0 +1,95 @@
...
+   */
+  protected abstract boolean predicate(SqlCall sqlCall);
+
+  /**
+   * Implementation of the transformation, returns the transformed SqlCall
+   */
+  protected abstract SqlCall transform(SqlCall sqlCall);
+
+  /**
+   * Public entry of the transformer, it returns the result of transformed SqlCall if `condition(SqlCall)` returns true,
```
aastha25, line 45: `if `predicate(SqlCall)` returns true,`

coral-common/src/main/java/com/linkedin/coral/common/transformers/SqlCallTransformer.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+    if (predicate(sqlCall)) {
+      return transform(sqlCall);
+    } else {
+      return sqlCall;
+    }
+  }
+
+  /**
+   * To get the RelDatatype of a SqlNode, we iterate through `topSelectNodes` from the latest visited to the oldest visited,
```
aastha25, line 60: could you add a note/todo here that this implementation is subject to change and also tag the corresponding coral issue?

ljfgem: Added a note, will open an issue after this PR is merged, so I didn't tag the issue.

wmoustafa:
LGTM except implicit call order assumption in `SqlCallTransformer.getRelDataType()`.
Please file issue for (1) the above and for (2) investigating using `SqlBasicCall` in `SqlCallTransformer` then we can merge after @aastha25 and @yiqiangin approval.

ljfgem:
> LGTM except implicit call order assumption in `SqlCallTransformer.getRelDataType()`. Please file issue for (1) the above and for (2) investigating using `SqlBasicCall` in `SqlCallTransformer` then we can merge after @aastha25 and @yiqiangin approval.

Thanks! Will open issues once this PR is merged. @aastha25 @yiqiangin please take a look if you get a chance.

wmoustafa:
Also please update the JavaDoc of the current `SqlCallTransformer.getRelDataType()` with the precise expectation from the caller of this API to behave correctly.

coral-common/src/main/java/com/linkedin/coral/common/transformers/SqlCallTransformers.java
```diff
@@ -0,0 +1,42 @@
...
+public class SqlCallTransformers {
+  private final List<SqlCallTransformer> operatorTransformers;
+
+  SqlCallTransformers(List<SqlCallTransformer> operatorTransformers) {
+    this.operatorTransformers = operatorTransformers;
+  }
+
+  public static SqlCallTransformers of(SqlCallTransformer... operatorTransformers) {
+    return new SqlCallTransformers(Arrays.asList(operatorTransformers));
+  }
```
wmoustafa, line 26: Usually `of` is associated with immutable collections. Not sure it is reasonable to have both `add` and `of`. Something to consider.

coral-common/src/main/java/com/linkedin/coral/common/transformers/SqlCallTransformer.java
```diff
@@ -0,0 +1,98 @@
...
+
+
+/**
+ * Abstract class for generic transformations on SqlCalls
+ */
+public abstract class SqlCallTransformer {
+  private SqlValidator sqlValidator;
+  private final List<SqlSelect> topSelectNodes = new ArrayList<>();
+
+  public SqlCallTransformer() {
```
aastha25, line 26: are we adding an empty constructor here to support Standard Sqlcall transformers where a sqlValidator is not required?

aastha25 (commented):
LGTM! 
Could you update the description and add the motivation for introducing sqlCallTransformers too? Something along the lines that "it can be used to support interconversions between languageSqlNode and CoralSqlNode, especially when knowledge about operand / operator data type is needed to make transformations".

coral-common/src/main/java/com/linkedin/coral/common/transformers/SqlCallTransformer.java
ljfgem: Yes, it's needed for transformers without SqlValidator.

coral-common/src/main/java/com/linkedin/coral/common/transformers/SqlCallTransformers.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public static SqlCallTransformers of(List<SqlCallTransformer> operatorTransformers) {
+    return new SqlCallTransformers(operatorTransformers);
+  }
+
+  public void register(SqlCallTransformer operatorTransformer) {
+    operatorTransformers.add(operatorTransformer);
```
yiqiangin: how about adding a null check on `operatorTransformers`?

yiqiangin (approved):
LGTM

ljfgem:
Thanks @wmoustafa @aastha25 @yiqiangin for your review! I have addressed all the comments and kicked off the regression test again, will merge it once the regression test passes.

PR #338, opened 2023-01-10
Initial draft of migrating CalciteTrinoUDFMap to the use of StandardOperatorTransformer

yiqiangin:
Based on Jiefan's [PR](https://github.com/linkedin/coral/pull/336/files#diff-2fb1eb21bc5936ee6b8406a4c9e238c018ac121a14cdc928e8e05a7677fb9887), this PR creates a new class of `CalciteTrinoOperatorTransformer` which extends the class of `OperatorTransformerList`. It is a singleton class and initialize all StandardOperatorTransformer according to `UDFTransformers` defined in `CalciteTrinoUDFMap`

Adding a class of `TrinoSqlUDFConverter` which extends the class of `SqlShuttle`. Calling it in `RelToTrinoConverter.convert()` to traverse SqlNode to convert UDF.
To be noted that, there are 5 unit test cases failing because of the following two reasons:

* The operator `timestamp` is not defined in `OP_MAP` in the class of `StandardOperatorTransformer.java`, so the following test cases fail because of `java.lang.UnsupportedOperationException` including:
   * HiveToTrinoConverterTest.testDateFormatFunction
   * HiveToTrinoConverterTest.testTypeCastForDataAddFunction
   * HiveToTrinoConverterTest.testTypeCastForDateDiffFunction
   * HiveToTrinoConverterTest.testViews[26]
* In the test case of RelToTrinoConverterTest.testMapStructAccess, the sql statement to be tested has the operator `[]` like `mcol[scol].IFIELD`.  This identifier is converted as `ITEM($3, $1)` in RelNode. In the previous implementation, the operator `ITEM` is converted into `element_at` in RelNode. However, `[]` is a literal in SqlIdentifier and not considered as an operator in SqlNode, so it cannot be converted into `element_at` by `StandardOperatorTransformer`.

### DIFF
--- a/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/CalciteTrinoOperatorTransformers.java
+++ b/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/CalciteTrinoOperatorTransformers.java
@@ -1,0 +1,209 @@
+/**
+ * Copyright 2023 LinkedIn Corporation. All rights reserved.
+ * Licensed under the BSD-2 Clause license.
+ * See LICENSE in the project root for license information.
+ */
+package com.linkedin.coral.trino.rel2trino;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.apache.calcite.sql.SqlOperator;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+
+import com.linkedin.coral.com.google.common.base.CaseFormat;
+import com.linkedin.coral.com.google.common.base.Converter;
+import com.linkedin.coral.com.google.common.collect.ImmutableMultimap;
+import com.linkedin.coral.common.functions.Function;
+import com.linkedin.coral.common.transformers.OperatorTransformer;
+import com.linkedin.coral.common.transformers.OperatorTransformerList;
+import com.linkedin.coral.common.transformers.StandardOperatorTransformer;
+import com.linkedin.coral.hive.hive2rel.functions.HiveRLikeOperator;
+import com.linkedin.coral.hive.hive2rel.functions.StaticHiveFunctionRegistry;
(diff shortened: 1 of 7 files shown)


yiqiangin:
> Since our goal is to move the transformations to happen at the `SqlNode` layer, how do we still keep `Calcite2TrinoUDFConverter` which operates at `RelNode` layer?

My understanding about the goal of the story I am working on is to move the transformation defined `CalciteTrinoUDFMap` from `RelNode` layer to `SqlNode` layer. As `Calcite2TrinoUDFConverter` covers other transformations which are not defined in `CalciteTrinoUDFMap`, it is still kept in this PR and can be removed with some future work. Pls correct me if I am wrong.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/CalciteTrinoOperatorTransformers.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+    return instance;
+  }
+
+  private static List<OperatorTransformer> initializeOperatorTransformers() {
+    List<OperatorTransformer> operatorTransformers = new ArrayList<>();
+    // conditional functions
+    operatorTransformers.add(createOperatorTransformer(hiveToCalciteOp("nvl"), 2, "coalesce"));
+    // Array and map functions
+    operatorTransformers.add(createOperatorTransformer(SqlStdOperatorTable.ITEM, 2, TrinoElementAtFunction.INSTANCE));
```
wmoustafa, line 52: Did we get rid of the map and used the list instead for every entry of the map? Does it make sense to use the entire map as one entry in the list of operator transformers? FYI @ljfgem

ljfgem: Yes, we got rid of the map and created transformer for each entry of the map. We can use the entire map as one entry in the list of operator transformers, but I think it's clearer to create transformer for each operator.

PR #340, opened 2023-01-12
Coral-Schema: Use "timestamp-millis" as the correct timestamp logicalType for Spark3 and Avro 1.10

ljfgem:
Previously, Avro 1.7 didn’t actually support timestamp logical type, so we used `"logicalType" : "timestamp"`, and added the handler for `"timestamp"` in LI-Spark2 specially. However, Spark3 doesn't recognize `"timestamp"` anymore, we should modify it to `"timestamp-millis"` which is officially supported by Avro 1.10 and [can be recognized by Spark3](https://github.com/apache/spark/blob/master/connector/avro/src/main/scala/org/apache/spark/sql/avro/AvroDeserializer.scala#L137).

Tested on the affected production view, it could return timestamp type as expected with this patch.


aastha25:
thanks for the PR @ljfgem! LGTM

PR #345, opened 2023-01-13
Coral-Schema: return nullable schema for array item and map value

ljfgem:
Previously, `ToNullableSchemaVisitor` couldn't make the array item or map value nullable, this PR is to fix this issue.

Tests:
1. Unit test
2. Regression test, no regression on prod views
3. Tested on the affected views on gateway, which can be queried with this change


aastha25:
LGTM, thanks for the fix @ljfgem!

PR #349, opened 2023-01-23
Coral-Trino: Migrate function operator transformers defined in CalciteTrinoUDFMap from RelNode layer to SqlNode layer

yiqiangin:
The code changes include:
- Adding a class of `OperatorBasedSqlCallTransformer` which extends `SqlCallTransformer` and its condition is implemented based on the input operator name and the number of the operands
- Adding the class of `LinkedInOperatorBasedSqlCallTransformer` which is a subclass of OperatorBasedSqlCallTransformer and transforms a Coral operator into a Trino operator on SqlNode layer for the LinkedIn specific functions
- Adding a class of `MapStructAccessOperatorTransformer` which is an ad-hoc SqlCallTransformer which converts the map struct access operator "[]" defined from Calcite in a SqlIdentifier into a UDF operator of "element_at", e.g. from "col"["field"] to element_at("col", "field")
- Removing the code of transforming standard UDF operators defined in `CalciteTrinoUDFMap` in the class of `Calcite2TrinoUDFConverter` on RelNode layer
- Adding a class named `CoralToTrinoSqlCallConverter` which extends the class of `SqlShuttle` and initialize a class of `SqlCallTransformers` which containing a list of `SqlCallTransformer` to traverse the hierarchy of a SqlCall and converts the functions from Coral operator to Trino operator if it is required
- Calling `CoralToTrinoSqlCallConverter` to convert function operators on SqlNode layer in the class of `RelToTrinoConverter`
- Adding the class of `ToDateOperatorTransformer` as an Ad Hoc SqlCallTransformer as it needs to check the configuration parameter of `AVOID_TRANSFORM_TO_DATE_UDF` in its condition function
- Some small changes on the names of date related UDF functions which should be recognized on SqlNode layer

**_To be noted that_** the migration from UDF operator transformers based on JSON infra to those implemented in Java code will be done in another several separated PRs

PR #351, opened 2023-02-03
Coral-Spark: Migrate some operator transformers from RelNode layer to SqlNode layer

ljfgem:
Changes:
1. Migrate `convertDaliUDF`, `convertBuiltInUDF` and `fallbackToHiveUdf` from RelNode layer in [IRRelToSparkRelTransformer](https://github.com/linkedin/coral/blob/master/coral-spark/src/main/java/com/linkedin/coral/spark/IRRelToSparkRelTransformer.java#L73) to SqlNode layer. We will migrate other remaining transformers and remove `IRRelToSparkRelTransformer` afterwards.
2. Override `aliasContext` in `CoralRelToSqlNodeConverter` to convert RelNode like `f(x).y.z` to SqlNode which meets our expectation, check its [comment](https://github.com/linkedin/coral/compare/master...ljfgem:coral:coral_spark_transformation_migration?expand=1#diff-c5c8865e887e815b03a63a0de8fe443d3e7eb0b0a913e9920255e9dc55888995R353) for details.
3. Enhance `SparkSqlRewriter` to avoid casting to any type containing `ROW` given Spark doesn't allow that.

Tests:
1. Existing unit tests which already covered tests for these changes, no regression
2. Regression test on all the production views, no regression
3. Querying several complex production views on gateway, no regression


wmoustafa (commented):
A few things:
1- Does it change the engine facing APIs?
2- Does it make sense to merge #349 first?
3- Could you see if you could reuse any of the transformers introduced in #349?

ljfgem:
@wmoustafa:
1- Does it change the engine facing APIs? 
No for this PR, given the 3 migrated transformers in this PR don't need type derivation.

2- Does it make sense to merge https://github.com/linkedin/coral/pull/349 first?
Don't think this PR depends on #349, I can introduce `SourceOperatorMatchSqlCallTransformer` and `OperatorRenameSqlCallTransformer` in this PR for `cardinality -> size` transformation. 
Besides, I think it makes more sense to merge this PR first, see my comment [here](https://github.com/linkedin/coral/pull/349#discussion_r1097575162). If we merge this PR first, #349 doesn't need [this part](https://github.com/linkedin/coral/pull/349/files#diff-a3a3daba44acc4e0b1208c73b4098f34d49eaecdfec6e29c8afbf289abf00a9aR299).
 
3- Could you see if you could reuse any of the transformers introduced in https://github.com/linkedin/coral/pull/349?
Two small transformers `SourceOperatorMatchSqlCallTransformer` and `OperatorRenameSqlCallTransformer` can be reused for `cardinality -> size` transformation.

coral-hive/src/main/java/com/linkedin/coral/transformers/CoralRelToSqlNodeConverter.java
```diff
@@ -345,4 +348,41 @@ private SqlNode generateRightChildForSqlJoinWithLateralViews(BiRel e, Result rig
...
+   *
+   * Calcite will convert this RelNode to a SqlIdentifier directly (check
+   * {@link org.apache.calcite.rel.rel2sql.SqlImplementor.Context#toSql(RexProgram, RexNode)}),
+   * which is not aligned with our expectation since we want to apply transformations on `f(x)` with
+   * {@link com.linkedin.coral.common.transformers.SqlCallTransformer}. Therefore, we override this
+   * method to convert `f(x)` to SqlCall, `.` to {@link com.linkedin.coral.common.functions.FunctionFieldReferenceOperator#DOT}
+   * and `y.z` to SqlIdentifier.
+   */
+  @Override
+  public Context aliasContext(Map<String, RelDataType> aliases, boolean qualified) {
```
aastha25, line 364: what is the coralSqlNode1 and coralSqlNode2 representation for such SQLs before & after introducing this change?

ljfgem: Can you clarify `coralSqlNode1` and `coralSqlNode2`?

aastha25: CoralSqlNode1 is the final sqlNode handed over to calcite for validation & relNode conversion here - https://github.com/linkedin/coral/blob/master/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveSqlToRelConverter.java#L57 

CoralSqlNode2 is the output of https://github.com/linkedin/coral/blob/master/coral-hive/src/main/java/com/linkedin/coral/transformers/CoralRelToSqlNodeConverter.java#L51 when using the [hive dialect ](https://github.com/linkedin/coral/blob/master/coral-hive/src/main/java/com/linkedin/coral/transformers/CoralRelToSqlNodeConverter.java#L71)

ljfgem: Since it only overrides a method in `CoralRelToSqlNodeConverter`, `coralSqlNode1` is not affected by it.

As for coralSqlNode2, it changes as I mentioned in the comment:
Before override: RelNode `f(x).y.z` will be converted to a SqlIdentifier (like a string), no SqlCall at all
After override: RelNode `f(x).y.z` will be converted to `(f(x).y).z`, where f(x) is a SqlCall (operator is `f`), `f(x).y` is a SqlCall (operator is `.`), `(f(x).y).z` is also a SqlCall (operator is `.`).

aastha25: I meant, for a given Hive SQL - "select f(x).y.z from foo", what is the `coralSqlNode1` we construct right now? and is the code change making sure that the `coralSqlNode2` we build is same as `coralSqlNode1` from LHS?

Lets try to add minimum and only necessary overrides in CoralRelToSqlNodeConverter

Also, Is there a UT that captures this corner case?

ljfgem: 1. Yes, for `select f(x).y.z from foo`, the current `coralSqlNode1` is `(f(x).y).z` (all SqlCalls instead of SqlIdentifier), so the code change makes sure that the `coralSqlNode2` we build is same as `coralSqlNode1` from LHS. 
2. I think it is already the minimum override for handling this corner case.
3. No UT for now, but I will add it. Would like to address comments together. Please review and comment if you have concerns on other parts.

coral-hive/src/main/java/com/linkedin/coral/transformers/CoralRelToSqlNodeConverter.java
```diff
@@ -345,4 +348,43 @@ private SqlNode generateRightChildForSqlJoinWithLateralViews(BiRel e, Result rig
...
+        if (rex.getKind() == SqlKind.FIELD_ACCESS) {
+          final List<String> accessNames = new ArrayList<>();
+          RexNode referencedExpr = rex;
+          // Use the loop to get the top-level struct (`f(x)` in the example above),
+          // and store the accessed field names ([`z`, `y`] in the example above, needs to be reversed)
+          while (referencedExpr.getKind() == SqlKind.FIELD_ACCESS) {
+            accessNames.add(((RexFieldAccess) referencedExpr).getField().getName());
+            referencedExpr = ((RexFieldAccess) referencedExpr).getReferenceExpr();
+          }
+          if (referencedExpr.getKind() == SqlKind.OTHER_FUNCTION || referencedExpr.getKind() == SqlKind.CAST) {
```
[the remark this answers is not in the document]
aastha25: Also, lets look at the more recent implementation here https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java#L627 which looks different. 
If this suits out requirements, we can copy paste it and then just call out that porting back a more recent commit.

coral-spark/src/main/java/com/linkedin/coral/spark/transformers/TransportableUDFTransformer.java
```diff
@@ -0,0 +1,97 @@
...
+  public static final String DALI_UDFS_IVY_URL_SPARK_2_12 =
+      "ivy://com.linkedin.standard-udfs-dali-udfs:standard-udfs-dali-udfs:2.0.3?classifier=spark_2.12";
+
+  public enum ScalaVersion {
+    SCALA_2_11,
+    SCALA_2_12
+  }
+
+  @Override
+  protected boolean condition(SqlCall sqlCall) {
```
aastha25, line 57: does it make sense to add another condition - 
`sqlCall.getOperator() instanceof  VersionedSqlUserDefinedFunction` ?
We seem to be assuming that in the transform()

coral-spark/src/main/java/com/linkedin/coral/spark/transformers/FallBackToHiveUDFTransformer.java
```diff
@@ -0,0 +1,78 @@
...
+
+
+/**
+ * After failing to transform UDF with {@link TransportableUDFTransformer},
+ * we use this transformer to fall back to the original Hive UDF defined in
+ * {@link com.linkedin.coral.hive.hive2rel.functions.StaticHiveFunctionRegistry}.
+ * This is reasonable since Spark understands and has ability to run Hive UDF.
+ * Check `CoralSparkTest#testFallBackToHiveUDFTransformer()` for an example.
+ */
+public class FallBackToHiveUDFTransformer extends SqlCallTransformer {
```
aastha25, line 32: General comment about the PR:

Coral-spark module should reflect the transformation of Coral IR -> Spark compatible SQL. It should not assume that source/original functions are hive specific. 
Can we rename avoidable references to hive and call them CoralIR instead?

coral-spark/src/main/java/com/linkedin/coral/spark/transformers/TransportableUDFTransformer.java
```diff
@@ -0,0 +1,97 @@
...
+ */
+public class TransportableUDFTransformer extends SqlCallTransformer {
+  private final String hiveUDFClassName;
+  private final String sparkUDFClassName;
+  private final String artifactoryUrlSpark211;
+  private final String artifactoryUrlSpark212;
+  private final Set<SparkUDFInfo> sparkUDFInfos;
+  private ScalaVersion scalaVersion;
+
+  public TransportableUDFTransformer(String hiveUDFClassName, String sparkUDFClassName, String artifactoryUrlSpark211,
```
aastha25, line 36: Should this have another subclass - LinkedinTransportFunctionTransformer to cleanly separate out LI stuff?

coral-spark/src/main/java/com/linkedin/coral/spark/transformers/TransportableUDFTransformer.java
```diff
@@ -0,0 +1,97 @@
...
+
+  public ScalaVersion getScalaVersionOfSpark() {
+    try {
+      String sparkVersion = SparkSession.active().version();
+      if (sparkVersion.matches("2\\.[\\d\\.]*"))
+        return ScalaVersion.SCALA_2_11;
+      if (sparkVersion.matches("3\\.[\\d\\.]*"))
+        return ScalaVersion.SCALA_2_12;
+      throw new IllegalStateException(String.format("Unsupported Spark Version %s", sparkVersion));
+    } catch (IllegalStateException | NoClassDefFoundError ex) {
```
aastha25, line 92: can refactor it to a `if-else if-else`  block

coral-spark/src/main/java/com/linkedin/coral/spark/transformers/FallBackToHiveUDFTransformer.java
```diff
@@ -0,0 +1,78 @@
...
+  public FallBackToHiveUDFTransformer(Set<SparkUDFInfo> sparkUDFInfos) {
+    this.sparkUDFInfos = sparkUDFInfos;
+  }
+
+  @Override
+  protected boolean condition(SqlCall sqlCall) {
+    final String functionClassName = sqlCall.getOperator().getName();
+    if (UNSUPPORTED_HIVE_UDFS.contains(functionClassName)) {
+      throw new UnsupportedUDFException(functionClassName);
+    }
```
aastha25, line 58: I understand why we want to throw an exception (thanks for the detailed comment). However, I'm unsure if condition() is the right place to do it? Maybe condition() can return true to target functions from `UNSUPPORTED_HIVE_UDFS` and then transform() can throw the  exception when those functions are present.

coral-spark/src/main/java/com/linkedin/coral/spark/transformers/FallBackToHiveUDFTransformer.java
aastha25, line 32: this also looks LI specific atm. Please rename the class to reflect that.

coral-spark/src/main/java/com/linkedin/coral/spark/transformers/FallBackToHiveUDFTransformer.java
```diff
@@ -0,0 +1,78 @@
...
+          "com.linkedin.coral.hive.hive2rel.CoralTestUnsupportedUDF");
+  private final Set<SparkUDFInfo> sparkUDFInfos;
+
+  public FallBackToHiveUDFTransformer(Set<SparkUDFInfo> sparkUDFInfos) {
+    this.sparkUDFInfos = sparkUDFInfos;
+  }
+
+  @Override
+  protected boolean condition(SqlCall sqlCall) {
+    final String functionClassName = sqlCall.getOperator().getName();
```
aastha25, line 55: Across the whole PR, I have a general concern around variable name in format - *ClassName. Can we not assume that it'll be a className? 
For instance here, lets say - 
`String operatorName = sqlCall.getOperator().getName();` and further illustrate in the comments with an example of operatorName.

coral-spark/src/main/java/com/linkedin/coral/spark/transformers/FallBackToHiveUDFTransformer.java
```diff
@@ -0,0 +1,78 @@
...
+          // for unit test
+          "com.linkedin.coral.hive.hive2rel.CoralTestUnsupportedUDF");
+  private final Set<SparkUDFInfo> sparkUDFInfos;
+
+  public FallBackToHiveUDFTransformer(Set<SparkUDFInfo> sparkUDFInfos) {
+    this.sparkUDFInfos = sparkUDFInfos;
+  }
+
+  @Override
+  protected boolean condition(SqlCall sqlCall) {
```
aastha25, line 54: same comment here - 
add another condition for sqlCall.getOperator() instanceof  VersionedSqlUserDefinedFunction

coral-spark/src/main/java/com/linkedin/coral/spark/transformers/TransportableUDFTransformer.java
```diff
@@ -0,0 +1,97 @@
...
+import org.slf4j.LoggerFactory;
+
+import com.linkedin.coral.common.transformers.SqlCallTransformer;
+import com.linkedin.coral.hive.hive2rel.functions.VersionedSqlUserDefinedFunction;
+import com.linkedin.coral.spark.containers.SparkUDFInfo;
+import com.linkedin.coral.spark.exceptions.UnsupportedUDFException;
+
+
+/**
+ * This transformer transforms legacy Hive UDFs to an equivalent Transportable UDF.
```
aastha25, line 25: .... to an equivatent 'registered' transport function

coral-spark/src/main/java/com/linkedin/coral/spark/SparkSqlRewriter.java
```diff
@@ -65,6 +64,23 @@ public SqlNode visit(SqlCall call) {
...
+  private boolean containsSqlRowTypeSpec(SqlDataTypeSpec sqlDataTypeSpec) {
+    if (sqlDataTypeSpec instanceof SqlRowTypeSpec) {
+      return true;
+    }
+    if (sqlDataTypeSpec instanceof SqlArrayTypeSpec) {
+      return containsSqlRowTypeSpec(((SqlArrayTypeSpec) sqlDataTypeSpec).getElementTypeSpec());
+    }
+    if (sqlDataTypeSpec instanceof SqlMapTypeSpec) {
+      return containsSqlRowTypeSpec(((SqlMapTypeSpec) sqlDataTypeSpec).getKeyTypeSpec())
+          || containsSqlRowTypeSpec(((SqlMapTypeSpec) sqlDataTypeSpec).getValueTypeSpec());
```
aastha25, line 79: is there a UT to point out why we needed to expand the criteria for containsSqlRowTypeSpec()? 
Can there be a sqlCall - CAST(operand to Array<row>) ?

coral-spark/src/main/java/com/linkedin/coral/spark/SparkSqlRewriter.java
```diff
@@ -65,6 +64,23 @@ public SqlNode visit(SqlCall call) {
     return super.visit(call);
   }
 
+  /**
+   * Check if the SqlDataTypeSpec contains `ROW`
+   */
+  private boolean containsSqlRowTypeSpec(SqlDataTypeSpec sqlDataTypeSpec) {
+    if (sqlDataTypeSpec instanceof SqlRowTypeSpec) {
```
aastha25, line 71: use if-else if - else

coral-hive/src/main/java/com/linkedin/coral/transformers/CoralRelToSqlNodeConverter.java
aastha25: Please add a UT for this behavior override.

aastha25 (commented):
Thanks for the PR @ljfgem

coral-hive/src/main/java/com/linkedin/coral/transformers/CoralRelToSqlNodeConverter.java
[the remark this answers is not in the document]
ljfgem: (1) Didn't add `ROW` operator because Coral is not using this kind of operator, but I added it back to be aligned with the upstream
(2) The current upstream impl is still converting the whole RexFieldAccess to a SqlIdentifier, so backporting that doesn't meet our requirement, we still needs to override this method.

coral-hive/src/main/java/com/linkedin/coral/transformers/CoralRelToSqlNodeConverter.java
```diff
@@ -345,4 +348,43 @@ private SqlNode generateRightChildForSqlJoinWithLateralViews(BiRel e, Result rig
 
     return SqlStdOperatorTable.AS.createCall(POS, asOperands);
   }
+
+  /**
+   * Override this method to handle the conversion for RelNode `f(x).y.z` where `f` is an operator, which
```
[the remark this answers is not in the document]
ljfgem: (1) Actually `f(x).y.z` is a `RexFieldAccess`, which is represented as `f(x).y.z`, modified `RelNode` to be `RexFieldAccess`.
(2) No, we need it for all the field access to a function call, for `f(x).y`, we still need to override it to ensure `f(x)` to be converted to a `SqlCall`

coral-spark/src/main/java/com/linkedin/coral/spark/CoralToSparkSqlCallConverter.java
```diff
@@ -0,0 +1,160 @@
...
+
+import static com.linkedin.coral.spark.transformers.TransportableUDFTransformer.*;
+
+
+/**
+ * This class extends the class of {@link org.apache.calcite.sql.util.SqlShuttle} and initialize a {@link com.linkedin.coral.common.transformers.SqlCallTransformers}
+ * which containing a list of {@link com.linkedin.coral.common.transformers.SqlCallTransformer} to traverse the hierarchy of a {@link org.apache.calcite.sql.SqlCall}
+ * and converts the functions from Coral operator to Spark operator if it is required
+ */
+public class CoralToSparkSqlCallConverter extends SqlShuttle {
```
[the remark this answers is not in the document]
ljfgem: I think it's better to have 2 separate class, given `CoralToSparkSqlCallConverter` contains lots of registrations.
At present, as mentioned in the class comment, `CoralToSparkSqlCallConverter` only contains the transformations defined with `SqlCallTransformer`, while `CoralSqlNodeToSparkSqlNodeConverter` is not that case. If we consider migrating `CoralSqlNodeToSparkSqlNodeConverter` to use `SqlCallTransformer` only in the future, we can merge them.

And added the note about transformer ordering.

coral-hive/src/main/java/com/linkedin/coral/transformers/CoralRelToSqlNodeConverter.java
```diff
@@ -345,4 +348,43 @@ private SqlNode generateRightChildForSqlJoinWithLateralViews(BiRel e, Result rig
...
+          // and store the accessed field names ([`z`, `y`] in the example above, needs to be reversed)
+          while (referencedExpr.getKind() == SqlKind.FIELD_ACCESS) {
+            accessNames.add(((RexFieldAccess) referencedExpr).getField().getName());
+            referencedExpr = ((RexFieldAccess) referencedExpr).getReferenceExpr();
+          }
+          if (referencedExpr.getKind() == SqlKind.OTHER_FUNCTION || referencedExpr.getKind() == SqlKind.CAST) {
+            SqlNode functionCall = toSql(program, referencedExpr);
+            Collections.reverse(accessNames);
+            for (String accessName : accessNames) {
+              functionCall = FunctionFieldReferenceOperator.DOT.createCall(SqlParserPos.ZERO, functionCall,
```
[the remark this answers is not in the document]
ljfgem: Yes, corresponding `coralSqlNode1` is using the same `FunctionFieldReferenceOperator.DOT` instead of a SqlIdentifier `.`.

coral-hive/src/main/java/com/linkedin/coral/transformers/CoralRelToSqlNodeConverter.java
```diff
@@ -345,4 +348,43 @@ private SqlNode generateRightChildForSqlJoinWithLateralViews(BiRel e, Result rig
...
+          final List<String> accessNames = new ArrayList<>();
+          RexNode referencedExpr = rex;
+          // Use the loop to get the top-level struct (`f(x)` in the example above),
+          // and store the accessed field names ([`z`, `y`] in the example above, needs to be reversed)
+          while (referencedExpr.getKind() == SqlKind.FIELD_ACCESS) {
+            accessNames.add(((RexFieldAccess) referencedExpr).getField().getName());
+            referencedExpr = ((RexFieldAccess) referencedExpr).getReferenceExpr();
+          }
+          if (referencedExpr.getKind() == SqlKind.OTHER_FUNCTION || referencedExpr.getKind() == SqlKind.CAST) {
+            SqlNode functionCall = toSql(program, referencedExpr);
```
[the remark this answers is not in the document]
ljfgem: Actually the override doesn't applies to `referencedExpr` conversion, given `referencedExpr` is already the function call, not the field access.

coral-spark/src/main/java/com/linkedin/coral/spark/transformers/FallBackToHiveUDFTransformer.java
ljfgem: I changed it to be `FallBackToLinkedInHiveUDFTransformer`, keep using `HiveUDF` because it only falls back the Hive specific UDFs (starting with `com.linkedin.`).

coral-spark/src/main/java/com/linkedin/coral/spark/transformers/TransportableUDFTransformer.java
ljfgem: I don't think that is necessary, given there is no other kind of transport UDFs, we can add them if needed in the future.

ljfgem:
@aastha25 Thanks for your review! I have addressed all the comments and replied.
@wmoustafa PTAL, thanks!

wmoustafa (approved):

PR #352, opened 2023-02-07
Coral-Service: Handle error message for local mode creation failure

ljfgem:
Previously, Coral-Service doesn't return the error message for local mode creation failure, and the UI is like:
Screen Shot 2023-02-07 at 1 52 17 PM

With this PR, the error message can be displayed properly:
Screen Shot 2023-02-07 at 1 51 40 PM


ljfgem:
@aastha25 FYI for the demo.

aastha25:
Thanks @ljfgem  for the PR! LGTM

PR #360, opened 2023-02-23
Coral-Common: Return 'string' for 'char' type in 'RelDataTypeToHiveTypeStringConverter'

ljfgem:
`char` is not a valid Hive type string, it should be mapped to `string` as the Hive type according to the schema of the corresponding production views.

Tests:
1. New UT
2. Regression test on prod views
3. Tested on the affected prod view with gateway Spark shell

### DIFF
--- a/coral-common/src/main/java/com/linkedin/coral/common/utils/RelDataTypeToHiveTypeStringConverter.java
+++ b/coral-common/src/main/java/com/linkedin/coral/common/utils/RelDataTypeToHiveTypeStringConverter.java
@@ -69,7 +69,7 @@ public class RelDataTypeToHiveTypeStringConverter {
       case BOOLEAN:
         return "boolean";
       case CHAR:
-        return "char";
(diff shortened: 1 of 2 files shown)


yiqiangin:
I remember Char is a valid type in Hive referring to https://cwiki.apache.org/confluence/display/hive/languagemanual+types#LanguageManualTypes-CharcharChar
Do you convert it to `string` because we cannot determine the length? How about converting it to `char(255)` as the maximum length is of char type in Hive is 255.

ljfgem:
> I remember Char is a valid type in Hive referring to https://cwiki.apache.org/confluence/display/hive/languagemanual+types#LanguageManualTypes-CharcharChar Do you convert it to `string` because we cannot determine the length? How about converting it to `char(255)` as the maximum length is of char type in Hive is 255.

Good point, I have considered that option. But given the Hive schemas of the corresponding views are all `string` rather than `char(x)`, I chose to return `string`.

yiqiangin (approved):

PR #361, opened 2023-02-24
Coral-Service: Enable the location of hive.properties to be configurable

nagarathnam200:
## Summary ##
This patch enables the location of "hive.properties" to be configurable. By default it falls back to the hive.properties file in the resources folder. This will be helpful when the service is deployed in a hadoop cluster and we need to rely on a cluster specifc config deployed in the containers to boot up the service.

## Testing done ##
./gradlew  clean build with all the unit tests passing

Verified that ./gradlew bootRun picks up the hive.properties in resources folder.

Verified that the ./gradlew bootRun --args='--hivePropsLocation=/tmp/hive.properties'  picks up the hiveProps from the location.

### DIFF
--- a/coral-service/src/main/java/com/linkedin/coral/coralservice/controller/TranslationController.java
+++ b/coral-service/src/main/java/com/linkedin/coral/coralservice/controller/TranslationController.java
@@ -31,8 +31,8 @@ import static com.linkedin.coral.coralservice.utils.TranslationUtils.*;
 @Service
 @Profile({ "remoteMetastore", "default" })
 public class TranslationController implements ApplicationListener<ContextRefreshedEvent> {
-  @Value("${hiveProps:}")
-  private String hiveProps;
+  @Value("${hivePropsLocation:}")
+  private String hivePropsLocation;
 
   private final static ImmutableMap<String, String> LANGUAGE_MAP =
       ImmutableMap.of("hive", "Hive QL", "trino", "Trino SQL", "spark", "Spark SQL");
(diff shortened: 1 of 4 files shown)


coral-service/src/main/java/com/linkedin/coral/coralservice/controller/TranslationController.java
```diff
@@ -30,14 +31,17 @@
 @Service
 @Profile({ "remoteMetastore", "default" })
 public class TranslationController implements ApplicationListener<ContextRefreshedEvent> {
+  @Value("${hiveProps:}")
+  private String hiveProps;
```
ljfgem, line 35: Let's modify the name to something like `hivePropsLocation`?

ljfgem (commented):
Thanks @nagarathnam200 for this PR!
Left a nit, and could you also modify the [corresponding part in README](https://github.com/linkedin/coral#to-run-coral-service-using-the-remote-metastore)?

coral-service/src/main/java/com/linkedin/coral/coralservice/controller/TranslationController.java
nagarathnam200: Ack. Done.

nagarathnam200:
> Thanks @nagarathnam200 for this PR! Left a nit, and could you also modify the [corresponding part in README](https://github.com/linkedin/coral#to-run-coral-service-using-the-remote-metastore)?

Thanks! Updated the section.

ljfgem (approved):
LGTM

PR #364, opened 2023-02-27
Coral-Hive: Avoid dropping 'CAST' on String

ljfgem:
## Change summary
Previously, the default value of [requireCastOnString](https://github.com/linkedin/linkedin-calcite/blob/c13ba44c6974574d07a47766e32e695e5c5e042f/core/src/main/java/org/apache/calcite/sql/SqlDialect.java#L612) is `false`, then Coral will drop `CAST` for RexNode like `CAST(number_string AS BIGINT) > 0`, which might cause translation quality issue.
For example, without explicit `CAST`, Spark will cast the `number_string` to `int` implicitly, but if the value of the `number_string` is greater than `Integer.MAX_VALUE`, the result of the condition is wrong.

## Tests
1. New UT
2. Regression test on production views
3. Test on the affected production views

### DIFF
--- a/coral-hive/src/main/java/com/linkedin/coral/transformers/CoralRelToSqlNodeConverter.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/transformers/CoralRelToSqlNodeConverter.java
@@ -81,7 +81,7 @@ public class CoralRelToSqlNodeConverter extends RelToSqlConverter {
         /**
          * The default value is `false`, then Coral will drop `CAST` for RexNode like `CAST(number_string AS BIGINT) > 0`,
          * which might cause translation quality issue.
-         * For example, without explict `CAST`, Spark will cast the `number_string` to {@link Integer} implicitly,
+         * For example, without explicit `CAST`, Spark will cast the `number_string` to {@link Integer} implicitly,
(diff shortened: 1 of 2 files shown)


yiqiangin (approved):

coral-hive/src/main/java/com/linkedin/coral/transformers/CoralRelToSqlNodeConverter.java
```diff
@@ -75,7 +75,20 @@ private static SqlDialect returnInstance() {
...
+      @Override
+      public boolean requireCastOnString() {
+        /**
+         * The default value is `false`, then Coral will drop `CAST` for RexNode like `CAST(number_string AS BIGINT) > 0`,
+         * which might cause translation quality issue.
+         * For example, without explicit `CAST`, Spark will cast the `number_string` to {@link Integer} implicitly,
+         * but if the value of the number_string is greater than {@link Integer.MAX_VALUE}, the result of the condition is wrong.
+         *
+         * Check `CoralSparkTest#testCastOnString` for an example.
+         */
```
wmoustafa, line 88: Could you update the comment to state this preserves existing cast and does not introduce new one?

ljfgem: Thanks for suggestion, done.

PR #369, opened 2023-03-17
Coral-Schema: Return nullable schema for inner fields created by operator

ljfgem:
## Change Summary
At present, we are using `RelDataType.isNullable()` to determine if an inner field should be nullable or not while converting `RelDataType` to Avro schema, but Spark expects all the inner fields created by operator to be nullable. This PR is to make Coral-Schema return nullable schema for inner fields created by operator to be aligned with Spark's expectation.

## Tests
1. Modified UTs, these changed fields are created by operator, they are nullable now with this PR
2. Regression test on all the prod views, no translation regression
3. Tested on the affected prod views, can be queried on Spark shell with this PR

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
@@ -6,7 +6,6 @@
 package com.linkedin.coral.schema.avro;
 
 import java.util.ArrayList;
-import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


yiqiangin (approved):
LGTM. It looks like this change is done for Spark. Pls make sure it does work for other query engines as well.

ljfgem:
> LGTM. It looks like this change is done for Spark. Pls make sure it does work for other query engines as well.

Thanks for review! Coral-Schema is only used for view queries on Spark at present.

PR #373, opened 2023-03-23
Coral-Trino: Remove 'Calcite2TrinoUDFConverter#adjustInconsistentTypesToEqualityOperator' and use 'CAST' Instead of 'TRY_CAST'

ljfgem:
## Summary
This PR removes the current usage of [adjustInconsistentTypesToEqualityOperator](https://github.com/linkedin/coral/blob/103092462dcb8ec453f08af08d7e33707128e993/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java#L426), which adds `TRY_CAST` for type mismatches within relational operators. For instance, `'1' = 1` in Hive SQL is converted to `TRY_CAST('1' AS INTEGER) = 1` in Trino SQL due to Trino's requirement for explicit casting.

However, when there is a type mismatch in relational operations like `'1' = 1`, Calcite automatically adds an explicit `CAST` in Coral IR while converting the SqlNode to Coral IR. As a result, the corresponding converted RexCall is 
`=(CAST('1'):INTEGER NOT NULL, 1)` and the respective part in Trino SQL becomes `CAST('1' AS INTEGER) = 1`. This makes the addition of `TRY_CAST` unnecessary.

By removing [adjustInconsistentTypesToEqualityOperator](https://github.com/linkedin/coral/blob/103092462dcb8ec453f08af08d7e33707128e993/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/Calcite2TrinoUDFConverter.java#L426), the only alteration is that the original `TRY_CAST` is replaced by `CAST`, as shown in the unit test changes of this PR. This is acceptable because:
- If `CAST` succeeds, then `CAST` is equivalent to `TRY_CAST`.
- If `CAST` fails, a query using `CAST` fails, while a query using `TRY_CAST` succeeds. However, since `TRY_CAST` returns `null` in this case, it may lead to query result data quality issues, which is undesirable.

## Tests
1. Updated unit tests
2. Regression tests on production views with no translation regression
3. Local Trino server regression tests on the production views with SQL changes, confirming that they can be queried successfully with this PR


yiqiangin (approved):
LGTM

wmoustafa:
As discussed in #377 starting from [this comment](https://github.com/linkedin/coral/pull/377#discussion_r1151168322), removing this transformation and relying on SqlNode to RelNode conversion to perform the type matching compromises Coral SqlNode IR. We might have to revert that change introduced by this PR and keep the transformation on the Trino side.

ljfgem:
> As discussed in #377 starting from [this comment](https://github.com/linkedin/coral/pull/377#discussion_r1151168322), removing the Trino-side transformation and relying on SqlNode to RelNode conversion to perform the type matching compromises Coral SqlNode IR. We might have to revert the change introduced by this PR and keep the transformation on the Trino side.

@wmoustafa I don't think it makes sense to revert this commit because:
1. This transformation is on RelNode layer, which should be removed anyway
2. The existing `adjustInconsistentTypesToEqualityOperator` transformation introduces `TRY_CAST` operator in CoralSqlNode2 which is not even a valid Coral operator (it can't be validated by HiveSqlValidator), so reverting this PR will cause a bigger divergence between CoralSqlNode1 and CoralSqlNode2

To address your concern, I think we can open a follow-up PR on HiveSqlNode -> CoralSqlNode1 to add explicit `CAST` if there is a type mismatch so that we don't rely on Calcite to do it for us.

PR #374, opened 2023-03-24
Add PR template

ljfgem:

This PR introduces a PR template to ensure future PRs are well-structured.

The PR template is currently used in this PR.
cc: @wmoustafa


ljfgem:
@wmoustafa Does this template look good to you?

wmoustafa:
`Does this PR introduce any user-facing change?` This can mean many things. For example, I do not view this PR as introducing user facing change since the APIs are backward compatible. Given that some of Coral APIs are public but still not meant for user consumption, we might be careful about how to phrase that question or remove it for now. In the long term, we should annotate our methods to be `dev-facing` if they are public but only for internal Coral use. FYI @aastha25.

ljfgem:
> `Does this PR introduce any user-facing change?` This can mean many things. For example, I do not view this PR as introducing user facing change since the APIs are backward compatible. Given that some of Coral APIs are public but still not meant for user consumption, we might be careful about how to phrase that question or remove it for now. In the long term, we should annotate our methods to be `dev-facing` if they are public but only for internal Coral use. FYI @aastha25.

@wmoustafa Thanks for your review, removed it.

wmoustafa:
Did you want to update the description to remove it as well?

ljfgem:
> Did you want to update the description to remove it as well?

Thanks for reminding, modified the description of this PR and my other pending PRs using this template.

.github/pull_request_template.md
```diff
@@ -0,0 +1,26 @@
+<!--
+Thank you for submitting a pull request! Here are some suggestions to help you:
+  1. If you're new to this, kindly review our contributor guidelines at https://github.com/linkedin/coral/blob/master/CONTRIBUTING.md.
+  2. Make sure you've added or executed the relevant tests for your PR.
```
wmoustafa, line 4: you have
added and executed

wmoustafa (approved):

PR #376, opened 2023-03-27
Create coral-incremental module and integrate with coral-dbt

AliceYeh12:
**Summary**
Introduce `coral-incremental`, which houses the incremental maintenance logic for the Coral incremental rewrite endpoint, which is used by `coral-dbt`.

**Changes/Additions**

- Create `coral-incremental` module
- Update Coral service logic to utilize coral-incremental rewrite functionality
- Modify `coral-dbt` code generation logic

**Testing**
Unit tests were written for the new coral-incremental logic (`RelToIncrementalSqlConverterTest.java`) and the endpoint and dbt integration were tested by running a local dbt project.


coral-dbt/src/main/resources/macros/coral_macros/spark/materializations/incremental_maintenance.sql
```diff
@@ -39,11 +43,17 @@
         ~ 'val start_snapshot_id = snap_ids.collect()(1)(0).toString\n'
         ~ 'val end_snapshot_id = snap_ids.collect()(0)(0).toString\n'
     %}
-    {% set create_df_code =
+--     Time travel to get original table before additions
+    {% set create_original_df_code =
```
wmoustafa, line 47: Should variable names be consistent (if they start with verb or not)?

coral-incremental/build.gradle
```diff
@@ -0,0 +1,14 @@
+dependencies {
+  compile project(':coral-hive')
+
+  testCompile(deps.'hive'.'hive-exec-core') {
+    exclude group: 'org.apache.avro', module: 'avro-tools'
+    // These exclusions are required to prevent duplicate classes since we include
+    // shaded jar above
+    exclude group: 'org.apache.calcite', module: 'calcite-core'
+    exclude group: 'org.apache.calcite', module: 'calcite-avatica'
```
wmoustafa, line 9: @ljfgem could you take a look?

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelIncrementalTransformer.java
```diff
@@ -0,0 +1,111 @@
...
+import org.apache.calcite.rel.logical.LogicalFilter;
+import org.apache.calcite.rel.logical.LogicalJoin;
+import org.apache.calcite.rel.logical.LogicalProject;
+import org.apache.calcite.rel.logical.LogicalTableScan;
+import org.apache.calcite.rel.logical.LogicalUnion;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexNode;
+
+
+public class RelIncrementalTransformer {
```
wmoustafa, line 30: `RelNodeIncrementalTransfomer` for consistency with the visualizer (interesting co-incidence).

wmoustafa: Can this class extend the `RelShuttleImpl`?

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelIncrementalTransformer.java
```diff
@@ -0,0 +1,111 @@
...
+    ArrayList<RexNode> projects = new ArrayList<>();
+    ArrayList<String> names = new ArrayList<>();
+    IntStream.range(0, incrementalJoin.getRowType().getFieldList().size()).forEach(i -> {
+      projects.add(rexBuilder.makeInputRef(incrementalJoin, i));
+      names.add(incrementalJoin.getRowType().getFieldNames().get(i));
+    });
+    return LogicalProject.create(incrementalJoin, projects, names);
+  }
+
+  private static RelNode transformToIncremental(RelNode target) {
```
wmoustafa, line 82: When we discussed this, we assumed the simple case where it goes up to the leaf and converts it to `table_delta`. I see you have implemented a more complex use case where it recursively calls `convertRelIncremental`. A few questions:
1- Have we verified the rewrites are correct? Let us write a short doc to illustrate with an example, or may be in the comments.
2- Since it is recursive with `convertRelIncremental`, can we move the logic to the visitors above?

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelIncrementalTransformer.java
AliceYeh12: Hmm, I'm not sure that it makes sense for the entire transformer class to be extending `RelShuttleImpl`. For example, something like `IRRelToSparkRelTransformer`, which the format of this class was based off of, does not extend `RelShuttleImpl`.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelIncrementalTransformer.java
AliceYeh12: 1. Yes, `RelToIncrementalSqlConverterTest` has a number of tests (with inputs and expected outputs) to verify the correctness of the rewrites. I'm happy to throw these into a doc as well if that would be more digestible.

2. Good point, will refactor this.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeIncrementalTransformer.java
```diff
@@ -0,0 +1,109 @@
...
+        return LogicalAggregate.create(transformedChild, aggregate.getGroupSet(), aggregate.getGroupSets(),
+            aggregate.getAggCallList());
+      }
+    };
+    return originalNode.accept(converter);
+  }
+
+  private static LogicalProject createProjectOverJoin(LogicalJoin join, RelNode left, RelNode right,
+      RexBuilder rexBuilder) {
+    LogicalJoin incrementalJoin =
```
wmoustafa, line 98: This function is not specific to incremental, right? Might consider changing the variable name.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeIncrementalTransformer.java
```diff
@@ -0,0 +1,109 @@
...
+        LogicalJoin.create(left, right, join.getCondition(), join.getVariablesSet(), join.getJoinType());
+    ArrayList<RexNode> projects = new ArrayList<>();
+    ArrayList<String> names = new ArrayList<>();
+    IntStream.range(0, incrementalJoin.getRowType().getFieldList().size()).forEach(i -> {
+      projects.add(rexBuilder.makeInputRef(incrementalJoin, i));
+      names.add(incrementalJoin.getRowType().getFieldNames().get(i));
+    });
+    return LogicalProject.create(incrementalJoin, projects, names);
```
wmoustafa, line 106: I recall we discussed creating the projection was not necessary?

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelToIncrementalSqlConverter.java
```diff
@@ -0,0 +1,23 @@
...
+
+
+public class RelToIncrementalSqlConverter {
+
+  public String convert(RelNode relNode) {
+    RelNode incrementalRelNode = RelNodeIncrementalTransformer.convertRelIncremental(relNode);
+    CoralRelToSqlNodeConverter converter = new CoralRelToSqlNodeConverter();
+    SqlNode sqlNode = converter.convert(incrementalRelNode);
+    return sqlNode.toSqlString(converter.INSTANCE).getSql();
+  }
```
wmoustafa, line 21: Can we move this to the test cases? Want to avoid defining public APIs yet.

coral-service/src/main/java/com/linkedin/coral/coralservice/utils/IncrementalUtils.java
```diff
@@ -0,0 +1,24 @@
...
+import com.linkedin.coral.incremental.RelToIncrementalSqlConverter;
+
+import static com.linkedin.coral.coralservice.utils.CoralProvider.*;
+
+
+public class IncrementalUtils {
+
+  public static String getIncrementalQueryFromUserSql(String query) {
+    RelNode originalNode = new HiveToRelConverter(hiveMetastoreClient).convertSql(query);
+    String incrementalQuery = new RelToIncrementalSqlConverter().convert(originalNode);
```
wmoustafa, line 20: If we move `RelToIncrementalSqlConverter` to the test cases, I think we have similar infra in `coral-service` already, right? We can use it.

coral-incremental/build.gradle
ljfgem: I think 
```
    exclude group: 'org.apache.avro', module: 'avro-tools'
    exclude group: 'org.apache.calcite', module: 'calcite-avatica'
```
are not necessary, and please modify the comment to be:
```
    // This exclusion is required to prevent duplicate classes since we include
    // shaded 'calcite-core' transitively
```
Other parts LGTM.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeIncrementalTransformer.java
AliceYeh12: It is intended to be used just within the incremental transformer.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeIncrementalTransformer.java
AliceYeh12: Looks like returning a LogicalJoin works correctly on the SQL level (passes the tests) but generates a RelNode representation that is different than what Coral would generate going from the same SQL -> RelNode. I think it might be better to create the projection to maintain consistency on the RelNode level.

coral-service/src/main/java/com/linkedin/coral/coralservice/utils/IncrementalUtils.java
```diff
@@ -0,0 +1,26 @@
...
+
+
+public class IncrementalUtils {
+
+  public static String getIncrementalQueryFromUserSql(String query) {
+    RelNode originalNode = new HiveToRelConverter(hiveMetastoreClient).convertSql(query);
+    RelNode incrementalRelNode = RelNodeIncrementalTransformer.convertRelIncremental(originalNode);
+    CoralSpark coralSpark = CoralSpark.create(incrementalRelNode);
+    return coralSpark.getSparkSql();
+  }
```
wmoustafa, line 24: Should this take a dialect as input? Right now it is assumed to be Spark, but what if we want to rewrite a Trino query in the future? This should apply all the way to the callers of this API on the client side.

wmoustafa: Let us unify the Dialect representation with whatever being used in translation utils. Consider creating an ENUM for dialects?

AliceYeh12: Looks like there is a defined mapping of supported dialects in the translation controller. Thinking of pulling from that for this endpoint.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeIncrementalTransformer.java
```diff
@@ -0,0 +1,109 @@
...
+import org.apache.calcite.rel.logical.LogicalFilter;
+import org.apache.calcite.rel.logical.LogicalJoin;
+import org.apache.calcite.rel.logical.LogicalProject;
+import org.apache.calcite.rel.logical.LogicalTableScan;
+import org.apache.calcite.rel.logical.LogicalUnion;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexNode;
+
+
+public class RelNodeIncrementalTransformer {
```
aastha25, line 30: Please add java doc for public classes.

aastha25: lets take documentation offline. We might want to first converge on what to capture as part of the java doc.

aastha25 (approved):
Great work, Alice :)

Issue #357, opened 2023-02-14
Avoid converting `named_struct` to `CAST(ROW AS RowType)` in `CoralSqlNode2 -> RelNode` conversion


ljfgem:
[Related discussion](https://github.com/linkedin/coral/pull/354#discussion_r1105032525)

In the translation path:
```
source SQL -> SourceLangSqlNode -> CoralSqlNode1 -> CoralRelNode -> CoralSqlNode2 -> TargetLangSqlNode -> target SQL
```
At present, for `named_struct` function in CoralSqlNode1, Calcite converts it to `CAST(ROW AS RowType)` in CoralRelNode, so in CoralSqlNode2, it's `CAST(ROW AS RowType)`, which is different from `named_struct` in CoralSqlNode1. Note that only `named_struct` is a valid Coral IR. 

So we should make the following changes:
1. Avoid converting `named_struct` to `CAST(ROW AS RowType)` in `CoralSqlNode2 -> RelNode` conversion, and remove [CastToNamedStructTransformer](https://github.com/linkedin/coral/blob/master/coral-spark/src/main/java/com/linkedin/coral/spark/transformers/CastToNamedStructTransformer.java)
2. Add a transformer in coral-trino to convert `named_struct` to `CAST(ROW AS RowType)`.

PR #377, opened 2023-03-27
[Coral-Spark] Remove unnecessary 'removeCastToEnsureCorrectNullability' transformation

ljfgem:

Previously we used `removeCastToEnsureCorrectNullability` to remove some `CAST`, because the schema for inner fields created by operator might be non-nullable while Spark expects all the inner fields created by operator to be nullable. However, since we have fixed nullability issue in Coral-Schema with #369 to return nullable schema for inner fields created by operator, we don't need to drop these `CAST` on purpose.

Example (thanks @wmoustafa for summarizing):
**Query:**
select cast(column as array(string)) as col2
**Schemas:**
calcite: col2: nullable array of non-nullable string
coral before #369: col2: nullable array of non-nullable string
coral after #369: col2: nullable array of nullable string
spark: col2: nullable array of nullable string
**Query:**
select column
No problem due to lack of operator around column. Everything is straight forward.
**Conclusion:**
The cast() example happens with any operator, not just cast. The `removeCastToEnsureCorrectNullability` fix is narrow-scoped solution that addresses the problem just for cast since it is easy to remove cast sometimes, but it is not a good thing to do, since it could be coming from the source language.

1. Regression test on all the production views, no translation failure
2. For all the views with different translated SQL, tested them with regression test on Spark Gateway, the new SQLs don't cause query  regression. 

SQL diff samples (format: `Spark SQL before this PR -> Spark SQL after this PR: comment`):
- `SELECT NULL` -> `SELECT CAST(NULL AS STRING)`: Original HQL contains `CAST`, was removed by `removeCastToEnsureCorrectNullability`
- `SELECT udf(col)` -> `SELECT CAST(udf(col) as BIGINT)`: Original HQL contains `CAST`, was removed by `removeCastToEnsureCorrectNullability`
- `SELECT CASE WHEN TRUE THEN NULL ELSE split('a b', ' ') END AS col` -> `SELECT CASE WHEN TRUE THEN CAST(NULL AS ARRAY<VARCHAR>) ELSE CAST(split('a b', ' ') AS ARRAY<VARCHAR>) END AS col`: Original HQL doesn't contain `CAST` (because HQL allows implicit CAST), Calcite adds explicit `CAST` while converting from CoralSqlNode1 to RelNode, this diff will not fail the query because after #369, both Spark and Coral-Schema generates nullable array of nullable string for `col`.


yiqiangin (approved):

coral-spark/src/test/java/com/linkedin/coral/spark/CoralSparkTest.java
```diff
@@ -829,16 +829,6 @@ public void testSelectStar() {
     assertEquals(expandedSql, targetSql);
   }
 
-  @Test
-  public void testRedundantCastRemovedFromCaseCall() {
-    final String sourceSql = "SELECT CASE WHEN TRUE THEN NULL ELSE split(b, ' ') END AS col1 FROM complex";
-    String expandedSql = getCoralSparkTranslatedSqlWithAliasFromCoralSchema(sourceSql);
-
-    String targetSql =
-        "SELECT CASE WHEN TRUE THEN NULL ELSE split(complex.b, ' ') END col1\n" + "FROM default.complex complex";
```
aastha25, line 838: why was this UT deleted?

ljfgem: This UT was added together with the transformation `removeCastToEnsureCorrectNullability`. After we remove it, the target SQL will be:
```
SELECT CASE WHEN TRUE THEN CAST(NULL AS ARRAY<STRING>) ELSE CAST(split(complex.b, ' ') AS ARRAY<STRING>) END col1
FROM default.complex complex
```
Because Calcite adds the explicit `CAST` to ensure the types match for the field `col1` in `CoralSqlNode1 -> Coral IR` conversion. The newly added `CAST` will not cause regression after we fix the Coral-Schema nullability issue in #369.

aastha25: not sure im following the description. 
For input SQL - "SELECT CASE WHEN TRUE THEN NULL ELSE split(b, ' ') END AS col1 FROM complex"
could you share - 
hiveSqlNode, coralSqlNode1, coralSqlNode2 representation?

When you say `The newly added CAST`, is it the same CAST(s) added on LHS or a different one?
Also, seems like we can retain this unit test

ljfgem: HiveSqlNode & CoralSqlNode1:
```
SELECT CASE WHEN TRUE THEN NULL ELSE `split`(`b`, ' ') END AS `col1`
FROM `complex`
```
CoralSqlNode2:
```
SELECT CASE WHEN TRUE THEN CAST(NULL AS ARRAY<VARCHAR(65535) CHARACTER SET `ISO-8859-1`>) ELSE CAST(`split`(`complex`.`b`, ' ') AS ARRAY<VARCHAR(65535) CHARACTER SET `ISO-8859-1`>) END AS `col1`
FROM `default`.`complex` AS `complex`
```
`The newly added CAST` is the one added in `CoralSqlNode2` above by Calcite while converting `CoralSqlNode1` to `Coral IR`, so it exists in `CoralSqlNode2`.
I don't think we need to retain this unit test because there is no corresponding implementation in Coral after this PR.

aastha25: there are two CAST operators in the CoralSqlNode2 representation. One around `NULL` and the other around `split`(`b`, ' ')`.
Are both added in CoralRelNode -> CoralSqlNode2 layer due to the override `requireCastOnString()` in https://github.com/linkedin/coral/blob/master/coral-hive/src/main/java/com/linkedin/coral/transformers/CoralRelToSqlNodeConverter.java#L81?

ljfgem: No, both are added while converting `CoralSqlNode1` to `Coral IR` in `HiveSqlToRelConverter`. `requireCastOnString` only ensures the `CAST` in `CAST(string AS type)` won't be dropped incorrectly.

aastha25: okay, so it exists in the RelNode representation? (1) How is this information captured in RelNode?
(2) Also, to avoid introducing disparity between CoralSqlNode1, 2, is it possible to add this change in CoralSqlNode1 as well?

wmoustafa: Thinking a bit, we might have a deeper issue here. What is the answer to this question: Does Coral SQL Node spec expect operand types to match or is it optional? If it expects them to match, then we need CoralSqlNode1 to have the `CAST`. If it is optional, then Trino still needs to transform to ensure they match. I think we have discussed this in the past.

ljfgem: I think it's optional as Calcite is able to add explicit `CAST` if there is a type mismatch.
As for Trino case, it's because Trino SQL dialect requires it, it doesn't imply that CoralSqlNode must mandate it.

wmoustafa: Not clear if this answers my question. Are you saying that it is optional in Coral IR spec?

aastha25: I would advocate for CoralSqlNode to be as detailed as possible. This will enrich the SqlNode representation and possibly help with sqlNode transformations across different dialects.

ljfgem: Yes, it's optional in coral IR spec because we can handle both during the conversion from SqlNode to RelNode.

wmoustafa: If it is optional in Coral SqlNode IR, then Trino must have a defense mechanism to re-instate it, right? Doing it in Sql-to-Rel is not an answer because that does not speak to the "Coral IR SqlNode spec".

ljfgem: I think Trino doesn't need a defense mechanism, because explicit `CAST` is mandatary in Coral IR RelNode, it will definitely appear in the CoralSqlNode2 and the result Trino SQL.

wmoustafa: As mentioned, handing it in Coral RelNode IR is not the answer, because this compromises the SqlNode spec. In fact, with Coral SqlNode IR, going through the RelNode is optional.

ljfgem: I see, thanks for explaining. Then I think we need to make types match mandatory in `Coral IR SqlNode spec`. I think we need to add this case to #357 as a follow-up work to make types match in CoralSqlNode1.
Besides, if we want to make `100% CoralSqlNode1 == CoralSqlNode2` for Coral IR SqlNode spec, we may also need to run a comparison test for all the production views to find all the cases where `CoralSqlNode1 != CoralSqlNode2` and fix them.

ljfgem:
@wmoustafa @aastha25 I have added the details to the `test done` part, PTAL.

coral-spark/src/test/java/com/linkedin/coral/spark/CoralSparkTest.java
wmoustafa: After offline discussion, there might be a miscommunication in this thread. My comments where mostly pertaining to changes like #373. That said, I think changes in this PR (#377) are fine, and will follow up on this thread in #373.

PR #383, opened 2023-04-04
[Coral-Trino] Migrate 'rlike' and 'regexp' operations from RelNode to SqlNode layer

ljfgem:

This PR migrates `rlike` and `regexp` operations from RelNode to SqlNode layer. 
Besides, it modified those `SqlKind` from `LIKE` to `OTHER_FUNCTION`, because if we keep using `SqlKind.LIKE`, Calcite will convert `NOT RLIKE` to `NOT LIKE` during `Coral RelNode -> CoralSqlNode2` conversion [here](https://github.com/linkedin/linkedin-calcite/blob/li-1.21.0/core/src/main/java/org/apache/calcite/rel/rel2sql/SqlImplementor.java#L653), then Hive SQL `SELECT '1' NOT RLIKE '^1$'` will be translated to `SELECT '1' NOT LIKE '^1$'`, which is not correct.

1. New UTs
2. Regression test on prod views, no translation failures, no result differences.

### DIFF
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveRLikeOperator.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveRLikeOperator.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2018-2021 LinkedIn Corporation. All rights reserved.
+ * Copyright 2018-2023 LinkedIn Corporation. All rights reserved.
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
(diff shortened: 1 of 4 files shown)


yiqiangin (approved):

wmoustafa:
Why does the `LIKE` bug manifest only in `NOT RLIKE` and not in `RLIKE` (without `NOT`)?

wmoustafa (approved):

ljfgem:
> Why does the `LIKE` bug manifest only in `NOT RLIKE` and not in `RLIKE` (without `NOT`)?

@wmoustafa Because Calcite only converts for `NOT LIKE` here:
```
      case NOT:
        RexNode operand = ((RexCall) rex).operands.get(0);
        final SqlNode node = toSql(program, operand);
        switch (operand.getKind()) {
        case IN:
          return SqlStdOperatorTable.NOT_IN
              .createCall(POS, ((SqlCall) node).getOperandList());
        case LIKE:
          return SqlStdOperatorTable.NOT_LIKE
              .createCall(POS, ((SqlCall) node).getOperandList());
```
Coral shouldn't use `SqlKind.LIKE` because `rlike/regexp != like`

PR #384, opened 2023-04-05
[Coral-Trino] Migrate 'CurrentTimestampTransformer' from RelNode to SqlNode layer

ljfgem:

This PR migrates `CurrentTimestampTransformer` from RelNode to SqlNode layer.
Besides, it corrects `public SqlNode visit(SqlCall call)` to make the traversal pre-order, because we want to apply each transformer to the input `SqlCall` only once. Previously without this change, `CurrentTimestampTransformer` will encounter stack overflow issue because the transformed SqlCall `CAST(CURRENT_TIMESTAMP AS TIMESTAMP(3))` also contains `CURRENT_TIMESTAMP`, changing the traversal to be preorder helps to avoid this issue.

1. Existing UTs which already covered this change
2. Regression test on all the production views, no translation failures or result differences

### DIFF
--- a/coral-spark/src/main/java/com/linkedin/coral/spark/CoralToSparkSqlCallConverter.java
+++ b/coral-spark/src/main/java/com/linkedin/coral/spark/CoralToSparkSqlCallConverter.java
@@ -170,7 +170,6 @@ public class CoralToSparkSqlCallConverter extends SqlShuttle {
 
   @Override
   public SqlNode visit(SqlCall call) {
-    final SqlCall transformedSqlCall = sqlCallTransformers.apply(call);
-    return super.visit(transformedSqlCall);
+    return sqlCallTransformers.apply((SqlCall) super.visit(call));
   }
 }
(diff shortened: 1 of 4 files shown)


coral-spark/src/main/java/com/linkedin/coral/spark/CoralToSparkSqlCallConverter.java
```diff
@@ -170,7 +170,6 @@ public CoralToSparkSqlCallConverter(Set<SparkUDFInfo> sparkUDFInfos) {
 
   @Override
   public SqlNode visit(SqlCall call) {
-    final SqlCall transformedSqlCall = sqlCallTransformers.apply(call);
```
yiqiangin, line 173: Why is this class modified? It looks like this class is in Coral-Spark not Coral-Trino

ljfgem: Good point, I removed this change since this PR should be for Coral-Trino migration only, will include the change for future migration in Coral-Spark.

yiqiangin (approved):

PR #385, opened 2023-04-06
[Coral-Trino] Migrate 'adjustReturnTypeWithCast' transformation from RelNode to SqlNode layer

ljfgem:

This PR migrates 'adjustReturnTypeWithCast' transformation from RelNode to SqlNode layer.

1. Existing UTs which covers the change, made 2 changes:
- Removed `CAST` in `TRANSFORM(\"if\"(\"$cor0\".\"b\" IS NOT NULL AND CAST(CARDINALITY(\"$cor0\".\"b\") AS INTEGER) > 0, \"$cor0\".\"b\", ARRAY[NULL]), x -> ROW(x)))`, because current `RelToTrinoConverter` uses SqlLiteral for [transform wrapper](https://github.com/linkedin/coral/blob/master/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/RelToTrinoConverter.java#L194), then `SqlCallTransformer` can't transform `CARDINALITY` in that SqlLiteral. But it's fine because we add `CAST` to make selected field types align between Hive and Trino, `CARDINALITY` here is generated by Coral without appearing in the selected fields, so it will not affect the view queries on Trino
- Modified `VARCHAR(65535)` to `VARCHAR`, it's fine because [the max length of VARCHAR is optional in Trino](https://trino.io/docs/current/language/types.html#varchar)
2. Regression test on all the production views, no regression after verification

### DIFF
--- a/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/ReturnTypeAdjustmentTransformer.java
+++ b/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/ReturnTypeAdjustmentTransformer.java
@@ -1,0 +1,69 @@
+/**
+ * Copyright 2023 LinkedIn Corporation. All rights reserved.
+ * Licensed under the BSD-2 Clause license.
+ * See LICENSE in the project root for license information.
+ */
+package com.linkedin.coral.trino.rel2trino.transformers;
+
+import java.util.Locale;
+import java.util.Map;
+
+import com.google.common.collect.ImmutableMap;
+
+import org.apache.calcite.sql.SqlBasicTypeNameSpec;
+import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlDataTypeSpec;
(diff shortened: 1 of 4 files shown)


yiqiangin (approved):

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/ReturnTypeAdjustmentTransformer.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.calcite.sql.type.SqlTypeName;
+
+import com.linkedin.coral.common.transformers.SqlCallTransformer;
+
+import static com.linkedin.coral.trino.rel2trino.CoralTrinoConfigKeys.*;
+
+
+/**
+ * This transformer casts the result of some Trino functions to the same return type as in Hive.
+ *
```
wmoustafa, line 28: Trino transformers should not reference Hive semantics as the API is Coral IR.

PR #391, opened 2023-04-17
Add Java version requirement in README

ljfgem:

Without specifying the required Java version, users may not use the correct Java version to build Coral, like #341. This PR is to add the Java version requirement.

No test.


aastha25 (approved):
thanks for the PR @ljfgem ! LGTM

PR #394, opened 2023-04-20
[WIP][Coral-Common] Remove CAST in HiveViewTable#toRel

ljfgem:

This PR is to remove all the `CAST` in [HiveViewTable.toRel](https://github.com/linkedin/coral/blob/master/coral-common/src/main/java/com/linkedin/coral/common/HiveViewTable.java#L49) to solve Nested View Stale Schema Issue.

#### Issue reproduction and RCA analysis
Run the following SQL on Hive CLI:
```
DROP TABLE IF EXISTS t;
CREATE TABLE t (s struct<a:int,b:boolean>);
CREATE OR REPLACE VIEW v1 AS SELECT * FROM t;
CREATE OR REPLACE VIEW v2 AS SELECT * FROM v1;
ALTER TABLE t CHANGE s s struct<a:int>;
```
Then schemas in HMS are:
```
t: (s struct<a:int>)
v1: (s struct<a:int,b:boolean>)
v2: (s struct<a:int,b:boolean>)
```
Note that after we change the schema of base table t, the schemas of views (v1, v2) will not be refreshed automatically.

While converting view v2 to RelNode, Coral will call the [HiveViewTable.toRel](https://github.com/linkedin/coral/blob/master/coral-common/src/main/java/com/linkedin/coral/common/HiveViewTable.java#L49) method while expanding nested view v1, [HiveViewTable.toRel](https://github.com/linkedin/coral/blob/master/coral-common/src/main/java/com/linkedin/coral/common/HiveViewTable.java#L49) is used to convert view v1 to RelNode.

In [HiveViewTable.toRel](https://github.com/linkedin/coral/blob/master/coral-common/src/main/java/com/linkedin/coral/common/HiveViewTable.java#L49), a key step is to [adjust the RelNode’s type with CAST](https://github.com/linkedin/coral/blob/master/coral-common/src/main/java/com/linkedin/coral/common/HiveViewTable.java#L62) if [isRowCastRequired](https://github.com/linkedin/coral/blob/master/coral-common/src/main/java/com/linkedin/coral/common/HiveViewTable.java#L80), so that v1 RelNode’s type will be aligned with its schema stored in HMS. In this example, the RelNode type derived from v1’s SQL is `RecordType(INTEGER a)`, while v1’s schema stored in HMS is `RecordType(INTEGER a, BOOLEAN b)`.

Based on the current logic of [isRowCastRequired](https://github.com/linkedin/coral/blob/master/coral-common/src/main/java/com/linkedin/coral/common/HiveViewTable.java#L80), if the types mismatch, we add CAST to the view’s schema stored in HMS [except extra fields are added on struct type columns of the base table](https://github.com/linkedin/coral/blob/master/coral-common/src/main/java/com/linkedin/coral/common/HiveViewTable.java#L77).

In the example above, since we removed field b on struct type column s from base table t, [HiveViewTable.toRel](https://github.com/linkedin/coral/blob/master/coral-common/src/main/java/com/linkedin/coral/common/HiveViewTable.java#L49) will add CAST to the RelNode, then the generated RelNode of v1 and v2 becomes:

```
LogicalProject(s=[CAST($0):RecordType(INTEGER a, BOOLEAN b)])
  LogicalProject(s=[$0])
    LogicalTableScan(table=[[hive, t]])
```

Then converted Spark SQL of v2 is: 
```
SELECT t.s s
FROM t
```
Note that `CAST` is [removed](https://github.com/linkedin/coral/blob/master/coral-spark/src/main/java/com/linkedin/coral/spark/SparkSqlRewriter.java#L41) from Spark SQL because Spark SQL doesn't support CAST to a row/struct.

And the converted Avro schema of v2 is:
```
{
  "type" : "record",
  "name" : "v2",
  "namespace" : "v2",
  "fields" : [ {
    "name" : "s",
    "type" : [ "null", {
      "type" : "record",
      "name" : "s",
      "namespace" : "v2.v2",
      "fields" : [ {
        "name" : "a",
        "type" : [ "null", "int" ]
      }, {
        "name" : "b",
        "type" : [ "null", "boolean" ]
      } ]
    } ],
    "doc" : "Field created in view by applying operator 'CAST' with argument(s): t.s"
  } ]
}
```
Note that the Coral schema will be aligned to the RelNode with CAST.

While querying v2 on Spark, Spark will execute the input Spark SQL (`SELECT t.s s FROM t`) and try to upcast the analyzed schema (`s: struct<a: int>`) to the schema generated by Coral-Schema above, then the upcast will fail:
```
org.apache.spark.sql.AnalysisException: Cannot up cast `s` from struct<a:int> to struct<a:int,b:boolean>.
```
So the RCA chain is:
1. Base table schema is changed (not adding extra fields on struct type columns of the base table) while the intermediate view’s and top view’s schema are stale
2. Coral CAST the RelNode based on the intermediate view’ stale schema
3. Coral-Schema produces Avro schema based on the RelNode with CAST
4. There is a mismatch between Spark’s analyzed Schema from Spark SQL and Coral Schema
5. Spark is not able to up cast the Spark’s analyzed Schema to Coral schema
6. Query fails

To solve the issue, we can remove all the `CAST` in [HiveViewTable.toRel](https://github.com/linkedin/coral/blob/master/coral-common/src/main/java/com/linkedin/coral/common/HiveViewTable.java#L49) so that whether the intermediate view’s schema is stale or not doesn’t matter anymore, the view v2 above can be queried with the latest base table’s schema.

This PR also includes some changes to fix the regressions after removing `CAST` in [HiveViewTable.toRel](https://github.com/linkedin/coral/blob/master/coral-common/src/main/java/com/linkedin/coral/common/HiveViewTable.java#L49), details will be added once the PR is ready for review.

TBA


andrewchoi5 (approved):

ljfgem:
Closing it given #409 has been merged.

PR #397, opened 2023-04-25
[Coral-Hive] Remove UDF definition from StaticHiveFunctionRegistry

aastha25:

Remove `SearchQueryUDF` UDF initialization from StaticRegistry and rely on dynamic resolution instead.
This UDF has an outdated return type registered. Instead of manually keeping up with any changes introduced in this UDF, we can rely on dynamic resolution instead. 

./gradlew build
integ tested the impacted views on the production cluster


ljfgem (approved):
LGTM, thanks @aastha25 !
Let's modify the title to be `[Coral-Hive] Remove UDF definition from StaticHiveFunctionRegistry`?

PR #398, opened 2023-04-25
Fix coral-dbt Python build issue

AliceYeh12:
This fixes a coral-dbt build issue by directing Python package installation to the proper location. 

The patch was tested locally and on a previously failing project build.

### DIFF
--- a/coral-dbt/build.gradle
+++ b/coral-dbt/build.gradle
@@ -14,6 +14,8 @@ spotless {
 task runCoralDbtTests {
   // Install Python dependencies
   def installModulesProcessBuilder = new ProcessBuilder([
+    'python3',
+    '-m',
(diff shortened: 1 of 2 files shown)


coral-dbt/build.gradle
```diff
@@ -14,6 +14,8 @@ spotless {
 task runCoralDbtTests {
   // Install Python dependencies
   def installModulesProcessBuilder = new ProcessBuilder([
+    'python3',
```
aastha25, line 17: does it mean the build will fail with python2?
if so, could you also update the Projetc's README to include a dependency on Python3?

aastha25 (approved):
thanks for the quick turnaround @AliceYeh12 !
LGTM

PR #399, opened 2023-04-25
Ensure pip installation on build

AliceYeh12:
coral-dbt fails on machines that do not have pip installed so we need to add pip installation to the build process.

The fix resolves the Jenkins build failure.

### DIFF
--- a/coral-dbt/build.gradle
+++ b/coral-dbt/build.gradle
@@ -12,53 +12,66 @@ spotless {
 }
 
 task runCoralDbtTests {
-  // Ensure pip is installed
-  def ensurePipProcessBuilder = new ProcessBuilder([
-    'python3',
-    '-m',
-    'ensurepip',
-    '--upgrade'
-  ])
-  def ensurePipProcess = ensurePipProcessBuilder.start()
-  ensurePipProcess.waitFor()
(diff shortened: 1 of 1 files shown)


coral-dbt/build.gradle
```diff
@@ -12,43 +12,74 @@ spotless {
...
-    '-m',
-    'pip',
-    'install',
-    '-r',
-    'requirements.txt'
-  ])
-  installModulesProcessBuilder.directory(file('src/main/resources/tests'))
-  def installModulesProcess = installModulesProcessBuilder.start()
-  installModulesProcess.waitFor()
+  try {
```
aastha25, line 33: this block shouldnt be triggered if the previous try-catch failed?

AliceYeh12: I was thinking in case it is a permissions problem but the user already has pip installed.

aastha25: but we didnt validate if pip is installed.

AliceYeh12: Yes, if that is the case, the second try-catch block will catch that.

aastha25 (approved):
LGTM

PR #400, opened 2023-04-26
[WIP] Add support for incremental rewrite of nested queries

AliceYeh12:

### DIFF
--- a/coral-incremental/src/main/java/com/linkedin/coral/incremental/IncrementalTransformerResults.java
+++ b/coral-incremental/src/main/java/com/linkedin/coral/incremental/IncrementalTransformerResults.java
@@ -1,0 +1,76 @@
+/**
+ * Copyright 2023 LinkedIn Corporation. All rights reserved.
+ * Licensed under the BSD-2 Clause license.
+ * See LICENSE in the project root for license information.
+ */
+package com.linkedin.coral.incremental;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.calcite.rel.RelNode;
+
+
+public class IncrementalTransformerResults {
+
+  private RelNode incrementalRelNode;
+  private RelNode refreshRelNode;
+  private Map<String, RelNode> intermediateQueryRelNodes;
+
+  public IncrementalTransformerResults() {
+    incrementalRelNode = null;
+    refreshRelNode = null;
+    intermediateQueryRelNodes = new HashMap<>();
+  }
+
+  public boolean existsIncrementalRelNode() {
+    return incrementalRelNode != null;
+  }
+
+  public RelNode getIncrementalRelNode() {
+    return incrementalRelNode;
+  }
+
+  public boolean existsRefreshRelNode() {
+    return refreshRelNode != null;
+  }
+
+  public RelNode getRefreshRelNode() {
+    return refreshRelNode;
+  }
+
+  public Map<String, RelNode> getIntermediateQueryRelNodes() {
+    return intermediateQueryRelNodes;
+  }
+
+  public boolean containsIntermediateQueryRelNodeKey(String name) {
+    return intermediateQueryRelNodes.containsKey(name);
+  }
+
+  public RelNode getIntermediateQueryRelNodeCorrespondingToKey(String name) {
+    return intermediateQueryRelNodes.get(name);
+  }
+
+  public void setIncrementalRelNode(RelNode incrementalRelNode) {
+    this.incrementalRelNode = incrementalRelNode;
+  }
+
+  public void setRefreshRelNode(RelNode refreshRelNode) {
+    this.refreshRelNode = refreshRelNode;
+  }
+
+  public void setIntermediateQueryRelNodes(Map<String, RelNode> intermediateQueryRelNodes) {
+    this.intermediateQueryRelNodes = intermediateQueryRelNodes;
+  }
+
+  public void addIntermediateQueryRelNode(String name, RelNode intermediateRelNode) {
+    this.intermediateQueryRelNodes.put(name, intermediateRelNode);
+  }
+
+  public void addMultipleIntermediateQueryRelNodes(Map<String, RelNode> intermediateQueryRelNodes) {
+    if (intermediateQueryRelNodes != null) {
+      this.intermediateQueryRelNodes.putAll(intermediateQueryRelNodes);
+    }
+  }
+
+}
--- a/coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeIncrementalTransformer.java
+++ b/coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeIncrementalTransformer.java
@@ -7,10 +7,12 @@ package com.linkedin.coral.incremental;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+import org.apache.calcite.plan.RelOptSchema;
 import org.apache.calcite.plan.RelOptTable;
 import org.apache.calcite.prepare.RelOptTableImpl;
 import org.apache.calcite.rel.RelNode;
@@ -29,14 +31,29 @@ import org.apache.calcite.rex.RexNode;
 
 public class RelNodeIncrementalTransformer {
 
+  private static RelOptSchema relOptSchema;
+
   private RelNodeIncrementalTransformer() {
   }
 
-  public static RelNode convertRelIncremental(RelNode originalNode) {
+  public static IncrementalTransformerResults performIncrementalTransformation(RelNode originalNode) {
+    IncrementalTransformerResults incrementalTransformerResults = convertRelIncremental(originalNode);
+    return incrementalTransformerResults;
+  }
+
+  private static IncrementalTransformerResults convertRelIncremental(RelNode originalNode) {
+    IncrementalTransformerResults incrementalTransformerResults = new IncrementalTransformerResults();
     RelShuttle converter = new RelShuttleImpl() {
       @Override
       public RelNode visit(TableScan scan) {
         RelOptTable originalTable = scan.getTable();
+
+        // Set relOptSchema
+        if (relOptSchema == null) {
+          relOptSchema = originalTable.getRelOptSchema();
+        }
+
+        // Create delta scan
         List<String> incrementalNames = new ArrayList<>(originalTable.getQualifiedName());
         String deltaTableName = incrementalNames.remove(incrementalNames.size() - 1) + "_delta";
         incrementalNames.add(deltaTableName);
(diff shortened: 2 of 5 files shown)


coral-incremental/src/test/java/com/linkedin/coral/incremental/RelToIncrementalSqlConverterTest.java
```diff
@@ -143,4 +110,12 @@ public void testSelectSpecificJoin() {
         + "INNER JOIN test.bar2_delta AS bar2_delta0 ON bar1_delta0.x = bar2_delta0.x) AS t0";
     assertEquals(getIncrementalModification(sql), expected);
   }
+
+  @Test
+  public void testJoinOverJoin() {
```
wmoustafa, line 115: Can we add 3 nested joins test case?

coral-incremental/src/main/java/com/linkedin/coral/incremental/IncrementalTransformerResults.java
```diff
@@ -0,0 +1,90 @@
...
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.calcite.rel.RelNode;
+
+
+public class IncrementalTransformerResults {
```
wmoustafa, line 16: I think the code will be much simpler if we get rid of this class, and move the members to the visitor itself.

coral-incremental/src/main/java/com/linkedin/coral/incremental/IncrementalTransformerResults.java
```diff
@@ -0,0 +1,90 @@
...
+
+  private RelNode incrementalRelNode;
+  private RelNode refreshRelNode;
+  private Map<String, RelNode> intermediateQueryRelNodes;
+  private List<String> intermediateOrderings;
```
wmoustafa, line 21: As we discussed does it make sense to get rid of `incrementalRelNode`? Is it a special case of `Map<String, RelNode> intermediateQueryRelNodes`?

coral-incremental/src/main/java/com/linkedin/coral/incremental/IncrementalTransformerResults.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Map;
+
+import org.apache.calcite.rel.RelNode;
+
+
+public class IncrementalTransformerResults {
+
+  private RelNode incrementalRelNode;
+  private RelNode refreshRelNode;
+  private Map<String, RelNode> intermediateQueryRelNodes;
```
wmoustafa, line 20: Should we do a map for deltas and a map for lasts?

coral-incremental/src/main/java/com/linkedin/coral/incremental/IncrementalTransformerResults.java
wmoustafa, line 21: `intermediateOrderings` --> `materializationOrder`?

coral-incremental/src/main/java/com/linkedin/coral/incremental/IncrementalTransformerResults.java
```diff
@@ -0,0 +1,90 @@
...
+  private List<String> intermediateOrderings;
+
+  public IncrementalTransformerResults() {
+    incrementalRelNode = null;
+    refreshRelNode = null;
```
wmoustafa, line 25: Should it be a list of refresh relnodes? We can remove from current PR till we converge on the best representation.

coral-incremental/src/main/java/com/linkedin/coral/incremental/IncrementalTransformerResults.java
wmoustafa: If we have 2 maps, then the keys may not require `_delta` and `_last` suffixes? Feel free to choose one that makes the implementation cleaner.

coral-incremental/src/main/java/com/linkedin/coral/incremental/IncrementalTransformerResults.java
```diff
@@ -0,0 +1,90 @@
...
+  public void setRefreshRelNode(RelNode refreshRelNode) {
+    this.refreshRelNode = refreshRelNode;
+  }
+
+  public void addIntermediateQueryRelNode(String name, RelNode intermediateRelNode) {
+    this.intermediateQueryRelNodes.put(name, intermediateRelNode);
+    addIntermediateOrdering(name);
+  }
+
+  public void addMultipleIntermediateQueryRelNodes(Map<String, RelNode> intermediateQueryRelNodes) {
```
wmoustafa, line 75: Getting rid of methods like this is why I believe moving the members of this class to the visitor class will make it look cleaner.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeIncrementalTransformer.java
```diff
@@ -49,61 +66,130 @@ public RelNode visit(TableScan scan) {
...
+        RelNode incrementalLeft = incrementalTransformerResultsLeft.getIncrementalRelNode();
+        RelNode incrementalRight = incrementalTransformerResultsRight.getIncrementalRelNode();
+        incrementalTransformerResults
+            .addMultipleIntermediateQueryRelNodes(incrementalTransformerResultsLeft.getIntermediateQueryRelNodes());
+        incrementalTransformerResults
+            .addMultipleIntermediateQueryRelNodes(incrementalTransformerResultsRight.getIntermediateQueryRelNodes());
 
         RexBuilder rexBuilder = join.getCluster().getRexBuilder();
 
+        // Check if we can replace the left and right nodes with a scan of a materialized table
```
wmoustafa, line 80: Separate to own method?

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeIncrementalTransformer.java
```diff
@@ -49,61 +66,130 @@ public RelNode visit(TableScan scan) {
...
+    return incrementalTransformerResults;
   }
 
-  private static LogicalProject createProjectOverJoin(LogicalJoin join, RelNode left, RelNode right,
+  private static String getTableNameFromDescription(RelNode relNode) {
+    String identifier = relNode.getDescription().split("#")[1];
+    return "Table#" + identifier;
+  }
+
+  private static LogicalProject createReplacementProjectNodeForGivenRelNode(String relOptTableName, RelNode relNode,
```
wmoustafa, line 170: Why is this needed? Hope it is not required. Any customization to the RelNodes means more fragility.

coral-incremental/src/test/java/com/linkedin/coral/incremental/RelToIncrementalSqlConverterTest.java
```diff
@@ -143,4 +110,31 @@ public void testSelectSpecificJoin() {
...
+  public void testNestedJoin() {
+    String nestedJoin = "SELECT a1, a2 FROM test.alpha JOIN test.beta ON test.alpha.a1 = test.beta.b1";
+    String sql = "SELECT a2, g1 FROM (" + nestedJoin + ") AS nj JOIN test.gamma ON nj.a2 = test.gamma.g2";
+    String expected = "SELECT t0.a2, t0.g1\n" + "FROM (SELECT *\n" + "FROM (SELECT *\n" + "FROM Table#0 AS Table#0\n"
+        + "INNER JOIN test.gamma_delta AS gamma_delta ON Table#0.a2 = gamma_delta.g2\n" + "UNION ALL\n" + "SELECT *\n"
+        + "FROM Table#0_delta AS Table#0_delta\n"
+        + "INNER JOIN test.gamma AS gamma ON Table#0_delta.a2 = gamma.g2) AS t\n" + "UNION ALL\n" + "SELECT *\n"
+        + "FROM Table#0_delta AS Table#0_delta0\n"
+        + "INNER JOIN test.gamma_delta AS gamma_delta0 ON Table#0_delta0.a2 = gamma_delta0.g2) AS t0";
+    assertEquals(getIncrementalModification(sql), expected);
```
wmoustafa, line 124: Can we assert all intermediate nodes (both _delta and _last)?

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeIncrementalTransformer.java
```diff
@@ -49,61 +66,130 @@ public RelNode visit(TableScan scan) {
...
       RexBuilder rexBuilder) {
-    LogicalJoin incrementalJoin =
-        LogicalJoin.create(left, right, join.getCondition(), join.getVariablesSet(), join.getJoinType());
+    RelOptTable table =
+        RelOptTableImpl.create(relOptSchema, relNode.getRowType(), Collections.singletonList(relOptTableName), null);
+    TableScan scan = LogicalTableScan.create(relNode.getCluster(), table);
+    return createProjectOverNode(scan, rexBuilder);
+  }
+
+  private static LogicalProject createProjectOverNode(RelNode relNode, RexBuilder rexBuilder) {
```
wmoustafa, line 178: How about making this `createJoin`, and having it construct both he join and projection node? That is semantically closer to what we are trying to achieve.

coral-incremental/src/main/java/com/linkedin/coral/incremental/IncrementalTransformerResults.java
AliceYeh12: Yes, this should be a list of refresh RelNodes—one for each aggregate query. Will remove for now.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeIncrementalTransformer.java
AliceYeh12: This is needed to convert a join to a Project-TableScan for the table/view substitutions but I'm not sure that there is necessarily any RelNode customization here.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeIncrementalTransformer.java
```diff
@@ -49,61 +66,130 @@ public RelNode visit(TableScan scan) {
...
             aggregate.getAggCallList());
       }
     };
-    return originalNode.accept(converter);
+    incrementalTransformerResults.setIncrementalRelNode(originalNode.accept(converter));
+    return incrementalTransformerResults;
   }
 
-  private static LogicalProject createProjectOverJoin(LogicalJoin join, RelNode left, RelNode right,
+  private static String getTableNameFromDescription(RelNode relNode) {
```
aastha25, line 165: all the utilities need a comment with an example.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeIncrementalTransformer.java
aastha25: Can you add a comment here to clarify what is the type of 'RelNode' passed here.

coral-incremental/src/main/java/com/linkedin/coral/incremental/IncrementalTransformerResults.java
aastha25: +1 to splitting out deltas & last

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeIncrementalTransformer.java
```diff
@@ -49,61 +66,130 @@ public RelNode visit(TableScan scan) {
       public RelNode visit(LogicalJoin join) {
         RelNode left = join.getLeft();
         RelNode right = join.getRight();
-        RelNode incrementalLeft = convertRelIncremental(left);
-        RelNode incrementalRight = convertRelIncremental(right);
+        IncrementalTransformerResults incrementalTransformerResultsLeft = convertRelIncremental(left);
+        IncrementalTransformerResults incrementalTransformerResultsRight = convertRelIncremental(right);
+        RelNode incrementalLeft = incrementalTransformerResultsLeft.getIncrementalRelNode();
+        RelNode incrementalRight = incrementalTransformerResultsRight.getIncrementalRelNode();
+        incrementalTransformerResults
```
aastha25, line 73: will it simplify the logic if we make `intermediateQueryRelNodes` global?

AliceYeh12: Yes, we are actually getting rid of IncrementalTransformerResults and keeping all logic inside of the transformer itself (larger discussion in the Slack channel).

coral-incremental/src/main/java/com/linkedin/coral/incremental/IncrementalTransformerResults.java
AliceYeh12: Sounds good, will split but maintain suffixes since I think this will come in handy on the code generation side.

coral-incremental/src/main/java/com/linkedin/coral/incremental/IncrementalTransformerResults.java
AliceYeh12: I'm now getting indices from the map so will get rid of this entirely.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeIncrementalTransformer.java
```diff
@@ -29,14 +31,29 @@
...
+
+  private static IncrementalTransformerResults convertRelIncremental(RelNode originalNode) {
+    IncrementalTransformerResults incrementalTransformerResults = new IncrementalTransformerResults();
     RelShuttle converter = new RelShuttleImpl() {
       @Override
       public RelNode visit(TableScan scan) {
         RelOptTable originalTable = scan.getTable();
+
+        // Set relOptSchema
+        if (relOptSchema == null) {
```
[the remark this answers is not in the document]
AliceYeh12: This check is to set `relOptSchema` only when it has not previously already been set. I don't think it's absolutely necessary but I just added a comment to explain the conditional's purpose.

PR #403, opened 2023-05-03
[Coral-common] Modify return type for Multiply operator

aastha25:

Given a SQL with multiply (*) operation between integer operands, Calcite by default evaluates the return type to be `INT`. This may not be the right return type, and hence the translated Spark & Trino SQLs could fail on the engine. 
This PR overrides this behavior to return the `BIGINT` return type instead. 

The PR also adds additional validation for similar override for the divide (/) operator. 

Updated UTs
./gradlew build
tested the patch against production views
tested the impacted view on the production cluster

### DIFF
--- a/coral-common/src/main/java/com/linkedin/coral/common/HiveTypeSystem.java
+++ b/coral-common/src/main/java/com/linkedin/coral/common/HiveTypeSystem.java
@@ -185,10 +185,10 @@ public class HiveTypeSystem extends RelDataTypeSystemImpl {
   @Override
   public RelDataType deriveDecimalMultiplyType(RelDataTypeFactory typeFactory, RelDataType type1, RelDataType type2) {
     if (SqlTypeUtil.isExactNumeric(type1) && SqlTypeUtil.isExactNumeric(type2)) {
-      if (SqlTypeUtil.isBigint(type1) || SqlTypeUtil.isBigint(type2)) {
-        return nullableType(typeFactory, SqlTypeName.BIGINT);
-      } else {
+      if (SqlTypeUtil.isDecimal(type1) || SqlTypeUtil.isDecimal(type2)) {
         return super.deriveDecimalMultiplyType(typeFactory, type1, type2);
+      } else if (SqlTypeUtil.isBigint(type1) || SqlTypeUtil.isBigint(type2)) {
+        return nullableType(typeFactory, SqlTypeName.BIGINT);
       }
     }
     return null;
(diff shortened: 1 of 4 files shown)


coral-common/src/main/java/com/linkedin/coral/common/HiveTypeSystem.java
```diff
@@ -168,11 +168,30 @@ public boolean isSchemaCaseSensitive() {
...
+   * This override makes sure that the multiply operator, "*", returns {@link org.apache.calcite.sql.type.ReturnTypes#BIGINT}
+   * if neither of the operands is decimal type, which is compatible with the expected data type in Hive/Spark.
+   */
+  @Override
+  public RelDataType deriveDecimalMultiplyType(RelDataTypeFactory typeFactory, RelDataType type1, RelDataType type2) {
+    if (SqlTypeUtil.isExactNumeric(type1) && SqlTypeUtil.isExactNumeric(type2)) {
+      if (SqlTypeUtil.isDecimal(type1) || SqlTypeUtil.isDecimal(type2)) {
+        return super.deriveDecimalMultiplyType(typeFactory, type1, type2);
+      } else {
+        return nullableType(typeFactory, SqlTypeName.BIGINT);
```
ljfgem, line 191: I tried on Hive CLI, it returns `BIGINT` only if there is at least one `BIGINT` operand.
i.e.
```
int * int -> int
bigint * int -> bigint
int * bigint -> bigint
bigint * bigint -> bigint
```

aastha25: I tested a UT in coral-schema with `CREATE VIEW v AS " + "SELECT 22*bc.Id AS Temp FROM basecomplex bc` and this generated [null, long] schema. 
Can you share how you are testing?

ljfgem: I am using Hive and Spark CLI:
```
hive (default)> create or replace view test_multiply as select 1*2;
OK
Time taken: 1.05 seconds
hive (default)> desc test_multiply;
OK
_c0                 	int       
```
and
```
scala> spark.sql("select 100000000*2000000000").printSchema
root
 |-- (100000000 * 2000000000): integer (nullable = false)
```

ljfgem: I think we can add an `if` block to determine if any operand is a `BIGINT` and return `BIGINT` it it's true.

aastha25: makes sense. It turns out users are expected to include explicit CAST to BIGINT if int multiplication can result in an overflow. I have revised the PR and tested the impacted production view

ljfgem (approved):
LGTM, thanks @aastha25!

PR #406, opened 2023-05-06
[Coral-Visualization] Create Visualization Util for RelNode representation

wmoustafa:
This patch introduces a visualization utility for the RelNode representation.  The main class is `RelNodeVisualizationShuttle` which implements a `RelNodeShuttle` that constructs a Graphvis graph while traversing the `RelNode` tree. Below are the two example renderings of the two test cases implemented in this PR.

`./gradlew build`


coral-visualization/src/main/java/com/linkedin/coral/vis/RelNodeVisualizationShuttle.java
```diff
@@ -0,0 +1,182 @@
...
+
+  @Override
+  public RelNode visit(TableFunctionScan scan) {
+    super.visit(scan);
+    nodeMap.put(scan, node(scan.getCall().toString()));
+    return scan;
+  }
+
+  @Override
+  public RelNode visit(LogicalValues values) {
```
aastha25, line 68: can we drop `LogicalValues` RelNode when its rowType is `RecordType(Integer Zero)`? 
what is the value added in the viz by expressing this node?

coral-visualization/src/main/java/com/linkedin/coral/vis/RelNodeVisualizationShuttle.java
```diff
@@ -0,0 +1,182 @@
...
+    String leftLabel = getLabel(leftChild, 0, true);
+    String rightLabel = getLabel(rightChild, leftChild.getRowType().getFieldCount(), false);
+    LinkTarget leftEdge = edge(leftChild, leftLabel);
+    LinkTarget rightEdge = edge(rightChild, rightLabel);
+    nodeMap.put(correlate, correlateNode.link(new LinkTarget[] { leftEdge, rightEdge }));
+    return correlate;
+  }
+
+  private String getLabel(RelNode target, int startIndex, boolean isCorrelateSide) {
+    String label = "";
```
aastha25, line 122: are we able to convert all `$i` representations to column names from rowType?
If so, we can directly use the column names in each node, ex: `Project: [a,b,c,d]` and `hive.test.foo [a,b]` (from scan's rowType) instead of adding text on the edges?

aastha25 (approved):
Mostly LGTM. I started some threads to discuss ideas.

PR #409, opened 2023-05-15
Generate Coral's RelNode for views from base table schema

aastha25:

This PR includes the following fixes:
(1) Includes changes from https://github.com/linkedin/coral/pull/394 which aimed to remove CAST in [HiveViewTable.toRel](https://github.com/linkedin/coral/blob/master/coral-common/src/main/java/com/linkedin/coral/common/HiveViewTable.java#L49) to solve Nested View Stale Schema Issue.
(2) Enables avro schema generation for a dataset with map field with NULL key
(3) When a field schema is a UnionType of a single data type, such as `uniontype<array<string>>`, surface the underlying field & its schema in Coral's SqlNode RelNode representation, such as `array<string>`. `SingleUnionFieldReferenceTransformer` has been introduced for backward compatibility. 

For a table `t1`: (f1: uniontype<int> ) and a view `v1` defined on top of it `SELECT custom_udf_to_remove_single_type_union(f1) AS f2 FROM t1` 
`v1` has a schema: (f2: int)
The RelNode representation, prior to dropping the CAST operator in the RelNode, would force the derived data type to be `int` for f2, as expected. Hence, the generated avro schema could be analyzed by the Spark engine. However, after dropping the CAST, as documented [here](https://github.com/linkedin/coral/pull/394), the derived type is `Struct<tag_0:int>`. The Avro schema is incorrectly generated and cannot be analysed by the Spark engine.

The resolution is to derive the data type of single uniontype fields like `f1` as the underlying data type (int in this case). 

./gradlew spotlessApply
./gradlew build
Updated UTs
Thoroughly tested production views across all supported engines.

### DIFF
--- a/coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
+++ b/coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2017-2022 LinkedIn Corporation. All rights reserved.
+ * Copyright 2017-2023 LinkedIn Corporation. All rights reserved.
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
--- a/coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
+++ b/coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
@@ -146,6 +146,9 @@ public class TypeConverter {
   public static RelDataType convert(UnionTypeInfo unionType, RelDataTypeFactory dtFactory) {
     List<RelDataType> fTypes = unionType.getAllUnionObjectTypeInfos().stream()
         .map(typeInfo -> convert(typeInfo, dtFactory)).collect(Collectors.toList());
+    if (fTypes.size() == 1) {
+      return dtFactory.createTypeWithNullability(fTypes.get(0), true);
+    }
     List<String> fNames = IntStream.range(0, unionType.getAllUnionObjectTypeInfos().size()).mapToObj(i -> "field" + i)
         .collect(Collectors.toList());
     fTypes.add(0, dtFactory.createSqlType(SqlTypeName.INTEGER));
--- a/coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
+++ b/coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
@@ -140,6 +140,7 @@ public class TypeConverter {
   // The schema of output Struct conforms to https://github.com/trinodb/trino/pull/3483
   // except we adopted "integer" for the type of "tag" field instead of "tinyint" in the Trino patch
   // for compatibility with other platforms that Iceberg currently doesn't support tinyint type.
+  // When the field count inside UnionTypeInfo is one, we surface the underlying RelDataType instead.
 
   // Note: this is subject to change in the future pending on the discussion in
   // https://mail-archives.apache.org/mod_mbox/iceberg-dev/202112.mbox/browser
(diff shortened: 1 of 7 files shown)


ljfgem:
@aastha25: Thanks for this PR. Could we create a PR for union type handling and keep #394 ? Or keep the author information of #394 if this PR can't be splitted?

aastha25:
@ljfgem Unfortunately the PR cant be split up. All these changes need to go in together. I have updated the author's information for the commits related to #394.

coral-hive/src/main/java/com/linkedin/coral/transformers/SingleUnionFieldReferenceTransformer.java
```diff
@@ -0,0 +1,49 @@
...
+ * This indicates that the first operand represents a Union data type with a single data type inside.
+ * (2) Examples of such SqlCalls include extract_union(product.value).tag_0 or (extract_union(product.value).id).tag_0.
+ * (3) The transformation for such SqlCalls is to return the first operand.
+ */
+public class SingleUnionFieldReferenceTransformer extends SqlCallTransformer {
+  private static final String TAG_0_OPERAND = "tag_0";
+
+  public SingleUnionFieldReferenceTransformer(SqlValidator sqlValidator) {
+    super(sqlValidator);
+  }
```
ljfgem, line 28: `sqlValidator` is not necessary if we don't need to call `getType(SqlNode)` method in this transformer?

ljfgem (commented):
Thank you @aastha25 for this PR!
Could you please add more details to the PR description and the code comment to clarify the necessity of the uniontype part changes? i.e. why it will fail if we don't make those changes, preferably with simple example?

coral-hive/src/main/java/com/linkedin/coral/transformers/SingleUnionFieldReferenceTransformer.java
aastha25: im using `getRelDataType()` [here](https://github.com/linkedin/coral/pull/409/files/162e2baa7f2cc107d8602219eab43bd15b4b42f9#diff-aca6363aefb30015f0a9c5fb6bef3aafd0f3638cc897c903e125bb67bb6f6393R35) to derive if the first operand is struct or not

coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java
```diff
@@ -140,12 +140,16 @@ public static RelDataType convert(StructTypeInfo structType, final RelDataTypeFa
...
   // for compatibility with other platforms that Iceberg currently doesn't support tinyint type.
+  // When the field count inside UnionTypeInfo is one, we surface the underlying RelDataType instead.
 
   // Note: this is subject to change in the future pending on the discussion in
   // https://mail-archives.apache.org/mod_mbox/iceberg-dev/202112.mbox/browser
   public static RelDataType convert(UnionTypeInfo unionType, RelDataTypeFactory dtFactory) {
     List<RelDataType> fTypes = unionType.getAllUnionObjectTypeInfos().stream()
         .map(typeInfo -> convert(typeInfo, dtFactory)).collect(Collectors.toList());
+    if (fTypes.size() == 1) {
+      return dtFactory.createTypeWithNullability(fTypes.get(0), true);
```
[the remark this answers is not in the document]
aastha25: Correct, we want to surface the underlying data type (ex int) when the base table schema has a single type union (ex uniontype<int>).

aastha25:
@ljfgem I have added an example clarifying the usecase in the PR description.

ljfgem (approved):
LGTM, thank you @aastha25 !

wmoustafa:
Should not this PR change the behavior of `CoalesceStructUtility` so that if the input is a field with single type, it returns a struct?

aastha25:
This PR sets the return type of `extract_union(string)` as `string`, making the operator a no-op. I can modify the comment on the class to document this change if required.

PR #414, opened 2023-05-26
[Coral-Schema] Uniquify namespace for fields with same name

aastha25:
Prior to this PR, the avro schema generated for struct fields in a dataset had a default namespace "rel_avro". Hence, if a dataset has nested struct fields with the same field name, `f1`, the namespace generated for both fields is `rel_avro.f1` .
This resulted in an error like:

`Caused by: java.lang.IllegalStateException: Cannot process recursive Avro record rel_avro.f1` 

from different code paths and at different stages in the processing logic such as [here](https://github.com/linkedin/coral/blob/9eeeffe3131acee724d033b8185fbc01f0d9cbda/coral-schema/src/main/java/com/linkedin/coral/schema/avro/AvroSchemaVisitor.java#L23), [here](https://github.com/linkedin/coral/blob/9eeeffe3131acee724d033b8185fbc01f0d9cbda/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java#L485) etc.

This PR also includes a unit test `testNestedRecordWithSameNameFields` which demonstrates this corner case and would fail without these code changes.

Even if the same name struct fields are not nested, it can result in the same exception during parsing/merging as the default namespace generated for both fields would be the same, `rel_avro.f1`.  To rectify this issue, `mergeUnionRecordSchema()` method  now re-generates the namespace for the fields in the two schemas (originalLeftSchema, originalRightSchema) when strictMode = false to uniquify the field namespaces. The same logic is applied on the final schema [here](https://github.com/linkedin/coral/blob/9eeeffe3131acee724d033b8185fbc01f0d9cbda/coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java#L205) when strictMode=false to generate the right field namespaces. However, this transformation is also needed for intermediate merge steps. 

./gradlew build
added & updated UTs
Regression tested against production views & the previously failing view.

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
@@ -52,7 +52,15 @@ class RelDataTypeToAvroType {
     Preconditions.checkNotNull(relDataType);
 
     if (relDataType instanceof RelRecordType || relDataType instanceof DynamicRecordType) {
-      return relRecordTypeToAvroType(relDataType, null, recordName, "rel_avro", null);
+      String uniqueNamespace;
+      if (relDataType.getFieldList() == null || relDataType.getFieldCount() == 0) {
+        uniqueNamespace = "rel_avro";
+      } else {
+        uniqueNamespace =
+            "namespace_from_first_nested_field_" + relDataType.getFieldList().get(0).getName().toLowerCase();
+      }
+
+      return relRecordTypeToAvroType(relDataType, null, recordName, uniqueNamespace, null);
     }
 
     if (relDataType instanceof BasicSqlType) {
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
@@ -7,6 +7,7 @@ package com.linkedin.coral.schema.avro;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import javax.annotation.Nonnull;
 
@@ -52,14 +53,7 @@ class RelDataTypeToAvroType {
     Preconditions.checkNotNull(relDataType);
 
     if (relDataType instanceof RelRecordType || relDataType instanceof DynamicRecordType) {
-      String uniqueNamespace;
-      if (relDataType.getFieldList() == null || relDataType.getFieldCount() == 0) {
-        uniqueNamespace = "rel_avro";
-      } else {
-        uniqueNamespace =
-            "namespace_from_first_nested_field_" + relDataType.getFieldList().get(0).getName().toLowerCase();
-      }
-
+      String uniqueNamespace = getUniqueNamespace(relDataType, recordName);
       return relRecordTypeToAvroType(relDataType, null, recordName, uniqueNamespace, null);
     }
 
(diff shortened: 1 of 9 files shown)


coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
```diff
@@ -52,7 +52,15 @@ static Schema relDataTypeToAvroTypeNonNullable(@Nonnull RelDataType relDataType,
     Preconditions.checkNotNull(relDataType);
 
     if (relDataType instanceof RelRecordType || relDataType instanceof DynamicRecordType) {
-      return relRecordTypeToAvroType(relDataType, null, recordName, "rel_avro", null);
+      String uniqueNamespace;
+      if (relDataType.getFieldList() == null || relDataType.getFieldCount() == 0) {
+        uniqueNamespace = "rel_avro";
+      } else {
+        uniqueNamespace =
+            "namespace_from_first_nested_field_" + relDataType.getFieldList().get(0).getName().toLowerCase();
```
yiqiangin, line 60: It looks like this naming pattern just adds the name of the first field in the struct as the suffix of the name space, so it seems cannot cover the corner case like 
```
struct f1 {
  struct f1 {
      struct f1
   }
}
```
all the fields are the first field in the struct in each level, so all the value of the name space still are the same like "namespace_from_first_nested_field_f1"
Please correct me if my understanding is wrong.

coral-schema/src/test/resources/testNullabilityExtractUnionUDF-expected.avsc
```diff
@@ -7,12 +7,13 @@
     "type" : [ "null", {
       "type" : "record",
       "name" : "c1",
-      "namespace" : "rel_avro",
+      "namespace" : "namespace_from_first_nested_field_tag_0",
```
wmoustafa, line 10: This sounds to be backward incompatible for DSM, correct?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
wmoustafa: +1 Not clear what the intention of the naming convention is.

aastha25: It doesnt cover the corner case that you mentioned @yiqiangin. However, its the minimum code changes needed to generate the schema successfully for the specified corner case.

coral-schema/src/test/resources/testNullabilityExtractUnionUDF-expected.avsc
wmoustafa: Can we make it more general? It is fine if it needs more code.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
wmoustafa: Can we make it more general? It is fine if it needs more code.

coral-schema/src/test/resources/testNullabilityExtractUnionUDF-expected.avsc
wmoustafa: Can we make this change kick in only when there is ambiguity, and keep the old code path as is?

aastha25: Thanks for the suggestions. I have modified the logic to kick in only when there's ambiguity.
Also, the modifications in the merge utility only impact non-DSM code paths. PTAL

yiqiangin (approved):

PR #421, opened 2023-06-12
Transform CAST(NULL AS NULL) in CoralRelToSqlNodeConverter

aastha25:
Introduce the transformation `CAST(NULL AS NULL)` -> `NULL` in `CoralRelToSqlNodeConverter` to ensure a valid SqlNode is generated from the Coral RelNode. 

This transformation was previously applied as a SqlCall transformer in `SparkSqlRewritter`. However, to achieve parity between SqlNode generated from RelNode in [RelToTrinoConverter](https://github.com/linkedin/coral/commit/711268b04cff8057d7327f1c72c8d516461c006d#diff-a3a3daba44acc4e0b1208c73b4098f34d49eaecdfec6e29c8afbf289abf00a9aR132) and CoralRelToSqlNodeConverter, this transformation will now be applied in a previous step. 

Refer to test [testSelectNullAs()](https://github.com/linkedin/coral/blob/856ab6a71596968c13c6c652bdcfc160988a9389/coral-spark/src/test/java/com/linkedin/coral/spark/CoralSparkTest.java#L340) for example usage.

./gradlew build
./gradlew spotlessApply
i-tested against production views

### DIFF
--- a/coral-hive/src/main/java/com/linkedin/coral/transformers/CoralRelToSqlNodeConverter.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/transformers/CoralRelToSqlNodeConverter.java
@@ -23,6 +23,7 @@ import org.apache.calcite.rel.core.Values;
 import org.apache.calcite.rel.logical.LogicalTableFunctionScan;
 import org.apache.calcite.rel.rel2sql.RelToSqlConverter;
 import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.rex.RexCall;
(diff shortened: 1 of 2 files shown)


yiqiangin (commented):
Is it possible to add a unit test for this function?

aastha25:
> Is it possible to add a unit test for this function?

@yiqiangin corresponding unit test is here [testSelectNullAs()](https://github.com/linkedin/coral/blob/856ab6a71596968c13c6c652bdcfc160988a9389/coral-spark/src/test/java/com/linkedin/coral/spark/CoralSparkTest.java#L340) 

without this transformation, the translated SQL will have `CAST(NULL AS NULL)` which cannot be analyzed by spark engine.

PR #424, opened 2023-06-13
[Coral-Common] Introduce TypeDerivationUtil for enhanced RelDataType derivation

aastha25:

This PR ports RelDataType derivation logic to a separate util class, `TypeDerivationUtil`. It also adds an enhancement in the logic by introducing a sqlShuttle - `SqlNodePreprocessorForTypeDerivation`. This shuttle operates on SqlSelect SqlCalls such as, `SELECT * FROM t` and corrects the representation of this sqlCall's `selectList` when it is marked NULL.

./gradlew spotlessApply
./gradlew build
i-tested against production views for all translation paths


coral-common/src/main/java/com/linkedin/coral/common/transformers/SqlCallTransformer.java
```diff
@@ -5,37 +5,32 @@
...
+import com.linkedin.coral.common.utils.TypeDerivationUtil;
 
 
 /**
  * Abstract class for generic transformations on SqlCalls
  */
 public abstract class SqlCallTransformer {
-  private SqlValidator sqlValidator;
-  private final List<SqlSelect> topSelectNodes = new ArrayList<>();
+  private TypeDerivationUtil _typeDerivationUtil;
```
yiqiangin, line 26: nit: why the naming pattern staring with `_` is used here?

coral-common/src/main/java/com/linkedin/coral/common/utils/TypeDerivationUtil.java
```diff
@@ -0,0 +1,96 @@
...
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.sql.util.SqlShuttle;
+import org.apache.calcite.sql.validate.SqlValidator;
+
+
+/**
+ * This is a utility class which helps derive the RelDataType of a given SqlNode.
+ */
+public class TypeDerivationUtil {
+  private final SqlValidator _sqlValidator;
```
yiqiangin, line 27: nit: same question about the naming pattern starting with `_`

coral-common/src/main/java/com/linkedin/coral/common/transformers/SqlCallTransformer.java
aastha25: thanks for pointing it out. Fixed it by removing `_`

PR #438, opened 2023-07-19
[Coral-Trino] Cast `char` fields, if necessary, to `varchar` type in the set operation - `RelToTrinoConverter`

findinpath:

In case of dealing with Hive views which make use of set operation (e.g. UNION) ensure that the fields from the inner SELECT statements have the same type as the field types of the set operation.

This patch adds the method `com.linkedin.coral.trino.rel2trino.RelToTrinoConverter#setOpToSql` which is used by all set operations `UNION`, `INTERSECT`, `MINUS`.

The proactive casting makes sure that Trino doesn't encounter anymore coercion issues while translating Hive views containing set statements.

Relates to https://github.com/trinodb/trino/issues/18337

This patch has unit test coverage.

### DIFF
--- a/coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
+++ b/coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
@@ -193,7 +193,55 @@ public class HiveToTrinoConverterTest {
             + "FROM \"test\".\"duplicate_column_name_a\" AS \"duplicate_column_name_a\"\n"
             + "LEFT JOIN (SELECT TRIM(\"duplicate_column_name_b\".\"some_id\") AS \"SOME_ID\", CAST(TRIM(\"duplicate_column_name_b\".\"some_id\") AS VARCHAR(65536)) AS \"$f1\"\n"
             + "FROM \"test\".\"duplicate_column_name_b\" AS \"duplicate_column_name_b\") AS \"t\" ON \"duplicate_column_name_a\".\"some_id\" = \"t\".\"$f1\") AS \"t0\"\n"
-            + "WHERE \"t0\".\"some_id\" <> ''" } };
+            + "WHERE \"t0\".\"some_id\" <> ''" },
+
+        { "test", "view_char_different_size_in_union", "SELECT CAST(\"table_with_mixed_columns\".\"a_char1\" AS VARCHAR(255)) AS \"col\"\n"
+            + "FROM \"test\".\"table_with_mixed_columns\" AS \"table_with_mixed_columns\"\n" + "UNION ALL\n"
+            + "SELECT CAST(\"table_with_mixed_columns0\".\"a_char255\" AS VARCHAR(255)) AS \"col\"\n"
+            + "FROM \"test\".\"table_with_mixed_columns\" AS \"table_with_mixed_columns0\"" },
+
+        { "test", "view_cast_char_to_varchar", "SELECT CAST(\"table_with_mixed_columns\".\"a_char1\" AS VARCHAR(65535)) AS \"col\"\n"
+            + "FROM \"test\".\"table_with_mixed_columns\" AS \"table_with_mixed_columns\"" },
+
+        { "test", "view_cast_char_to_varchar_in_union", "SELECT CAST(\"table_with_mixed_columns\".\"a_char1\" AS VARCHAR(65535)) AS \"col\"\n"
+            + "FROM \"test\".\"table_with_mixed_columns\" AS \"table_with_mixed_columns\"\n" + "UNION ALL\n"
+            + "SELECT CAST(CASE WHEN \"table_with_mixed_columns0\".\"a_char1\" IS NOT NULL THEN \"table_with_mixed_columns0\".\"a_char1\" ELSE 'N' END AS VARCHAR(65535)) AS \"col\"\n"
+            + "FROM \"test\".\"table_with_mixed_columns\" AS \"table_with_mixed_columns0\"" },
+
+        { "test", "view_cast_char_to_varchar_in_union_flipped", "SELECT CAST(CASE WHEN \"table_with_mixed_columns\".\"a_char1\" IS NOT NULL THEN \"table_with_mixed_columns\".\"a_char1\" ELSE 'N' END AS VARCHAR(65535)) AS \"col\"\n"
+            + "FROM \"test\".\"table_with_mixed_columns\" AS \"table_with_mixed_columns\"\n" + "UNION ALL\n"
+            + "SELECT CAST(\"table_with_mixed_columns0\".\"a_char1\" AS VARCHAR(65535)) AS \"col\"\n"
+            + "FROM \"test\".\"table_with_mixed_columns\" AS \"table_with_mixed_columns0\"" },
+
+        { "test", "view_cast_char_to_varchar_with_other_fields_in_union", "SELECT CAST(\"table_with_mixed_columns\".\"a_char1\" AS VARCHAR(65535)) AS \"text\", \"table_with_mixed_columns\".\"a_boolean\" AS \"a_boolean\", \"table_with_mixed_columns\".\"a_smallint\" AS \"a_number\"\n"
+            + "FROM \"test\".\"table_with_mixed_columns\" AS \"table_with_mixed_columns\"\n" + "UNION ALL\n"
+            + "SELECT CAST(CASE WHEN \"table_with_mixed_columns0\".\"a_char1\" IS NOT NULL THEN \"table_with_mixed_columns0\".\"a_char1\" ELSE 'N' END AS VARCHAR(65535)) AS \"text\", \"table_with_mixed_columns0\".\"a_boolean\" AS \"a_boolean\", \"table_with_mixed_columns0\".\"a_integer\" AS \"a_number\"\n"
+            + "FROM \"test\".\"table_with_mixed_columns\" AS \"table_with_mixed_columns0\"" },
+
+        { "test", "view_char_and_null_in_union", "SELECT \"table_with_mixed_columns\".\"a_char1\" AS \"text\"\n"
+            + "FROM \"test\".\"table_with_mixed_columns\" AS \"table_with_mixed_columns\"\n" + "UNION ALL\n"
+            + "SELECT NULL AS \"text\"\n"
+            + "FROM \"test\".\"table_with_mixed_columns\" AS \"table_with_mixed_columns0\"" },
+
+        { "test", "view_different_numerical_types_in_union", "SELECT *\n" + "FROM (SELECT *\n"
+            + "FROM (SELECT \"table_with_mixed_columns\".\"a_tinyint\" AS \"a_number\", \"table_with_mixed_columns\".\"a_float\" AS \"a_float\"\n"
+            + "FROM \"test\".\"table_with_mixed_columns\" AS \"table_with_mixed_columns\"\n" + "UNION ALL\n"
+            + "SELECT \"table_with_mixed_columns0\".\"a_smallint\" AS \"a_number\", \"table_with_mixed_columns0\".\"a_float\" AS \"a_float\"\n"
+            + "FROM \"test\".\"table_with_mixed_columns\" AS \"table_with_mixed_columns0\") AS \"t1\"\n" + "UNION ALL\n"
+            + "SELECT \"table_with_mixed_columns1\".\"a_integer\" AS \"a_number\", \"table_with_mixed_columns1\".\"a_float\" AS \"a_float\"\n"
+            + "FROM \"test\".\"table_with_mixed_columns\" AS \"table_with_mixed_columns1\") AS \"t3\"\n" + "UNION ALL\n"
+            + "SELECT \"table_with_mixed_columns2\".\"a_bigint\" AS \"a_number\", \"table_with_mixed_columns2\".\"a_float\" AS \"a_float\"\n"
+            + "FROM \"test\".\"table_with_mixed_columns\" AS \"table_with_mixed_columns2\"" },
+
+        { "test", "view_union_no_casting", "SELECT \"table_with_mixed_columns\".\"a_tinyint\" AS \"a_tinyint\", \"table_with_mixed_columns\".\"a_smallint\" AS \"a_smallint\", \"table_with_mixed_columns\".\"a_integer\" AS \"a_integer\", \"table_with_mixed_columns\".\"a_bigint\" AS \"a_bigint\", \"table_with_mixed_columns\".\"a_float\" AS \"a_float\"\n"
+            + "FROM \"test\".\"table_with_mixed_columns\" AS \"table_with_mixed_columns\"\n" + "UNION ALL\n"
+            + "SELECT \"table_with_mixed_columns0\".\"a_tinyint\" AS \"a_tinyint\", \"table_with_mixed_columns0\".\"a_smallint\" AS \"a_smallint\", \"table_with_mixed_columns0\".\"a_integer\" AS \"a_integer\", \"table_with_mixed_columns0\".\"a_bigint\" AS \"a_bigint\", \"table_with_mixed_columns0\".\"a_float\" AS \"a_float\"\n"
+            + "FROM \"test\".\"table_with_mixed_columns\" AS \"table_with_mixed_columns0\"" },
+
+        { "test", "fuzzy_union_view_char_casting", "SELECT *\n"
+            + "FROM \"test\".\"table_with_string_column\" AS \"table_with_string_column\"\n" + "UNION ALL\n"
+            + "SELECT CAST(\"table_with_char32_column\".\"a\" AS VARCHAR(65536)) AS \"a\"\n"
+            + "FROM \"test\".\"table_with_char32_column\" AS \"table_with_char32_column\"" } };
   }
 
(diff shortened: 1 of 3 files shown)


coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
```diff
@@ -193,7 +193,37 @@ public Object[][] viewTestCasesProvider() {
             + "FROM \"test\".\"duplicate_column_name_a\" AS \"duplicate_column_name_a\"\n"
             + "LEFT JOIN (SELECT TRIM(\"duplicate_column_name_b\".\"some_id\") AS \"SOME_ID\", CAST(TRIM(\"duplicate_column_name_b\".\"some_id\") AS VARCHAR(65536)) AS \"$f1\"\n"
             + "FROM \"test\".\"duplicate_column_name_b\" AS \"duplicate_column_name_b\") AS \"t\" ON \"duplicate_column_name_a\".\"some_id\" = \"t\".\"$f1\") AS \"t0\"\n"
-            + "WHERE \"t0\".\"some_id\" <> ''" } };
+            + "WHERE \"t0\".\"some_id\" <> ''" },
+
+        { "test", "view_cast_char_to_varchar", "SELECT CAST(\"table_with_char_columns\".\"ccol1\" AS VARCHAR(65535)) AS \"col\"\n"
+            + "FROM \"test\".\"table_with_char_columns\" AS \"table_with_char_columns\"" },
+
+        { "test", "view_cast_char_to_varchar_in_union", "SELECT CAST(\"table_with_char_columns\".\"ccol1\" AS VARCHAR(65535)) AS \"col\"\n"
```
losipiuk, line 201: where does the `\"table_with_char_columns\".\"ccol1\"` come from? Is output of union aliased - I am probably blind - but I do not see it.

findinpath: See `TestUtils` changes.

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
```diff
@@ -193,7 +193,37 @@ public Object[][] viewTestCasesProvider() {
...
+            + "FROM \"test\".\"table_with_char_columns\" AS \"table_with_char_columns\"\n" + "UNION ALL\n"
+            + "SELECT CAST(CASE WHEN \"table_with_char_columns0\".\"ccol1\" IS NOT NULL THEN \"table_with_char_columns0\".\"ccol1\" ELSE 'N' END AS VARCHAR(65535)) AS \"col\"\n"
+            + "FROM \"test\".\"table_with_char_columns\" AS \"table_with_char_columns0\"" },
+
+        { "test", "view_cast_char_to_varchar_in_union_flipped", "SELECT CAST(CASE WHEN \"table_with_char_columns\".\"ccol1\" IS NOT NULL THEN \"table_with_char_columns\".\"ccol1\" ELSE 'N' END AS VARCHAR(65535)) AS \"col\"\n"
+            + "FROM \"test\".\"table_with_char_columns\" AS \"table_with_char_columns\"\n" + "UNION ALL\n"
+            + "SELECT CAST(\"table_with_char_columns0\".\"ccol1\" AS VARCHAR(65535)) AS \"col\"\n"
+            + "FROM \"test\".\"table_with_char_columns\" AS \"table_with_char_columns0\"" },
+
+        { "test", "view_cast_numerical_to_tinyint_in_union", "SELECT *\n" + "FROM (SELECT *\n"
```
findinpath, line 211: There is a discrepancy between the output type of the fields from `UNION` between Hive and Coral.


```
0: jdbc:hive2://localhost:10000/default> create table numtbl(a_tinyint tinyint,  a_bigint bigint);
No rows affected (0.27 seconds)
0: jdbc:hive2://localhost:10000/default> create view v_numbtbl as
. . . . . . . . . . . . . . . . . . . .> select a_tinyint num from numtbl
. . . . . . . . . . . . . . . . . . . .> union
. . . . . . . . . . . . . . . . . . . .> select a_bigint num from numtbl;
No rows affected (0.242 seconds)
0: jdbc:hive2://localhost:10000/default> insert into numtbl values (1, 100000);
INFO  : Compiling command(queryId=hive_20230720134207_ca29b2fb-05ce-4c6c-b5b5-3695f49c5825): insert into numtbl values (1, 100000)

0: jdbc:hive2://localhost:10000/default> select * from v_numbtbl;

INFO  : Completed executing command(queryId=hive_20230720134524_8c36a4b2-86c8-49c9-9562-8f707050020a); Time taken: 3.914 seconds
INFO  : OK
+----------------+
| v_numbtbl.num  |
+----------------+
| 1              |
| 100000         |
+----------------+
2 rows selected (4.077 seconds)


0: jdbc:hive2://localhost:10000/default> describe v_numbtbl;
| num       | bigint     |          |

```

```
trino> select * from hive.default.v_numbtbl;

Query 20230720_075750_00003_yt6dq, FAILED, 1 node
http://localhost:8080/ui/query.html?20230720_075750_00003_yt6dq
Splits: 14 total, 0 done (0.00%)
CPU Time: 0.0s total,   500 rows/s, 4.39KB/s, 66% active
Per Node: 0.0 parallelism,     0 rows/s,     1B/s
Parallelism: 0.0
Peak Memory: 3.61KB
11.98 [1 rows, 9B] [0 rows/s, 1B/s]

Query 20230720_075750_00003_yt6dq failed: Out of range for tinyint: 100000
io.trino.spi.TrinoException: Out of range for tinyint: 100000
	at io.trino.type.BigintOperators.castToTinyint(BigintOperators.java:182)
```

Coral interprets the output field to be `TINYINT` while Hive considers the output to be `BIGINT`.

aastha25:
@findinpath can you implement this transformation as a `SqlCallTransformer` in `HiveSqlNodeToCoralSqlNodeConverter`? For example: a `UnionSqlCallTransformer` where the relDataType of the union-ed branches are compared and if there's a mismatch, CAST is introduced.

wmoustafa:
> @findinpath can you implement this transformation as a `SqlCallTransformer` in `HiveSqlNodeToCoralSqlNodeConverter`? For example: a `UnionSqlCallTransformer` where the relDataType of the union-ed branches are compared and if there's a mismatch, CAST is introduced.

+1 since we are migrating all RelNode language specific transformations from the SqlNode layer, either on the Hive side or Trino side.

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
findinpath: @aastha25 , @wmoustafa do you have any hints why is there actually a difference in handling the outcome of the field types from UNION statements between Hive and Coral?

At the moment, because of this difference, adding the changes from this PR would actually cause a regression in translating Hive views in Trino.

aastha25: @findinpath I looked into this issue and these are my findings:
(1) Coral converts input hive SQL definition to Coral IR - RelNode representation. This RelNode carries the data type details for the fields. 
(2) A 'UNION' is converted into a `LogicalUnion` RelNode. The data type for this is derived in Calcite [here](https://github.com/apache/calcite/blob/bdfb17029f7e205f895dc3dfd0f37c8ff2520823/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactoryImpl.java#L218) 
(3) The logic in `leastRestrictiveStructuredType()` is selecting the overall data type using the data type of union's first branch (type0)
(4) Hence:
SQL: `SELECT  a_tinyint FROM table_with_numerical_columns UNION ALL SELECT  a_integer FROM table_with_numerical_columns` -----> tinyint
SQL: `SELECT  a_integer FROM table_with_numerical_columns UNION ALL SELECT  a_tinyint FROM table_with_numerical_columns` -----> integer

To fix this issue, we could add a `UnionSqlCallTransformer` in hive SQL -> Coral RelNode translation path. This transformer will detect the difference in the two union branches and introduce an explicit CAST operator in one of the branches. That way Calcite will be able to derive the correct datatype for the UNION.

aastha25: and trino leverages RelNode's relDataType to derive overall rowType [here](https://github.com/trinodb/trino/blob/f858552f59054deac9fee8015b9806926404d482/plugin/trino-hive/src/main/java/io/trino/plugin/hive/ViewReaderUtil.java#L230)

wmoustafa:
Just discussed with @aastha25 after testing the queries natively on Trino. It sounds the SQL text is correct, but the `UNION` operator type derivation in Calcite returns the type of the left operand as the type of the row. I think the correct solution is to extend the `UNION` operator and override the type derivation so that it does not always use the left operand's type but instead consider upcast cases like this. Coral implements a similar approach in `HiveUnnest` operator.

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
findinpath: > To fix this issue, we could add a UnionSqlCallTransformer in hive SQL -> Coral RelNode translation path.

> Coral implements a similar approach in HiveUncollect operator.

@aastha25 , @wmoustafa  I'm investigating this path in Coral. Thank you for your feedback.

wmoustafa: @findinpath Thanks. Both quoted lines represent different approaches. The former rewrites the SQL to add `CAST` and the latter keeps the `SQL` but adjusts the return type (i.e., type inference). The latter is the one referred to in [this comment](https://github.com/linkedin/coral/pull/438#issuecomment-1649042335). As mentioned, after discussing, it sounds more correct to go with the latter approach (i.e., keeping the SQL as is, but adjusting `UNION` type inference).

findinpath: I see your point for the use case

```
hive> create table tbl2(a_integer integer, a_bigint bigint);
hive> create view v_tbl2 as select a_integer as num from tbl2 union all select a_bigint as num from tbl2;
hive> describe v_tbl2

+-----------+------------+--------------------+
+-----------+------------+--------------------+
| num       | bigint     |                          |
+-----------+------------+--------------------+
```


When debugging in Trino I see that the row type when selecting from `v_tbl2` has the outcome
```
RecordType(INTEGER num)
```

So, yes, we'd need to help out Coral to return the right type.

In the case of the following scenario

```
hive> create table tbl(col char(1));
hive> create view v_tbl as select cast(col as varchar(65535)) active_flag from tbl  union all select coalesce(col, 'N') as active_flag from tbl;
hive> describe v_tbl;

+-----------------+------------+----------------+
|   col_name   | data_type  | comment  |
+-----------------+------------+----------------+
| active_flag    | string     |                     |
+-----------------+------------+----------------+
```

When debugging in Trino I see that the row type when selecting from `v_tbl` has the outcome

```
RecordType(VARCHAR(65535) active_flag)
```

What would be the "right" type to infer in such a case on Coral side?

findinpath: Slight update from my side - I discovered that the [use-case](https://github.com/trinodb/trino/issues/18337) that I was trying to solve is rather related to a Trino [limitation](https://github.com/trinodb/trino/issues/9031) related to  `varchar(n)` coercing to `char(n)` (which should actually be the other way around).
I'll be back eventually on this PR after addressing the Trino issue.

wmoustafa: Thanks @findinpath!

findinpath: I've analyzed with @findepi the situation at hand and we probably have two things we need to do on the Coral side:

- follow-up on the suggestion from @wmoustafa 
> Coral implements a similar approach in HiveUncollect operator.

- use explicit `CAST` solution only for the situation where we're dealing with `CHAR` column types when the outcome of the UNION is a `VARCHAR` type (to reduce the risk of back-fire for this trick) because the [limitation](https://github.com/trinodb/trino/issues/9031) from Trino is not going to be changed in the near future.


I'm continuing with this PR and try to come up with a solution.

findinpath: I did further investigate the [char & varchar union case](https://github.com/trinodb/trino/issues/18337) that i am supposed to be solving.
Fortunately enough, the method `org.apache.calcite.sql.type.SqlTypeFactoryImpl#leastRestrictiveSqlType` does handle correctly the upcasting when dealing with `char` and `varchar` bounded types.

In case of exact numeric types, the method doesn't seem to be checking the _limit_ of the types 

```
((BasicSqlType) type).getLimit(true, SqlTypeName.Limit.OVERFLOW, false)
```

, but only the precision

```
              if (type.getPrecision()
                  > resultType.getPrecision()) {
                resultType = type;
              }
```

reason why the first branch specified in the UNION of numeric types is considered the "least restrictive sql type".
This seems to be rather a Calcite related issue.

In the light of the above findings, I'll modify the PR to actually apply the explicit `CAST` only to `varchar` & `char`  `UNION` use cases. However, I'll highlight the inconsistency of handling `UNION` of numeric types to be eventually handled at a later time in Calcite.

findinpath: @aastha25  I tried the path of creating a specialized `UnionSqlCallTransformer` , but I stumbled into exceptions like the following:

```
java.lang.RuntimeException: Failed to derive the RelDataType for SqlNode: SELECT CAST(`table_with_mixed_columns`.`a_char1` AS VARCHAR(65535)) AS `text`, `table_with_mixed_columns`.`a_boolean`, `table_with_mixed_columns`.`a_smallint` AS `a_number`
FROM `test`.`table_with_mixed_columns` AS `table_with_mixed_columns` with topSqlNode: SELECT CAST(`table_with_mixed_columns`.`a_char1` AS VARCHAR(65535)) AS `text`, `table_with_mixed_columns`.`a_boolean`, `table_with_mixed_columns`.`a_smallint` AS `a_number`
FROM `test`.`table_with_mixed_columns` AS `table_with_mixed_columns`
	at com.linkedin.coral.common.utils.TypeDerivationUtil.getRelDataType(TypeDerivationUtil.java:87)
	at com.linkedin.coral.common.transformers.SqlCallTransformer.deriveRelDatatype(SqlCallTransformer.java:66)
	at com.linkedin.coral.trino.rel2trino.transformers.UnionSqlCallTransformer.transform(UnionSqlCallTransformer.java:50)

```

findinpath:
@aastha25  / @wmoustafa  AC . Could you please review again the PR?

findinpath:
@aastha25 , @wmoustafa  can we please proceed with the review for this PR?

wmoustafa:
@findinpath Thank you for filing #442. I still believe #442 is the way to go since:
* The issue is very specific to Trino so handling on the output side of the transformation is typically the expectation.
* Specifically since this issue is sort of an unexpected behavior in Trino, updating Coral IR to accommodate it can cause significant backward incompatibilities and in an unexpected manner.

@aastha25 mentioned she will be able to help you make more progress on #442 especially that she has seen the error before. So it could be a minor issue after all.

findinpath:
>  The issue is very specific to Trino

Yes, this is true. This is why the change I'm suggesting is done in `coral-trino` library.
The change in `RelToTrinoConverter` is adding the exceptional handling for handling `UNION` statements on which a resultset containing a `char` column is to mixed with a resultset containing a `varchar` column to cope with the https://github.com/trinodb/trino/issues/9031 limitation from Trino.

> @aastha25 mentioned she would be able to help you make more progress on https://github.com/linkedin/coral/pull/442

Sure, if there a possibility to work with `SqlCallTransformer` I don't have anything against following this path.
I've stumbled on doing the translation of `SELECT * FROM tablea UNION SELECT char_col FROM tableb`. I'm not sure how to go forward on this path.
The current PR (see the attached tests) can successfully handle such situations.

Also another point in going forward with https://github.com/linkedin/coral/pull/438 is the amount of code to be changed/maintained to address https://github.com/trinodb/trino/issues/9031 use-cases is quite straight-forward. For the `SqlCallTransformer` implementation we need much more code to be introduced in the library.

wmoustafa:
@findinpath, the place where you made the changes in this PR (438) is moving to `coral-common` soon to convege the "Coral IR" in one place. That is why we cannot make Trino-specific changes there.

The type derivation of `*` and expansion of `*` on the SqlNode layer is not supported now. I think it is possible to do, but we can get to it later. My suggestion is to proceed with #442 and get back to `*` soon as a follow up.

findinpath:
> `RelToTrinoConverter` is moving to coral-common soon to converge the "Coral IR" in one place. 

Closing the current PR, because it doesn't have chances to land on `master` .
Continuing work with #442

PR #442, opened 2023-08-10
[Coral-Trino] Cast char fields, if necessary, to varchar type in the set operation - `SqlCallTransformer`

findinpath:

In case of dealing with Hive views which make use of set operation (e.g. UNION) ensure that the `char` fields from the inner SELECT statements have the same type as the output field types of the set operation.

Due to wrong coercion between `varchar` and `char` in Trino, as described in https://github.com/trinodb/trino/issues/9031
, a work-around needs to be applied in case of translating Hive views which contain a UNION dealing with char and varchar types. The work-around consists in the explicit cast of the field having char type towards varchar type corresponding of the set operation output type.

This PR is a spin-off from https://github.com/linkedin/coral/pull/438 using a slighly different functionality to achieve a similar functionality.

From a technical design point of view, the present PR extends `SqlCallTransformer` to modify on the fly the select branch of the set operation which needs explicit casting for its char columns (which need to be changed to varchar type due to https://github.com/trinodb/trino/issues/9031 Trino limitation.

The patch has been tested with automatic unit tests.

### DIFF
--- a/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/UnionSqlCallTransformer.java
+++ b/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/UnionSqlCallTransformer.java
@@ -1,0 +1,180 @@
+/**
+ * Copyright 2023 LinkedIn Corporation. All rights reserved.
+ * Licensed under the BSD-2 Clause license.
+ * See LICENSE in the project root for license information.
+ */
+package com.linkedin.coral.trino.rel2trino.transformers;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.sql.SqlBasicTypeNameSpec;
+import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlCallBinding;
+import org.apache.calcite.sql.SqlDataTypeSpec;
+import org.apache.calcite.sql.SqlIdentifier;
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlNodeList;
+import org.apache.calcite.sql.SqlSelect;
+import org.apache.calcite.sql.SqlTypeNameSpec;
+import org.apache.calcite.sql.fun.SqlCastFunction;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.sql.validate.SqlValidator;
+import org.apache.calcite.sql.validate.SqlValidatorScope;
+
+import com.linkedin.coral.common.transformers.SqlCallTransformer;
+import com.linkedin.coral.common.utils.TypeDerivationUtil;
+
+import static org.apache.calcite.rel.rel2sql.SqlImplementor.POS;
+import static org.apache.calcite.sql.parser.SqlParserPos.ZERO;
+
+
+/**
+ * This transformer class is used to adapt expressions in
+ * set statements (e.g.: `UNION`, `INTERSECT`, `MINUS`)
+ * in case that the branches of the set statement contain fields in
+ * char family which have different types.
+ * The `char` fields which are differing from the expected `varchar` output
+ * of the set statement will be adapted through an explicit `CAST` to the `varchar` type.
+ *
+ * @see <a href="https://github.com/trinodb/trino/issues/9031">Change char/varchar coercion in Trino</a>
+ */
+public class UnionSqlCallTransformer extends SqlCallTransformer {
+
+  public UnionSqlCallTransformer(TypeDerivationUtil typeDerivationUtil) {
+    super(typeDerivationUtil);
+  }
+
+  @Override
+  protected boolean condition(SqlCall sqlCall) {
+    return sqlCall.getOperator().kind == SqlKind.UNION || sqlCall.getOperator().kind == SqlKind.INTERSECT
+        || sqlCall.getOperator().kind == SqlKind.MINUS;
+  }
+
+  @Override
+  protected SqlCall transform(SqlCall sqlCall) {
+    List<SqlNode> operandsList = sqlCall.getOperandList();
+    if (sqlCall.getOperandList().isEmpty()) {
+      return sqlCall;
+    }
+    Integer selectListSize = null;
+    // Ensure all the operand are SELECT nodes and they have the same number of expressions projected
+    for (SqlNode operand : operandsList) {
+      if (operand.getKind() == SqlKind.SELECT && ((SqlSelect) operand).getSelectList() != null) {
+        SqlSelect select = (SqlSelect) operand;
+        List<SqlNode> selectList = select.getSelectList().getList();
+        if (selectListSize == null) {
+          selectListSize = selectList.size();
+        } else if (selectListSize != selectList.size()) {
+          return sqlCall;
+        }
+      } else {
+        return sqlCall;
+      }
+    }
+
+    List<Optional<RelDataType>> leastRestrictiveSelectItemTypes = new ArrayList<>(selectListSize);
+    for (int i = 0; i < selectListSize; i++) {
+      List<RelDataType> selectItemTypes = new ArrayList<>();
+      boolean selectItemTypesDerived = true;
+      for (SqlNode operand : operandsList) {
+        SqlSelect select = (SqlSelect) operand;
+        List<SqlNode> selectList = select.getSelectList().getList();
+        SqlNode selectItem = selectList.get(i);
+        if (selectItem.getKind() == SqlKind.IDENTIFIER && ((SqlIdentifier) selectItem).isStar()) {
+          // The type derivation of * on the SqlNode layer is not supported now.
+          return sqlCall;
+        }
+
+        try {
+          selectItemTypes.add(deriveRelDatatype(selectItem));
+        } catch (RuntimeException e) {
+          // The type derivation may fail for complex expressions
+          selectItemTypesDerived = false;
+          break;
+        }
+      }
+
+      Optional<RelDataType> leastRestrictiveSelectItemType =
+          selectItemTypesDerived ? Optional.ofNullable(leastRestrictive(selectItemTypes)) : Optional.empty();
+      leastRestrictiveSelectItemTypes.add(leastRestrictiveSelectItemType);
+    }
+
+    boolean operandsUpdated = false;
+    for (SqlNode operand : operandsList) {
+      SqlSelect select = (SqlSelect) operand;
+      List<SqlNode> selectList = select.getSelectList().getList();
+      List<SqlNode> rewrittenSelectList = null;
+      for (int i = 0; i < selectList.size(); i++) {
+        SqlNode selectItem = selectList.get(i);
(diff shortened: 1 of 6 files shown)


coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/TestUtils.java
```diff
@@ -378,6 +378,42 @@ public static void initializeTablesAndViews(HiveConf conf) throws HiveException,
 
     run(driver, "CREATE TABLE test.table_with_binary_column (b binary)");
 
+    run(driver,
```
findinpath, line 381: Note to self: add a test where a `select *` is made between two tables:
- table one containing a char, and an int column 
- table two containing a varchar and an int column


`select * from table1 union all select * from table2` 

Not sure whether in this case the cast for the char column should probably kick and rewrite the `select *` to selecting all the columns by name from table1

findinpath: Already handled via the test query:

```
CREATE VIEW IF NOT EXISTS test.fuzzy_union_view_single_branch_evolved AS
SELECT * from test.tableB union all SELECT * from test.tableC
```

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/UnionSqlCallTransformer.java
```diff
@@ -0,0 +1,165 @@
...
+        for (int i = 0; i < selectNodes.size(); i++) {
+          SqlNode sqlNode = selectNodes.get(i);
+          if (sqlNode.getKind() == SqlKind.IDENTIFIER && ((SqlIdentifier) sqlNode).isStar()) {
+            // The type derivation of * on the SqlNode layer is not supported now.
+            return sqlCall;
+          }
+
+          Optional<RelDataType> selectNodeDataType;
+          try {
+            selectNodeDataType = Optional.of(deriveRelDatatype(sqlNode));
```
findinpath, line 83: @aastha25 in some cases while running the `HiveToTrinoConverterTest#testViews` I stumbled on `RuntimeException` issued from `deriveRelDataType(SqlNode)`.

```
java.lang.RuntimeException: Failed to derive the RelDataType for SqlNode: CAST(row("b"."b1") as row("b1" varchar)) AS `b` with topSqlNode: SELECT `tablef`.`a`, `generic_project`(`tablef`.`b`, 'b', 'struct<b1:string>') AS `b`
FROM `test`.`tablef` AS `tablef`
	at com.linkedin.coral.common.utils.TypeDerivationUtil.getRelDataType(TypeDerivationUtil.java:87)
	at com.linkedin.coral.common.transformers.SqlCallTransformer.deriveRelDatatype(SqlCallTransformer.java:68)
```

Is this expected?
I currently workaround this limitation by just working with column types for which the type can actually be derived.

Also side note - Why throwing the very generic `RuntimeException` in `TypeDerivationUtil` instead of specialized exceptions?

aastha25: @findinpath were you able to isolate the test case which is leading to the failure?

findinpath: Tests failing:

- `fuzzy_union_view_double_branch_evolved_different`
- `fuzzy_union_view_more_than_two_branches_evolved`
- `fuzzy_union_view_map_with_struct_value_evolved`
- `fuzzy_union_view_array_with_struct_value_evolved`
- `fuzzy_union_view_deeply_nested_struct_evolved`
- `fuzzy_union_view_deeply_nested_complex_struct_evolved`

wmoustafa: @ljfgem Can we fix the generic type derivation method and also throw the underlying error?

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/UnionSqlCallTransformer.java
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  @Override
+  protected SqlCall transform(SqlCall sqlCall) {
+    List<SqlNode> operandsList = sqlCall.getOperandList();
+    List<List<Optional<RelDataType>>> columnsTypesLists = new ArrayList<>();
+    for (SqlNode operand : operandsList) {
+      if (operand.getKind() == SqlKind.SELECT && ((SqlSelect) operand).getSelectList() != null) {
+        SqlSelect select = (SqlSelect) operand;
+        List<SqlNode> selectNodes = select.getSelectList().getList();
```
wmoustafa, line 67: Does it make sense to rename this to `selectList` and the individual elements to be `selectItem`?

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/UnionSqlCallTransformer.java
wmoustafa, line 83: `selectItemDataType`

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/UnionSqlCallTransformer.java
```diff
@@ -0,0 +1,165 @@
...
+          }
+          columnsTypesLists.get(i).add(selectNodeDataType);
+        }
+      } else {
+        return sqlCall;
+      }
+    }
+
+    List<SqlNode> updatedOperands = new ArrayList<>(sqlCall.getOperandList().size());
+    for (SqlNode operand : operandsList) {
```
wmoustafa, line 96: Any reason we iterate on the operandsList twice? Can we merge this loop and the above? I see below we cast to SqlSelect in the first loop we check if it is SqlSelect. In both we iterate on the selectNodes, etc. I think there is room for consolidation? PS. In any case let us reuse the naming conventions suggested above.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/UnionSqlCallTransformer.java
```diff
@@ -0,0 +1,165 @@
...
+        if (columnsTypesLists.get(i).stream().anyMatch(columnType -> !columnType.isPresent())) {
+          // Couldn't determine the type for all the expressions
+          continue;
+        }
+        RelDataType inferredColumnType =
+            leastRestrictive(columnsTypesLists.get(i).stream().map(Optional::get).collect(Collectors.toList()));
+
+        SqlNode selectNode = selectNodes.get(i);
+        RelDataType selectNodeDataType = deriveRelDatatype(selectNodes.get(i));
+        if (!selectNodeDataType.equals(inferredColumnType) && selectNodeDataType.getSqlTypeName() == SqlTypeName.CHAR
```
wmoustafa, line 114: is `!selectNodeDataType.equals(inferredColumnType)` redundant? The 2nd and 3rd conditions seem to imply that.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/UnionSqlCallTransformer.java
```diff
@@ -0,0 +1,165 @@
...
+
+      List<SqlNode> rewrittenSelectNodes = new ArrayList<>();
+      boolean useRewrittenSelectNodes = false;
+      for (int i = 0; i < selectNodes.size(); i++) {
+
+        if (columnsTypesLists.get(i).stream().anyMatch(columnType -> !columnType.isPresent())) {
+          // Couldn't determine the type for all the expressions
+          continue;
+        }
+        RelDataType inferredColumnType =
```
wmoustafa, line 109: All types are technically inferred. Did you mean `leastRestrictiveType`?

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/UnionSqlCallTransformer.java
```diff
@@ -0,0 +1,165 @@
...
+            leastRestrictive(columnsTypesLists.get(i).stream().map(Optional::get).collect(Collectors.toList()));
+
+        SqlNode selectNode = selectNodes.get(i);
+        RelDataType selectNodeDataType = deriveRelDatatype(selectNodes.get(i));
+        if (!selectNodeDataType.equals(inferredColumnType) && selectNodeDataType.getSqlTypeName() == SqlTypeName.CHAR
+            && inferredColumnType.getSqlTypeName() == SqlTypeName.VARCHAR) {
+          // Work-around for the Trino limitation in dealing UNION statements between `char` and `varchar`.
+          // See https://github.com/trinodb/trino/issues/9031
+          SqlNode rewrittenSelectNode = castNode(selectNode, inferredColumnType);
+          if (!useRewrittenSelectNodes) {
```
wmoustafa, line 119: I am not sure what this flag is used for. Could you clarify (maybe as inline comments) Let us clarify when it is set/reset and the logic of the if/else branches below.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/UnionSqlCallTransformer.java
```diff
@@ -0,0 +1,165 @@
...
+        @Override
+        public RelDataType deriveType(SqlValidator validator, SqlValidatorScope scope, SqlCall call) {
+          SqlCallBinding opBinding = new SqlCallBinding(validator, scope, call);
+          return inferReturnType(opBinding);
+        }
+      }.createCall(ZERO, node, getSqlDataTypeSpecForCasting(type));
+    }
+  }
+
+  private static SqlDataTypeSpec getSqlDataTypeSpecForCasting(RelDataType relDataType) {
```
wmoustafa, line 160: Seems there is no need for the `ForCasting` part since this is not specific to casing?

wmoustafa:
@KevinGe00 Can we run integration tests on this PR?

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/UnionSqlCallTransformer.java
findinpath: We need to find the least restrictive type for each select item index from the operands - this is one looping.

Once we have the least restrictive type, we need another looping to verify whether the operands need to be updated.

I didn't find a better alternative to do this.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/UnionSqlCallTransformer.java
findinpath: Algorithm sligthly updated.
The reasoning behind the new flag `operandsUpdated` should be obvious.

findinpath:
@wmoustafa addressed comments.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/UnionSqlCallTransformer.java
ljfgem: Sure, will take a look.

wmoustafa (approved):
Confirmed with @KevinGe00 that integration tests passed. Thanks @findinpath for working on this. Appreciate the spin-off from https://github.com/linkedin/coral/pull/438.

PR #444, opened 2023-08-30
[Coral-Spark] Introduce SqlNode representation in CoralSpark

aastha25:

This PR adds SqlNode as a class member for CoralSpark. This will enable downstream consumers to leverage the CoralSpark API to convert Coral IR to Spark compatible SqlNode representation and perform further transformations using SqlShuttles, as necessary. The final transformed SqlNode can be converted to SQL by utilizing `CoralSpark.constructSparkSQL(SqlNode)`.

./gradlew build

### DIFF
--- a/coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
+++ b/coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
@@ -42,8 +42,9 @@ public class CoralSpark {
 
   private final List<String> baseTables;
   private final List<SparkUDFInfo> sparkUDFInfoList;
-  private final String sparkSql;
   private final HiveMetastoreClient hiveMetastoreClient;
+  private final SqlNode sparkSqlNode;
+  private String sparkSql;
 
   @Deprecated
   private CoralSpark(List<String> baseTables, List<SparkUDFInfo> sparkUDFInfoList, String sparkSql) {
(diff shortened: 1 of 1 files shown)


coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
```diff
@@ -241,4 +266,20 @@ public List<SparkUDFInfo> getSparkUDFInfoList() {
...
   }
+
+  /**
+   * Getter for the SqlNode representation
+   *
+   * @return SqlNode
+   */
+  public SqlNode getSqlNode() {
+    return sqlNode;
+  }
```
wmoustafa, line 277: I do not think we should "set" the SqlNode. Can we avoid that?

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
```diff
@@ -241,4 +266,20 @@ public List<SparkUDFInfo> getSparkUDFInfoList() {
...
+
+  /**
+   * Setter for the Spark SQL
+   */
+  public void setSparkSql(String sparkSql) {
+    this.sparkSql = sparkSql;
```
wmoustafa, line 283: Passing Spark SQL in constructor as well as setter does not sound right. What is the reasoning? Can we avoid it?

aastha25: we dont need the setter with Spark-View-Catalog. Removed it

wmoustafa (approved):

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
```diff
@@ -42,23 +43,26 @@ public class CoralSpark {
 
   private final List<String> baseTables;
   private final List<SparkUDFInfo> sparkUDFInfoList;
-  private final String sparkSql;
   private final HiveMetastoreClient hiveMetastoreClient;
+  private final SqlNode sqlNode;
```
ljfgem, line 47: I think `sparkSqlNode` is more accurate?

aastha25: Agreed, but let's introduce LanguageSqlNode terminology after the Coral IR APIs are completed and published.

PR #446, opened 2023-09-01
[Coral-Hive] Loose UDF format requirement

ljfgem:

Previously, Coral requires `com.linkedin.xxx` UDFs to be in `databaseName_tableName_udfName` format, this PR is to loose this requirement, so that `udfName` can also be resolved. The change is backward compatible.

1. new UT
2. tested on Spark shell for the affected production view


ljfgem:
@wmoustafa @aastha25 @yiqiangin: PTAL

yiqiangin (approved):

wmoustafa (approved):

PR #447, opened 2023-09-05
[Coral-Trino] Support ARRAY() value in NAMED_STRUCT Coral IR Operator 

aastha25:
Prior to this PR, following type of Coral IR SqlCalls:
 `NAMED_STRUCT('value', ARRAY())` 
was translated incorrectly to Trino SQL. Previous output: 
`CAST(ROW(ARRAY[]) AS ROW('value' ARRAY<NULL>))` 

Trino requires a non-null datatype assigned explicitly to the column `value`. This PR makes `VARCHAR` that default assignment, which is same data type as the implicit type assigned in Coral IR. New translated trino SQL:
`CAST(ROW(ARRAY[]) AS ROW('value' ARRAY<VARCHAR(65535)>))`

./gradlew build
i-tested with production views
previously failing production view works with this change on trino


ljfgem (approved):
LGTM

PR #449, opened 2023-09-07
Make Coral IR friendly to missing table alias prefix

yiqiangin:

As Coral IR depends on the code of Calcite to generate `RelDataType` for each column, RelDataType of a column of struct type is always generated with StructKind.FULLY_QUALIFIED as shown in [RelDataTypeFactoryImpl](https://github.com/linkedin/linkedin-calcite/blob/li-1.21.0/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeFactoryImpl.java#L153). Therefore `SqlValidator` from Calcite always requires a fully qualified name of a field from a column of struct type like `tbl.structCol.field`. If the query defines a field without table alias prefix like `structCol.field`, `SqlValidator` takes `structCol` as the table name and throw an exeption with the message of `table structCol not found`.
In order to make Coral IR work without table alias prefix for a field from a column of struct type, the following changes are made:
- Adding a new class of `CoraJavaTypeFactoryImpl` which extends the class of `JavaTypeFactoryImpl` (which is a subclass of RelDataTypeFactoryImpl) and it overrides the function of `createStructField` with `StructKind.PEEK_FIELDS_NO_EXPAND`. Noted that the reason why `StructKind.PEEK_FIELDS_NO_EXPAND`is used not ``StructKind.PEEK_FIELDS` or `StructKind.PEEK_FIELDS_DEFAULT` is the other two options would make the translation of `select *` not work as expected.
- `RelDataTypeFactory` associated with `RelOptCluster` is created in the code path of `Frameworks.withPrepare(FrameworkConfig config,BasePrepareAction<R> action)` which are all the code from `Calcite` which is called in `RelBuilder.create()`, however we cannot change the code in Calcite to create an object of `CoralJavaTypeFactoryImpl` in this path. Therefore a new class of `HiveRelBuilder` is created to override the function of `RelBuilder.create()` where an object of `CoralJavaTypeFactoryImpl` and associated `RelOptCluster` is created to replace the object of `JavaTypeFactoryImpl` and associated `RelOptCluster` which are created in the code path of Calcite to create an object of `RelBuilder`. It is used by `HiveToRelConverter`.

- Adding some unit tests as `HiveToTrinoConverterTest.testSqlSelectAliasAppenderTransformerWithoutTableAliasPrefix`, `CoralSparkTest.testStructColProjectionWithoutTableAliasPrefix` and `CoralSparkTest.testStructColProjectionWithTableAliasPrefix` 
- Fixing some existing unit test cases
- ./gradlew clean build
- manual test on a view where the field does not have table alias prefix  in its View Expanded Text


coral-common/src/main/java/com/linkedin/coral/common/HiveCommonRelBuilder.java
```diff
@@ -0,0 +1,29 @@
...
+import org.apache.calcite.plan.Context;
+import org.apache.calcite.plan.RelOptCluster;
+import org.apache.calcite.plan.RelOptSchema;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.tools.FrameworkConfig;
+import org.apache.calcite.tools.Frameworks;
+import org.apache.calcite.tools.RelBuilder;
+
+
+public class HiveCommonRelBuilder extends RelBuilder {
```
aastha25, line 17: Please check if this override can be introduced in `HiveRelConverter` class and if we can use that [here ](https://github.com/linkedin/coral/pull/449/files#diff-525c66fa9dca629a4d819ec31869600a6d3559ae0287d1c572b93f7fd2aff785R173)

yiqiangin: Yes, I have merged the code into `HiveRelConverter` and move it to `coral-common` module

coral-common/src/main/java/com/linkedin/coral/common/CoralJavaTypeFactoryImpl.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.List;
+
+import org.apache.calcite.jdbc.JavaTypeFactoryImpl;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeSystem;
+import org.apache.calcite.rel.type.StructKind;
+
+
+public class CoralJavaTypeFactoryImpl extends JavaTypeFactoryImpl {
```
aastha25, line 16: Please add java doc to document why the override is being introduced.

coral-common/src/main/java/com/linkedin/coral/common/CoralJavaTypeFactoryImpl.java
```diff
@@ -0,0 +1,25 @@
...
+  public CoralJavaTypeFactoryImpl(RelDataTypeSystem typeSystem) {
+    super(typeSystem);
+  }
+
+  @Override
+  public RelDataType createStructType(final List<RelDataType> typeList, final List<String> fieldNameList) {
+    return createStructType(StructKind.PEEK_FIELDS_NO_EXPAND, typeList, fieldNameList);
```
aastha25, line 23: how did we choose between `PEEK_FIELDS` vs `PEEK_FIELDS_NO_EXPAND`?

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
```diff
@@ -848,4 +848,16 @@ public void testSqlSelectAliasAppenderTransformer() {
         + "WHERE \"tablea\".\"a\" > 5";
     assertEquals(expandedSql, expected);
   }
+
+  @Test
+  public void testSqlSelectAliasAppenderTransformerWithoutTableAliasPrefix() {
```
aastha25, line 853: can we add a similar test in `CoralSparkTest` as well?

aastha25 (commented):
thanks for the PR @yiqiangin

coral-common/src/main/java/com/linkedin/coral/common/CoralJavaTypeFactoryImpl.java
aastha25: lets also capture that in the java doc with examples.

yiqiangin: I tried to use all three options `PEEK_FIELDS`, `PEEK_FIELDS_DEFAULT` and `PEEK_FIELDS_NO_EXPAND`. The use of  `PEEK_FIELDS` and `PEEK_FIELDS_DEFAULT` would result in the failure of translation of `select *`.  It should be still translated as `select *` in the target SQL, however, it is translated as `select tbl.col1 as col1 .... tbl.coln as coln` when those two options are used. So only `PEEK_FIELDS_NO_EXPAND` is applicable

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
yiqiangin: Done.

coral-common/src/main/java/com/linkedin/coral/common/CoralJavaTypeFactoryImpl.java
yiqiangin: Done

rzhang10 (approved):

PR #460, opened 2023-10-05
[Coral-Trino] Remove characters sets from TrinoSqlDialect

KevinGe00:
Previously, `charSetName` was dropped from VARCHAR/CHAR fields when a Coral RelNode is being converted to a Trino SqlNode. This happened by overriding `supportsCharSet` to be false in `TrinoSqlDialect`.  As art of merging `TrinoSqlDialect` -> `CoralSqlDialect`, we want `TrinoSqlDialect` to match `CoralSqlDialect` meaning the override needed to be undone.

Now, `charSetName` is instead dropped from VARCHAR/CHAR fields dropped as part of SqlNode rewriting in `TrinoSqlRewriter` by adding a datatype replacement rule.

1. clean build
2. new unit test to test character sets are removed in `CAST` calls that are nested
3. i-test passed

### DIFF
--- a/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/TrinoSqlDialect.java
+++ b/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/TrinoSqlDialect.java
@@ -25,11 +25,6 @@ public class TrinoSqlDialect extends SqlDialect {
     super(context);
   }
 
-  @Override
-  public boolean supportsCharSet() {
-    return false;
-  }
-
   /**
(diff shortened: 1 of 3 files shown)


aastha25:
thanks for the PR @KevinGe00!
Please add an example in the description elaborating the previous and newer workflow - that charSetName was dropped from VARCHAR/CHAR  fields when Coral RelNode converted to Trino SqlNode. Now it is dropped as part of SqlNode rewriting.

aastha25 (approved):

PR #462, opened 2023-10-11
Register UDF of RedactFieldIf in Coral

yiqiangin:

Register the UDF named `RedactFieldIf` in Coral. The definition of the UDF is as follows:
```
name = "REDACT_FIELD_IF"
value = "_FUNC_(condition:Boolean, input: Object, path: String, devalutValue: Object)
"Returns the input with the element at path redacted if condition is true, else returns the input unmodified"
```

- `./gradlew clean build`
- `the view with this UDF can be translated successfully in coral-tools`


ljfgem (approved):
Let's merge if the view can be translated with this PR.

aastha25 (approved):
Have we verified that the UDF couldnt be resolved dynamically.

yiqiangin:
> Have we verified that the UDF couldnt be resolved dynamically.

Yes, this UDF cannnot be resolved dynamically the same as the UDF of `HasMemberConsent` which was registered before.

Issue #459, opened 2023-10-05
Bug: `FromUnixtimeOperatorTransformer` possibly built on false supposisions.


findinpath:
The `FromUnixtimeOperatorTransformer` from https://github.com/linkedin/coral/pull/426/files#diff-e26e325f65fb1f3341468a92d2e1051ee2b8b94b56c74203054aca3b2c790e4e
seems to be built on false premises.

The function https://trino.io/docs/current/functions/datetime.html#from_unixtime returns a `timestamp(3) with time zone`

However, the `FromUnixtimeOperatorTransformer` transforms `from_unixtime` expressions to
`FORMAT_DATETIME(FROM_UNIXTIME(10000), 'yyyy-MM-dd HH:mm:ss')` which is a varchar
This is causing the failure of statements like the following:

```
select 
  CAST(
    CAST(
      "at_timezone" (
        "format_datetime" (
          "from_unixtime" (
            CAST(
              "test_from_utc_timestamp_source"."source_float" AS DOUBLE
            )
          ),
          'yyyy-MM-dd HH:mm:ss'
        ),
        "$canonicalize_hive_timezone_id" ('America/Los_Angeles')
      ) AS TIMESTAMP (3)
    ) AS VARCHAR (65535)
  ) AS "ts_float"
from test_from_utc_timestamp_source 
```
with the exception:
```
Caused by: io.trino.spi.TrinoException: Unexpected parameters (varchar, varchar) for function at_timezone. Expected: at_timezone(timestamp(p) with time zone, interval day to second), at_timezone(timestamp(p) with time zone, varchar(x))
```

Related PR: https://github.com/linkedin/coral/pull/426

findinpath:
Steps to reproduce the issue

hive
```
CREATE TABLE test_from_utc_timestamp_source (source_float float);
CREATE VIEW test_from_utc_timestamp_view AS SELECT CAST(from_utc_timestamp(source_float, 'America/Los_Angeles') AS STRING) ts_float FROM test_from_utc_timestamp_source;

select * from test_from_utc_timestamp_view;

-- 1970-01-30 21:30:00
```

trino
```
insert into hive.default.test_from_utc_timestamp_source values (2592000.0);
```

trino version 420 while using coral version 2.2.9
```
SELECT * FROM hive.default.test_from_utc_timestamp_view;
[...]

PR #464, opened 2023-10-15
[Coral-Trino] Avoid accidental translation of Trino `from_unixtime` SQL call

findinpath:
### What changes are proposed in this pull request, and why are they necessary?

Avoid the accidental translation of the Trino `from_unixtime` introduced through #426 

Context https://github.com/linkedin/coral/issues/459#issuecomment-1762160693

Fixes https://github.com/linkedin/coral/issues/459

### How was this patch tested?

Regular unit tests

PR #426, opened 2023-06-14
[Coral-Trino] Migrate FROM_UNIXTIME and FROM_UTC_TIMESTAMP

aastha25:
Before this PR, `FROM_UNIXTIME` and `FROM_UTC_TIMESTAMP`operators were transformed to trino engine compatible operators in the RelShuttle `Calcite2TrinoUDFConverter`.
This PR migrates the transformation to SqlShuttles  - `CoralToTrinoSqlCallConverter` and `DataTypeDerivedSqlCallConverter`  respectively.

### How was this patch tested?

./gradlew spotlessApply
./gradlew build
modified UTs
i-tested against production views

PR #467, opened 2023-10-20
[Coral-Trino] Avoid accidental translation of Trino `from_unixtime` SQL call - SqlFunction implementation

findinpath:

Avoid the accidental translation of the Trino from_unixtime introduced through https://github.com/linkedin/coral/pull/426

Context https://github.com/linkedin/coral/issues/459#issuecomment-1762160693

Fixes https://github.com/linkedin/coral/issues/459

Spin-off from https://github.com/linkedin/coral/pull/464 to showcase the solution based on `TimestampFromUnixtime` `SqlOperator` implementation.

Integration tests


wmoustafa:
I have made a small fix in `TrinoSqlDialect` that fixes 2 text cases. The third case fails because `from_unixtime_nanos`, which I think needs a similar fix?

wmoustafa:
Fixed the remaining test case.

wmoustafa (approved):

PR #473, opened 2023-11-06
[Coral-Spark] Modify coral hive parser and coral spark writer dialect to generate spark compliant escaped string literal

rzhang10:

Without this patch, a string literal of `select 'I'm'` will be re-written in spark-sql output of `select I''m`, the `''` double single quote escaping is a ANSI sql escaping standard but is not what is recognized by the spark sql parser. Instead, spark uses the backslash for escaping quotes inside quotes in sql string, i.e. the correct output should be `select 'I\'m'`.

This patch adds the customizations in the `SparkSqlDialect` to make the generated output sql consistency use single quote for quoting a outermost level string literal and use `\'` to quote internal literals occurences of `'`, for `"`, it will appear as-is since it will never need to be escaped because the outermost quote is always single quote.

Added 6 unit tests to demonstrate the new behavior:
| hive | spark |
| --- | --- |
| select 'abc' | select 'abc' |
| select "abc" | select 'abc' |
| select 'abc[\\'xyz\\']' | select 'abc[\\'xyz\\']' |
| select "abc['xyz']" | select 'abc[\\'xyz\\']' |
| select 'abc["xyz"]' | select 'abc[\\'xyz\\']' |
| select "abc[\\"xyz\\"]" | select 'abc["xyz"]' |

Run regression tests against all views and results looks fine. Also tested against the view sql which exposed this issue on spark, the spark read result literals are correct now.

### DIFF
--- a/coral-spark/src/main/java/com/linkedin/coral/spark/dialect/SparkSqlDialect.java
+++ b/coral-spark/src/main/java/com/linkedin/coral/spark/dialect/SparkSqlDialect.java
@@ -1,10 +1,10 @@
 /**
- * Copyright 2018-2023 LinkedIn Corporation. All rights reserved.
- * Licensed under the BSD-2 Clause license.
- * See LICENSE in the project root for license information.
+ * Copyright 2018-2023 LinkedIn Corporation. All rights reserved. Licensed under the BSD-2 Clause
+ * license. See LICENSE in the project root for license information.
  */
 package com.linkedin.coral.spark.dialect;
 
+import org.apache.calcite.avatica.util.Casing;
 import org.apache.calcite.config.NullCollation;
 import org.apache.calcite.sql.SqlCall;
 import org.apache.calcite.sql.SqlDialect;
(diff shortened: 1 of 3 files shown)


ljfgem:
Thanks Raymond for this pr, could you add regression tests for both spark and trino? And could you also test it on spark shell against the target view to ensure it works as expected?

rzhang10:
> Thanks Raymond for this pr, could you add regression tests for both spark and trino? And could you also test it on spark shell against the target view to ensure it works as expected?

Yeah I just triggered int-test for both trino and spark, let's wait for the restults.

coral-spark/src/main/java/com/linkedin/coral/spark/dialect/SparkSqlDialect.java
```diff
@@ -30,8 +31,12 @@
  */
 public class SparkSqlDialect extends SqlDialect {
 
-  public static final SparkSqlDialect INSTANCE = new SparkSqlDialect(
-      emptyContext().withDatabaseProduct(DatabaseProduct.HIVE).withNullCollation(NullCollation.HIGH));
+  public static final SqlDialect.Context DEFAULT_CONTEXT =
+      SqlDialect.EMPTY_CONTEXT.withDatabaseProduct(DatabaseProduct.SPARK).withLiteralQuoteString("'")
+          .withLiteralEscapedQuoteString("\\'").withNullCollation(NullCollation.LOW)
+          .withUnquotedCasing(Casing.UNCHANGED).withQuotedCasing(Casing.UNCHANGED).withCaseSensitive(false);
```
ljfgem, line 37: In previous impl, we are using `NullCollation.HIGH`,  `Casing.TO_UPPER` for `withUnquotedCasing` and `true` for `withCaseSensitive`, why are we changing here?

ljfgem (commented):
Thank you @rzhang10 for this PR, left one comment

coral-spark/src/main/java/com/linkedin/coral/spark/dialect/SparkSqlDialect.java
rzhang10: I think they are more correct for spark-sql? And seems there's no regressions for this change so I think it's fine to add them?

see ref: https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql/dialect/SparkSqlDialect.java#L47

https://community.databricks.com/t5/data-engineering/spark-is-case-sensitive-spark-is-not-case-sensitive-by-default/td-p/13954#:~:text=is%20case%20sensitive%3F-,Spark%20is%20not%20case%20sensitive%20by%20default.,will%20get%20column%20ambiguity%20error.

ljfgem (approved):
LGTM, thank you @rzhang10 for the quick fix!

coral-spark/src/main/java/com/linkedin/coral/spark/dialect/SparkSqlDialect.java
rzhang10: I think these configs are mostly used when using calcite parser to parser a sql dialect, we are not using them here, so basically they are not effective at all.

Maybe later when we want to use calcite to parse spark-sql in coral, they can become useful.

PR #474, opened 2023-11-07
[Coral-Spark] Add quoting for reserved keywords

rzhang10:

This PR adds quoting for reserved keywords to the generated spark-sql string from coral-spark, using the quoting character "`". The complete list of sql keywords are defined here: https://cwiki.apache.org/confluence/display/hive/languagemanual+ddl#LanguageManualDDL-Keywords,Non-reservedKeywordsandReservedKeywords

This is implemented via overriding the `identifierNeedsQuote` method from SqlDialect class, and calcite internally will hook this into its sql writer to write quoted identifier if it contains a reserved keyword.

tested via existing unit tests (with modifying 3 unit tests to have backticks since they have keywords in the statements), and the added unit test `testQuotingKeywords`


ljfgem (approved):
lgtm, let's merge once we verify that the affected view can be queried with this PR.

PR #481, opened 2023-12-06
Introduce OrdinalReturnTypeInferenceV2 to infer RexCall's return type

aastha25:

Given a SQL definition - `SELECT udf(StructCol) AS modCol FROM foo` 

Prior to this PR, Coral-Schema generates a nullable schema for `modCol` with top level and inner fields set to nullable even when `StructCol` is non-nullable. Since `udf`'s return type nullability inference is unavailable to `RelToAvroSchemaConverter`, this shuttle assumes the schema after applying th UDF is nullable.

This PR introduces a new `SqlReturnTypeInference` - `OrdinalReturnTypeInferenceV2`which has been leveraged in `RelToAvroSchemaConverter` to derive udf - `cast_nullability`'s return type schema (and its nullability info) more accurately. This inference strategy has been applied when the SqlCall's return type is same as one of the operands type AND that operand refers to a field from the input schema.

./gradlew build
added new UTs.


coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
```diff
@@ -667,6 +667,8 @@ public boolean isOptional(int i) {
         family(SqlTypeFamily.STRING, SqlTypeFamily.ANY, SqlTypeFamily.TIMESTAMP));
     createAddUserDefinedFunction("com.linkedin.policy.decoration.udfs.RedactFieldIf", ARG1,
         family(SqlTypeFamily.BOOLEAN, SqlTypeFamily.ANY, SqlTypeFamily.STRING, SqlTypeFamily.ANY));
+    createAddUserDefinedFunction("li_groot_cast_nullability", new OrdinalReturnTypeInferenceV2(1),
+        family(SqlTypeFamily.ANY, SqlTypeFamily.ANY));
```
ljfgem, line 671: Why do we need 2 input parameters for this function? Asking because in the unit test, two parameters are same.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
```diff
@@ -667,6 +667,8 @@ public boolean isOptional(int i) {
         family(SqlTypeFamily.STRING, SqlTypeFamily.ANY, SqlTypeFamily.TIMESTAMP));
     createAddUserDefinedFunction("com.linkedin.policy.decoration.udfs.RedactFieldIf", ARG1,
         family(SqlTypeFamily.BOOLEAN, SqlTypeFamily.ANY, SqlTypeFamily.STRING, SqlTypeFamily.ANY));
+    createAddUserDefinedFunction("li_groot_cast_nullability", new OrdinalReturnTypeInferenceV2(1),
```
ljfgem, line 670: Just to confirm, looks like user needs to modify their view definition to use this udf?

ljfgem (commented):
Thanks @aastha25 for this PR!
Could you kick off a regression test for it for safety?
And have we tested on the target view? Could you add the test results in the testing done?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
aastha25: yes, the affected user would need to define a new view which would contain this UDF. The new view isnt published yet

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
aastha25: `li_groot_cast_nullability` will take 2 args which may / maynot be the same and can have any type. The idea is to cast the nullability of ARG2 onto ARG1 where ARG1 can be same as ARG2 or a processed version of ARG2 such as udf(ARG2).

aastha25:
@ljfgem 

> Thanks @aastha25 for this PR! Could you kick off a regression test for it for safety? 
This is a new code path and only targets the UDF schema generated for the newly defined SqlReturnTypeInference.  

> And have we tested on the target view? Could you add the test results in the testing done?
The target view is unavailable. Hence the UDF has been tested within the realms of Coral.

ljfgem:
@aastha25 Thanks for confirmation.
Could we collaborate with the user to produce a test view and ensure this PR works for it on Spark shell?

aastha25:
> @aastha25 Thanks for confirmation. Could we collaborate with the user to produce a test view and ensure this PR works for it on Spark shell?

unfortunately the view will be created & tested by the user next week, after Coral & Spark enable support for it. In case of any issues, we'll need to work on a potential follow up PR.

ljfgem (approved):
LGTM

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
```diff
@@ -442,6 +436,22 @@ public RexNode visitCall(RexCall rexCall) {
...
+        RexNode operand = rexCall.operands.get(index);
+
+        if (operand instanceof RexInputRef) {
+          appendRexInputRefField((RexInputRef) operand);
+          return rexCall;
+        }
+      }
+
       RelDataType fieldType = rexCall.getType();
       boolean isNullable = SchemaUtilities.isFieldNullable(rexCall, inputSchema);
```
wmoustafa, line 456: Is this the reason `OrdinalReturnTypeInference` does not work out of the box?

aastha25: `[OrdinalReturnTypeInference](https://github.com/linkedin/linkedin-calcite/blob/li-1.21.0/core/src/main/java/org/apache/calcite/sql/type/OrdinalReturnTypeInference.java#L25)` cannot be directly used because the private field 'ordinal' has no public accessor method. This class supports type derivation via method `RelDataType inferReturnType(SqlOperatorBinding opBinding)`. To leverage this API, we need to make coral-schema complaint with Coral IR and enable type derivation in coral-schema, similar to our ongoing work in coral-spark & coral-trino

wmoustafa: So once we append a field, what gets used at the end to infer its type if not `inferReturnType`?

aastha25: we append the field AFTER extracting the ordinal [here](https://github.com/linkedin/coral/pull/481/files#diff-a2e05f91f268adfd9edbe9dda395197fad075217aaac529dbee5d794a8f6866cR446) and deriving its field type [here](https://github.com/linkedin/coral/pull/481/files#diff-a2e05f91f268adfd9edbe9dda395197fad075217aaac529dbee5d794a8f6866cR559) 

Since this is a rexShuttle, the field types from the inputSchema are available in the RelNode representation. 

`SchemaUtilities.isFieldNullable(rexCall, inputSchema)` derives incorrect nullability (always nullable) for the top level fields and `SchemaUtilities.makeNullable` [here](https://github.com/linkedin/coral/pull/481/files#diff-af50d4a6ecbf662497acd9da8f31d6fb50300e833ddad2716ec6814a9732d243R89) derives incorrect nullability (always nullable) for the inner fields. 

By introducing the `if` condition above in line 445, we avoid those codepaths

PR #483, opened 2024-01-09
[Coral-Trino] Modify transformation for CURRENT_TIMESTAMP

ljfgem:

In Hive, `CURRENT_TIMESTAMP` returns a `UTC` timestamp, whereas in Trino, it returns a timestamp based on the session's timezone. Therefore, to be compatible with Hive query result, `AT TIME ZONE 'UTC'` should be added to the translated Trino SQL. For example, `SELECT CURRENT_TIMESTAMP` is transformed into `SELECT CAST(CURRENT_TIMESTAMP AT TIME ZONE 'UTC' AS TIMESTAMP(3))`.

Modified unit tests.


wmoustafa:
Could you evaluate the Spark behavior in case Spark is used on the LHS? What triggered this issue? Is it still required? FYI @phd3.

PR #484, opened 2024-01-10
[Coral-Schema] Add DATE type support in RelDataTypeToAvroType

ljfgem:

This patch added the `DATE` type support in `RelDataTypeToAvroType` to prevent the exception `java.lang.UnsupportedOperationException: DATE is not supported.`

1. Unit test
2. Tested on the affected production view, which could be translated well with this PR
3. Tested on Spark shell, the affected production view can be queried well with this PR

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
@@ -115,8 +115,7 @@ class RelDataTypeToAvroType {
         schema.addProp("logicalType", "timestamp-millis");
         return schema;
       case DATE:
-        // Spark recognizes the data type of {"type": "int", "logicalType": "date"} as a date type:
-        // https://github.com/apache/spark/blob/master/connector/avro/src/main/scala/org/apache/spark/sql/avro/AvroDeserializer.scala#L145
+        // In Avro, "date" type is represented as {"type": "int", "logicalType": "date"}.
         schema = Schema.create(Schema.Type.INT);
         schema.addProp("logicalType", "date");
         return schema;
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2019-2023 LinkedIn Corporation. All rights reserved.
+ * Copyright 2019-2024 LinkedIn Corporation. All rights reserved.
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
@@ -90,6 +90,7 @@ class RelDataTypeToAvroType {
   }
 
   private static Schema basicSqlTypeToAvroType(BasicSqlType relDataType) {
+    Schema schema;
     switch (relDataType.getSqlTypeName()) {
       case BOOLEAN:
         return Schema.create(Schema.Type.BOOLEAN);
@@ -110,8 +111,12 @@ class RelDataTypeToAvroType {
       case NULL:
         return Schema.create(Schema.Type.NULL);
       case TIMESTAMP:
-        Schema schema = Schema.create(Schema.Type.LONG);
+        schema = Schema.create(Schema.Type.LONG);
         schema.addProp("logicalType", "timestamp-millis");
+        return schema;
+      case DATE:
+        schema = Schema.create(Schema.Type.INT);
+        schema.addProp("logicalType", "date");
         return schema;
       case DECIMAL:
         JsonNodeFactory factory = JsonNodeFactory.instance;
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
@@ -115,6 +115,8 @@ class RelDataTypeToAvroType {
         schema.addProp("logicalType", "timestamp-millis");
         return schema;
       case DATE:
+        // Spark recognizes the data type of {"type": "int", "logicalType": "date"} as a date type:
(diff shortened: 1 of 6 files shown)


coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
```diff
@@ -110,9 +111,13 @@ private static Schema basicSqlTypeToAvroType(BasicSqlType relDataType) {
       case NULL:
         return Schema.create(Schema.Type.NULL);
       case TIMESTAMP:
-        Schema schema = Schema.create(Schema.Type.LONG);
+        schema = Schema.create(Schema.Type.LONG);
         schema.addProp("logicalType", "timestamp-millis");
         return schema;
+      case DATE:
+        schema = Schema.create(Schema.Type.INT);
```
aastha25, line 118: Since there's no `Date` type schema in avro https://github.com/apache/avro/blob/branch-1.10/lang/java/avro/src/main/java/org/apache/avro/Schema.java#L125, how did we choose 'int' here? 

What is the default spark derived schema for 'Date' type field?

coral-schema/src/test/resources/base-date-type-field.avsc
```diff
@@ -0,0 +1,12 @@
+{
+  "type" : "record",
+  "name" : "basedatetypefield",
+  "namespace" : "coral.schema.avro.base.date.type.field",
+  "fields" : [ {
+    "name" : "Date_Field",
+    "type" : [ "null", {
+      "type" : "int",
+      "logicalType" : "date"
+    } ]
```
aastha25, line 10: is this the hive schema representation of 'Date' data type column? who injects the 'logicalType'? why is this not just 
```
{
          "name" : "Date_Field",
          "type" : [ "null", "date" ]
        },
```

aastha25 (commented):
thanks for the PR!

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
```diff
@@ -110,9 +111,13 @@ private static Schema basicSqlTypeToAvroType(BasicSqlType relDataType) {
       case NULL:
         return Schema.create(Schema.Type.NULL);
       case TIMESTAMP:
-        Schema schema = Schema.create(Schema.Type.LONG);
+        schema = Schema.create(Schema.Type.LONG);
         schema.addProp("logicalType", "timestamp-millis");
         return schema;
+      case DATE:
```
aastha25, line 117: can we also add an inline comment here documenting the reason for this transformation?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
ljfgem: https://github.com/apache/spark/blob/master/connector/avro/src/main/scala/org/apache/spark/sql/avro/AvroDeserializer.scala#L145
Spark regards this avro schema as `date` type.

coral-schema/src/test/resources/base-date-type-field.avsc
ljfgem: This is Avro schema representing `date` type, we don't have `"type" : [ "null", "date" ]` in Avro.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
ljfgem: Added

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
aastha25: so spark derived `date` type is cast-able to coral derived 'int' type?

aastha25:
have we validated that this approach works with and without viewshift for the impacted dataset?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
ljfgem: No, Spark AvroDeserializer would return `date` type directly for this case, and `spark.table.printSchema` also returns `date` type.

ljfgem:
> have we validated that this approach works with and without viewshift for the impacted dataset?

@aastha25 Tested it on the viewshift replacement view, which couldn't be translated & queried without this PR, but can be translated & queried with this PR.

aastha25:
> @aastha25 Tested it on the target viewshift replacement view, which couldn't be translated & queried without this PR, but can be translated & queried with this PR.

It'll be good to test the top level view (which invokes this nested viewshift replacement view), to make sure that the schema generated for viewshift view (after comparing spark inferred schema & coral provided schema) is compatible with the top level view.

aastha25 (approved):

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java
```diff
@@ -110,9 +111,15 @@ private static Schema basicSqlTypeToAvroType(BasicSqlType relDataType) {
...
+        // Spark recognizes the data type of {"type": "int", "logicalType": "date"} as a date type:
+        // https://github.com/apache/spark/blob/master/connector/avro/src/main/scala/org/apache/spark/sql/avro/AvroDeserializer.scala#L145
```
wmoustafa, line 119: Why is the Spark comment relevant? 
Especially, the code below converts a Calcite Date Type to Avro Date logical type.

ljfgem: In our use case, Avro schema needs to be deserializer by Spark for upcast, but agreed that the Spark-related comment should not be added here since Coral-Schema is not just used for Spark.

PR #494, opened 2024-03-01
[Coral-Hive] Create new object for ShiftArrayIndexTransformer

ljfgem:

Prior to this PR, `ShiftArrayIndexTransformer` modifies the index of the SqlCall directly, which might transform the same object multiple times if it appears multiple times in SqlNode and generate the wrong index. This PR fixes this issue by creating new object for ShiftArrayIndexTransformer.

1. Tested on the affected prod view, worked as expected
2. Regression test on all prod views, no regression
3. Note: I couldn't reproduce the issue by UT somehow, still investigating the reason. If it still can't be reproduced by UT locally after spending much time, the PR can be merged and we can make UT as a follow-up work.

### DIFF
--- a/coral-hive/src/main/java/com/linkedin/coral/transformers/ShiftArrayIndexTransformer.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/transformers/ShiftArrayIndexTransformer.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2022-2023 LinkedIn Corporation. All rights reserved.
+ * Copyright 2022-2024 LinkedIn Corporation. All rights reserved.
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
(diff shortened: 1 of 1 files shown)


coral-hive/src/main/java/com/linkedin/coral/transformers/ShiftArrayIndexTransformer.java
```diff
@@ -42,16 +42,18 @@ public boolean condition(SqlCall sqlCall) {
   @Override
   public SqlCall transform(SqlCall sqlCall) {
     final SqlNode itemNode = sqlCall.getOperandList().get(1);
+    SqlNode newIndex;
```
aastha25, line 45: could you add a todo for the unit test here so that the pending item is tracked?

wmoustafa:
Could you add a test that fails when this patch is not applied?

ljfgem:
> Could you add a test that fails when this patch is not applied?

@wmoustafa As I mentioned in the description, I couldn't reproduce the issue by UT somehow, still investigating the reason. I have added a TODO item in the comment, can merge it first if it's urgent to unblock the release.

wmoustafa (approved):

PR #496, opened 2024-03-05
[Coral-Trino] Revert "remove char sets from TrinoSqlDialect (#460)"

ljfgem:

This PR reverts #460 due to its failure to remove CHARACTER SET from the Trino SQL in some production views. This caused the translated Trino SQL to become unparsable. Reverting it has resolved the issue. We intend to revert it in the Trino hotfix branch and revisit it later.

1. Tested on the production view, worked as expected
2. Regression test on all production views, no regression


ljfgem:
@aastha25 @KevinGe00 Regression test finished without regression, PTAL.

aastha25 (approved):
lgtm

PR #501, opened 2024-03-29
Add missing numeric precisions in Hive type system

KevinGe00:
Previously, unions between a `bigint` column and `int` column would result in an `int` column. Which is incorrect since `bigint` is a superset of `int`, i.e `bigint` is less restrictive than `int`, and not the other way around. This bug was caused by missing numeric precision valeus in `HiveTypeSystem` which is used in calcite to determine the least restrictive type between two columns. 

There a few changes in this changes, including some subsequent fixes that are needed after adding the precisions in:
1. Add precision values in `HiveTypeSystem` for all integer types: tinyint, smallint, int, bigint
- tinyint and smallint are needed as now that we have concrete precisions for int and bigint, we would mislead Calcite that int and bigint could have a lower precisions than tinyint and smallint if they also don't have concrete precisions to be compared to.
2. After adding in the precisions, a bug was revealed where we were incorrectly converting types to their respective specs in `FromUtcTimestampOperatorTransformer`. So replace that method with Calcite's already existing util function to properly convert types to specs for us.

1. clean build
4. new UT
5. expected regressions, example:

Snippet of Input Hive query:
```
CASE WHEN `table`.`a` = CAST(1 AS BIGINT)
```
 where column `a` is of type `int`

Previous Trino Translation:
```
CASE WHEN "table"."a" = CAST(CAST(1 AS BIGINT) AS INTEGER)
```
Since Calcite couldn't correctly figure out the least restrictive type because of missing precisions, it would cast the bigint type as int instead of the other way around.

New Trino Translation (regresssion):
```
CASE WHEN CAST("table"."a" AS BIGINT) = CAST(1 AS BIGINT)
```
Casting happens on the less restrictive type (int)


wmoustafa:
Should the precision be patched in Coral or Calcite? Should it go in LHS or IR? Sounds Calcite + IR is the correct answer based on recent discussions?

KevinGe00:
> Should the precision be patched in Coral or Calcite? 

There is no error in Calcite code/precision. What's happening is that Calcite uses whatever implementation of the interface `RelDataTypeSystem`, in the case of Coral using Calcite, we pass in `HiveTypeSystem` which implements `RelDataTypeSystem`. Therefore we need to patch in Coral.

> Should it go in LHS or IR?

`HiveTypeSystem` is used on the LHS in `HiveToRel`s `SqlToRelConverter` so patching `HiveTypeSystem` would automatically fix it in LHS.

> Sounds Calcite + IR is the correct answer based on recent discussions?

When you say IR do you mean that we need to align on a unified Coral type system whether that's using `HiveTypeSystem` or maybe Calcite's [default type system](https://jarvis.corp.linkedin.com/codesearch/result/?name=RelDataTypeSystemImpl.java&path=linkedin-calcite%2Fcore%2Fsrc%2Fmain%2Fjava%2Forg%2Fapache%2Fcalcite%2Frel%2Ftype&reponame=linkedin%2Flinkedin-calcite)? And once we align on a type system we use it in `CoralSqlNode1 -> CoralRelNode`?

I may be misunderstanding what you mean by IR here but an IR type system doesn't make sense to me. We need type systems to encapsulate differences in behaviors concerning type limits and behaviors. For example, in the default SQL system, a DECIMAL can have maximum precision 19, but Hive overrides to 38. We would need this information when converting a Hive input query.

wmoustafa:
Ok let me clarify. For Coral vs Calcite: I was under the impression that Hive INT and LONG will be converted to Calcite INT and LONG and then Calcite is responsible to internally figure out each's precision. Looking at the class contract, it seems that it does expect us to fill in the precession as part of the contract, and not simply return INT or LONG and that is it. So now it makes sense that we are obligated to fill in this value as part of the contract.

For the second part of the question, what I meant by an IR type system is the "Coral Type system" that all other systems get mapped to (e..g, from LHS). Upon further look, it sounds that this is indeed a Coral type system and not a Hive-specific type system, and the classname is just misleading due to historical reasons. Why is it a Coral type system: because it is in the `common` module. I now recall I have moved this from the hive module to the common module because it is independent of the LHS. So we are good there. Can you add an item to the backlog to rename this class to CoralTypeSystem?

coral-common/src/main/java/com/linkedin/coral/common/HiveTypeSystem.java
```diff
@@ -29,6 +29,10 @@ public class HiveTypeSystem extends RelDataTypeSystemImpl {
   private static final int DEFAULT_CHAR_PRECISION = 255;
   private static final int MAX_BINARY_PRECISION = Integer.MAX_VALUE;
   private static final int MAX_TIMESTAMP_PRECISION = 9;
+  private static final int DEFAULT_TINYINT_PRECISION = 3;
+  private static final int DEFAULT_SMALLINT_PRECISION = 5;
+  private static final int DEFAULT_INTEGER_PRECISION = 10;
+  private static final int DEFAULT_BIGINT_PRECISION = 19;
```
aastha25, line 35: could you link the reference for these values?

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/HiveToRelConverterTest.java
```diff
@@ -622,6 +622,15 @@ public void testNameSakeColumnNamesShouldGetUniqueIdentifiers() {
     assertEquals(generated, expected);
   }
 
+  @Test
+  public void testUnionIntAndBigInt() {
+    final String sql = "SELECT a FROM test.tableOne UNION ALL SELECT z FROM test.tableTwo";
```
aastha25, line 627: (1) can we add a couple more assertions in the unit test - 
(int union all bigint), (bigint union all int), (tinyint union all bigint) etc

(2) could you add another UT (here & in HiveToTrino tests) for `CASE WHEN `table`.`a` = CAST(1 AS BIGINT)` type queries as mentioned in the PR description

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/FromUtcTimestampOperatorTransformer.java
```diff
@@ -122,12 +120,6 @@ protected SqlCall transform(SqlCall sqlCall) {
   }
 
   private SqlCall castOperand(SqlNode operand, RelDataType relDataType) {
-    return SqlStdOperatorTable.CAST.createCall(ZERO, operand, getSqlDataTypeSpecForCasting(relDataType));
-  }
-
-  private SqlDataTypeSpec getSqlDataTypeSpecForCasting(RelDataType relDataType) {
```
aastha25, line 128: could you add an example of Coral translation for before & after this code change? Since there's no difference in unit tests, does that mean this is a bug or an unnecessary transformation?

aastha25 (commented):
Thanks for the PR @KevinGe00!
has this patch been tested against the li-trino-hotfix branch? could you add that in the PR description?

coral-common/src/main/java/com/linkedin/coral/common/HiveTypeSystem.java
KevinGe00: I updated the javadoc to specify where the values are coming from, this entire class was copied from Hive source code but has now deviated. We have a backlog item mentioned above to eventually rename this to `CoralTypeSystem`.

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/HiveToRelConverterTest.java
KevinGe00: (1) The reason I originally didn't add more tests is because Calcite already [has tests](https://github.com/apache/calcite/blob/d5b6b5c01ff17cac100a591a4cc4c9e83d4e1ace/core/src/test/java/org/apache/calcite/sql/type/SqlTypeFactoryTest.java#L54) using the same precision values. But definitely agree more tests in Coral don't hurt. I added tests comparing INT to the other 3 integer types since those are the most common comparisons.

(2) Added.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/FromUtcTimestampOperatorTransformer.java
KevinGe00: > could you add an example of Coral translation for before & after this code change

After adding in the precisions, the first case in `testSubstrWithTimestampOperator` in `HiveToTrinoConverterTest` started to fail because we were creating a faulty type spec that fails [this assertion ](https://jarvis.corp.linkedin.com/codesearch/result/?name=SqlBasicTypeNameSpec.java&path=linkedin-calcite%2Fcore%2Fsrc%2Fmain%2Fjava%2Forg%2Fapache%2Fcalcite%2Fsql&reponame=linkedin%2Flinkedin-calcite#219)in the Calcite's derive type function for `SqlBasicTypeNameSpec`. Therefore `getSqlDataTypeSpecForCasting` is a bug, if you look at  [`SqlTypeUtil.convertTypeToSpec`](https://jarvis.corp.linkedin.com/codesearch/result/?name=SqlTypeUtil.java&path=linkedin-calcite%2Fcore%2Fsrc%2Fmain%2Fjava%2Forg%2Fapache%2Fcalcite%2Fsql%2Ftype&reponame=linkedin%2Flinkedin-calcite#1023) there are more involved steps that involve using precision/scale values.

KevinGe00:
@aastha25 

> has this patch been tested against the li-trino-hotfix branch?

I did run tests againt the li-trino-hotfix branch but as expected there were regressions outside the scope of the PR since I branched this PR from master.

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
```diff
@@ -994,4 +994,19 @@ public void testSqlSelectAliasAppenderTransformerWithoutTableAliasPrefix() {
         + "WHERE \"tablea\".\"a\" > 5";
     assertEquals(expandedSql, expected);
   }
+
+  @Test
+  public void testIntCastToBigIntDuringComparison() {
+    // We're testing that a comparison between INT and BIGINT sees a cast on the more restrictive type to the
+    // less restrictive type and not the other way around. In other words, the INT is cast to BIGINT.
+    RelNode relNode = TestUtils.getHiveToRelConverter().convertSql(
+        "SELECT CASE WHEN a_integer = a_bigint THEN 'abc' ELSE 'def' END FROM test.table_from_utc_timestamp");
```
wmoustafa, line 1003: Are we using the best table for this test case `table_from_utc_timestamp`? Table name might be misleading since it indicates it has to do with timestamp?

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
```diff
@@ -994,4 +994,19 @@ public void testSqlSelectAliasAppenderTransformerWithoutTableAliasPrefix() {
         + "WHERE \"tablea\".\"a\" > 5";
     assertEquals(expandedSql, expected);
   }
+
+  @Test
+  public void testIntCastToBigIntDuringComparison() {
+    // We're testing that a comparison between INT and BIGINT sees a cast on the more restrictive type to the
```
wmoustafa, line 1000: Why is this required? What happens if you query Trino directly with a comparison operation without the CAST?

KevinGe00: It's not explicitly required (tested in trino to confirm), however, Calcite (during validation) will implicitly type coerce comparisons between 2 types by casting the more restrictive type to the more restrictive type. This PR just makes sure we give Calcite the right precision values to figure out correctly which type is more/less restrictive.

coral-trino/src/test/java/com/linkedin/coral/trino/rel2trino/HiveToTrinoConverterTest.java
KevinGe00: Added a new test table `tableInt` in rel2trino test utils.

wmoustafa (approved):

PR #502, opened 2024-04-01
[Coral-Trino] Remove unused HiveToTrinoConverter

ebyhr:


ebyhr:
cc: @wmoustafa

wmoustafa:
Since this is a public API, it could potentially be used by other external libraries. Prefer to keep. @aastha25 @KevinGe00 FYI.

PR #507, opened 2024-05-29
Correctly handle single type uniontypes in Coral

KevinGe00:
### Intro
The goal of the PR is to fix the unaccounted for side effects of single uniontype (a uniontype holding only one data type such as `uniontype<array<string>>`) related changes from #409.

### Example 1 revealing issues surrounding view text translations on field references for extractions on single union datatypes: 
**Necesary Context:** There is a Spark-specific mechanism that unwraps the uniontype to just the single underlying data type when reading from avro schemas. [Reference](https://spark.apache.org/docs/latest/sql-data-sources-avro.html#supported-types-for-avro---spark-sql-conversion). 

**Behavior before this PR:**

Hive
```
Table t1 that has the columns:
union_col           	uniontype<string>
struct_col           	struct<union_field:uniontype<string>>

View v1: select extract_union(union_col).tag_0 as a, extract_union(struct_col).union_field.tag_0 as b from t1
a：string
b: string

Both columns extract out the underlying field in the single uniontypes.
```
Spark
View v1 translated to: 
```
select coalesce_struct(union_col) a, coalesce_struct(struct_col).union_field b from t1
```

Trino
View v1 translated to: 
```
select extract_union(union_col) as a, extract_union(struct_col).union_field as b from t1
```

Issues:
1) In Trino, single uniontypes like `union_col` are not unwrapped like it is in spark, therefore `extract_union(union_col)` would produce `struct<tag_0:string>` so we need the `.tag_0` reference to extract out the underlying datatype.

2) We shouldn’t be taking 2 operators (extract_union UDF call + field reference) and then club them into a single identifier as it is not modular enough. We cannot assume  `extract_union` is never called on its own and need `extract_union` on single unions to work on it's own.

**Behavior after this PR:**

Hive portion remains the same as before

Spark
View v1 translated to: 
```
select coalesce_struct(union_col, 'uniontype<string>').tag_0, (coalesce_struct(struct_col, 'struct<union_field:uniontype<string>>').union_field).tag_0 b from t1
```

Trino
View v1 translated to: 
```
select extract_union(union_col).tag_0 as a, extract_union(struct_col).union_field.tag_0 as b from t1
```

Solution to issues:
1) Remove `SingleUnionFieldReferenceTransformer` entirely (which we no longer need anyways based on (2)), which was in Hive LHS. Now the Trino translation path for `extract_union` on single unions are left untouched.

2) Made changes in the `coalesce_struct` UDF to add a second optional schema string parameter, which is generated and passed in `ExtractUnionFunctionTransformer` while we still we have context of the Hive type and any single uniontypes. Now `coalesce_struct` UDF knows which fields (nested or not) came from a Hive single uniontype and can coalesce it as such.

### Example 2 revealing issues surrounding schema type derivations for Trino single uniontype: 

**Behavior before this PR:**

Hive
```
Table t1 that has the columns:
struct_col           	struct<union_field:uniontype<array<string>>>

View v1: select struct_col as struct_col from t1
struct_col           	struct<union_field:uniontype<array<string>>>
```

`struct_col` column Rel datatype in v1 RelNode
```
RecordType(
  RecordType:peek_no_expand(
    VARCHAR(2147483647) ARRAY union_field, 
    ) 
    struct_col)
```
Issue: Because of this change to [surface the underlying field & its schema in Coral's SqlNode RelNode representation when field is single uniontype](https://github.com/linkedin/coral/pull/409/files#diff-1513cf0b5cf70e78bb5b4008b15a7545a5513ec18b72997557e92eaf04add357R150), the Rel type of `struct_col` is simply `struct<union_field:array<string>>` instead of `struct<union_field:uniontype<array<string>>>`. This change was intended to fix an issue where the Avro schema for `extract_union` calls on single unions (generated from RelNode using coral-schema) could not be analyzed by the Spark engine. 

However, Coral generates Trino schemas also using RelNode, and Trino now cannot analyze the schema. We have a type coercion problem where Trino cannot coerce a col of type: `row(union_field row(tag tinyint, field0 array(varchar)))` ([a struct representing a single uniontype](https://github.com/trinodb/trino/pull/3483)) to the type of the column stored in the view definition (generated from Coral's RelNode with the uniontype unwrapped) `row(union_field array(varchar))`. 

**Behavior after this PR:**

Hive portion remains the same as before

`struct_col` column Rel datatype in v1 RelNode
```
RecordType(
  RecordType:peek_no_expand(
       RecordType:peek_no_expand(INTEGER tag, VARCHAR(2147483647) ARRAY field0) union_field) 
  struct_col)

```

Solution to issue: [Remove unwrapping of single uniontypes of RelNodes in Hive LHS](https://github.com/linkedin/coral/compare/master...KevinGe00:coral:single-unions?expand=1#diff-1513cf0b5cf70e78bb5b4008b15a7545a5513ec18b72997557e92eaf04add357L150) 

_Note this change doesn't break Spark's type derivation for single uniontypes at LinkedIn. There is no regression on the schema produced on a column that is a single uniontype in Hive, verified by creating a view with a single uniontype column and describing it in spark shell with this changes from this PR._

- Regression test results for trino, spark and avro are all as expected
- Unit tests
- Tested these changes with new `coalesec_struct` UDF changes to ensure entire UX is as expected


coral-spark/src/main/java/com/linkedin/coral/spark/transformers/ExtractUnionFunctionTransformer.java
```diff
@@ -56,6 +75,17 @@ protected SqlCall transform(SqlCall sqlCall) {
         createSqlOperator(COALESCE_STRUCT, CoalesceStructUtility.COALESCE_STRUCT_FUNCTION_RETURN_STRATEGY);
     if (operandList.size() == 1) {
       // one arg case: extract_union(field_name)
+      RelDataType operandType = deriveRelDatatype(sqlCall.operand(0));
+
+      if (containsSingleUnionType(operandType)) {
+        // Pass in schema string to keep track of the original Hive schema containing single uniontypes so coalesce_struct
```
wmoustafa, line 81: Expand the comment to explain why it is important for coalesce_struct to know this distinction.

coral-spark/src/main/java/com/linkedin/coral/spark/transformers/ExtractUnionFunctionTransformer.java
```diff
@@ -30,14 +39,24 @@
  * See {@link CoalesceStructUtility#COALESCE_STRUCT_FUNCTION_RETURN_STRATEGY} and its comments for more details.
  *
  * Check `CoralSparkTest#testUnionExtractUDF` for examples.
+ *
+ * Note that uniontypes holding a single need to be handled specially in Spark as there is a Spark-specific mechanism
+ * that unwraps a single uniontype (a uniontype holding only one data type) to just the single underlying data type.
+ * Reference: https://spark.apache.org/docs/latest/sql-data-sources-avro.html#supported-types-for-avro---spark-sql-conversion
+ *
+ * Check `CoralSparkTest#testUnionExtractUDFOnSingleTypeUnions` for examples.
+ *
```
wmoustafa, line 48: This should state that the behavior is specific to base tables. Giving an example query helps. In fact, you might write a few illustrating different expected behaviors of this transformer.

coral-spark/src/main/java/com/linkedin/coral/spark/DataTypeDerivedSqlCallConverter.java
```diff
@@ -0,0 +1,47 @@
...
+
+
+/**
+ * DataTypeDerivedSqlCallConverter transforms the sqlCalls
+ * in the input SqlNode representation to be compatible with Trino engine.
+ * The transformation may involve change in operator, reordering the operands
+ * or even re-constructing the SqlNode.
+ *
+ * All the transformations performed as part of this shuttle require RelDataType derivation.
+ */
```
wmoustafa, line 29: This description is confusing (also class name is confusing).

Why would a spark Shuttle do something specific to Trino?

The description sounds too generic and arbitrary without specific contract.

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
```diff
@@ -101,30 +102,44 @@ private static CoralSpark createWithAlias(RelNode irRelNode, List<String> aliase
...
+    SqlNode sparkSqlNode = coralSqlNodeWithRelDataTypeDerivedConversions
+        .accept(new CoralSqlNodeToSparkSqlNodeConverter()).accept(new CoralToSparkSqlCallConverter(sparkUDFInfos));
     return sparkSqlNode.accept(new SparkSqlRewriter());
   }
 
   public static String constructSparkSQL(SqlNode sparkSqlNode) {
     return sparkSqlNode.toSqlString(SparkSqlDialect.INSTANCE).getSql();
   }
 
+  private static boolean isSelectStar(SqlNode node) {
```
wmoustafa, line 134: What does this function have to do with single unions?

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
```diff
@@ -101,30 +102,44 @@ private static CoralSpark createWithAlias(RelNode irRelNode, List<String> aliase
...
-  private static SqlNode constructSparkSqlNode(RelNode sparkRelNode, Set<SparkUDFInfo> sparkUDFInfos) {
+  private static SqlNode constructSparkSqlNode(RelNode sparkRelNode, Set<SparkUDFInfo> sparkUDFInfos,
+      HiveMetastoreClient hmsClient) {
     CoralRelToSqlNodeConverter rel2sql = new CoralRelToSqlNodeConverter();
     SqlNode coralSqlNode = rel2sql.convert(sparkRelNode);
-    SqlNode sparkSqlNode = coralSqlNode.accept(new CoralSqlNodeToSparkSqlNodeConverter())
-        .accept(new CoralToSparkSqlCallConverter(sparkUDFInfos));
+
+    SqlNode coralSqlNodeWithRelDataTypeDerivedConversions =
+        coralSqlNode.accept(new DataTypeDerivedSqlCallConverter(hmsClient, coralSqlNode, sparkUDFInfos));
```
wmoustafa, line 123: Did not follow the intuition behind this (specifically in the context of single union types).

coral-spark/src/main/java/com/linkedin/coral/spark/transformers/ExtractUnionFunctionTransformer.java
KevinGe00: > This should state that the behavior is specific to base tables.

I'm not sure this unwrapping behavior specific to base tables.  I created a hive base table with a single union column, then created a view that selects all on the base table. Running `desc` on both table and view in Spark yield the same schema (single union is unwrapped).

> Giving an example query helps

On line 47, the comment points to [`CoralSparkTest.testUnionExtractUDFOnSingleTypeUnions`](https://github.com/linkedin/coral/pull/507/files#diff-83e46e8495b55d1ca847275a08bdbfe14395caf725b7a7253fcf69b457303bd4R425) which has examples covering simple and complex cases of Hive queries that would induce this transformer, and also the expected behavior.

coral-common/src/main/java/com/linkedin/coral/common/utils/RelDataTypeToHiveTypeStringConverter.java
```diff
@@ -41,6 +41,12 @@ public class RelDataTypeToHiveTypeStringConverter {
   private RelDataTypeToHiveTypeStringConverter() {
   }
 
+  public RelDataTypeToHiveTypeStringConverter(boolean convertUnionTypes) {
```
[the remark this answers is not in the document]
KevinGe00: Done.

coral-common/src/main/java/com/linkedin/coral/common/utils/RelDataTypeToHiveTypeStringConverter.java
```diff
@@ -110,6 +116,15 @@ public static String convertRelDataType(RelDataType relDataType) {
    */
   private static String buildStructDataTypeString(RelRecordType relRecordType) {
     List<String> structFieldStrings = new ArrayList<>();
+
+    // Convert single uniontypes back to Hive representation so coalesce_struct UDF can handle
+    // single uniontypes in Spark correctly
```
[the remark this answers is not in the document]
KevinGe00: Removed reference.

coral-spark/src/main/java/com/linkedin/coral/spark/DataTypeDerivedSqlCallConverter.java
KevinGe00: This class is a copy of [`DataTypeDerivedSqlCallConverter`](https://github.com/linkedin/coral/blob/master/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/DataTypeDerivedSqlCallConverter.java) in `rel2trino`. I could create a follow up PR to edit the name and javadoc for both `DataTypeDerivedSqlCallConverter` classes? 

> Why would a spark Shuttle do something specific to Trino?

Forgot to change "Trino engine." to "Spark engine." after copying, fixed.

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
KevinGe00: During coral spark RHS, RelNode -> SqlNode translation layer, there didn't used to be a need for transformers that require rel type derivation. However now, we need type derivation in `ExtractUnionFunctionTransformer` (which lives in `DataTypeDerivedSqlCallConverter`) to detect `extract_union` calls on **_single uniontypes_**. If we do detect a call on single uniontypes, only then pass in the schema string when transforming `extract_union` call to `coalesce_struct`.

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
KevinGe00: My apologies for not explaining earlier. After [adding in type derivation transformations on coral spark RHS](https://github.com/linkedin/coral/pull/507#discussion_r1683655254), there was a side-effect where the old [detection for select star queries no longer worked](https://github.com/linkedin/coral/pull/507/files#diff-dd2d5ae6e8e0d060ead980d488b9b9329bb10599982cfdef69a77432448482dbL108) requiring us to update the detection logic. This is a more robust check anyhow.

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
KevinGe00: Coincidentally this was a needed step for Coral IR upgrades, `Introduce API to enable type derivation in the SqlNode transformation layer` that we validated is doable. cc: @aastha25

coral-spark/src/main/java/com/linkedin/coral/spark/transformers/ExtractUnionFunctionTransformer.java
wmoustafa: > I'm not sure this unwrapping behavior specific to base tables. I created a hive base table with a single union column, then created a view that selects all on the base table. Running desc on both table and view in Spark yield the same schema (single union is unwrapped).

I meant explaining that the main reason for unwrapping relies in base tables. The current description is a bit vague and may not help with the thought process.

Stating examples in the comments helps vs pointing to the test cases.

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
wmoustafa: I think we need a discussion around this. Our objective here is to standardize `SqlNode` to `SqlNode` conversions to happen strictly through `SqlCallTransformers`. We should discuss if this API is sufficient and if not, how to organize/standardize things that happen outside it. Objective is to minimize ad hoc transformations, and this seems to add an ad hoc transformation.

coral-spark/src/main/java/com/linkedin/coral/spark/transformers/ExtractUnionFunctionTransformer.java
KevinGe00: @wmoustafa Updated this piece of javadoc to include end2end reasoning + examples in comments. PTAL

coral-spark/src/main/java/com/linkedin/coral/spark/CoralSpark.java
KevinGe00: Discussed offline that we organize CoralRelNode -> LanguageSqlNode as 3 logical steps:
1. Apply SqlShuttle for list of `SqlCall` transformers that require rel type derivation
2. Apply SqlShuttle for list of `SqlCall` transformers that do not require rel type derivation 
3. SqlNode transformations that cannot be done at the SqlCall layer (in coral-spark it is the `CoralSqlNodeToSparkSqlNodeConverter` class)

1 and 2 must be separated into 2 steps in that order as intermixing type derivation transformers with no type derivation transformers causes failures due to no type derivation transformers introducing certain operators that the type derivation util cannot yet handle.

A future PR will be set up to refactor these steps into a well documented class that loops through the 3 `SqlShuttles`.

wmoustafa (approved):

PR #508, opened 2024-06-05
[Coral-Hive] Extend VersionedSqlUserDefinedFunction to support version-specific function names

ljfgem:

This PR is to support versioning UDF resolution for one unversioned UDF static registration.
For example, user can register `com.linkedin.abc` once and then use variations with standard versioning prefix in Hive view definitions. A UDF class is considered versioned if the prefix before the first dot follows `coral_udf_version_(\d+|x)_(\d+|x)_(\d+|x)` format.
To achieve the goal, we made the following changes:
1. modify `HiveFunctionResolver` to allow versioning UDF resolution while static registry only contains unversioned class name.
2. add `functionClassName` and `getShortFunctionName()` to `VersionedSqlUserDefinedFunction`.

After successful resolution in coral-hive, coral-spark would be able to return the UDF with versioning class name (`functionClassName`) so that Spark is able to register the versioning UDF class well. And coral-trino will use registered short name or fall back to convert the camel case class name to lowercase underscore function name (`getShortFunctionName()`).

1. New unit tests in coral-hive, coral-spark and coral-trino, worked as expected.
4. Regression test on prod views.


wmoustafa:
> A UDF is considered shaded if the prefix before the first dot contains an underscore.

This convention is brittle.

Why is shading indicated by an underscore? Can we use something that utilizes patterns/templates or maybe suffixes? We will have to be explicit about how to resolve conflicts in case of multiple names/classes match.

By the way, resolving conflicts could happen in the "underscore" based approach, so it is not unique to the more generic solution suggested above. One resolution strategy is to fail loading the view if there is a collision.

We need a representation that does not open the door for too generic names like `.*` but also flexible enough to allow for minor things like version number.

ljfgem:
> > A UDF is considered shaded if the prefix before the first dot contains an underscore.
> 
> This convention is brittle.
> 
> Why is shading indicated by an underscore? Can we use something that utilizes patterns/templates or maybe suffixes? We will have to be explicit about how to resolve conflicts in case of multiple names/classes match.
> 
> By the way, resolving conflicts could happen in the "underscore" based approach, so it is not unique to the more generic solution suggested above. One resolution strategy is to fail loading the view if there is a collision.
> 
> We need a representation that does not open the door for too generic names like `.*` but also flexible enough to allow for minor things like version number.

@wmoustafa The shading prefix resolution only happens for static registered UDFs in `a_b.xyz` format (must contain dot and the prefix before dot must contain underscore), so there won't be issue if we ensure that the future registered UDFs follow this pattern (no underscore in the prefix before dot).
But agreed that we can create a more deterministic and well specified template for our use.

coral-spark/src/main/java/com/linkedin/coral/spark/transformers/FallBackToLinkedInHiveUDFTransformer.java
```diff
@@ -66,15 +67,15 @@ protected SqlCall transform(SqlCall sqlCall) {
       throw new UnsupportedUDFException(operatorName);
     }
     final String viewDependentFunctionName = operator.getViewDependentFunctionName();
+    String versionedFunctionName = FunctionUtils.getVersionedFunctionName(viewDependentFunctionName, operatorName);
```
wmoustafa, line 70: Since we leverage `VersionedSqlUserDefinedFunction` jointly with this new code, is there a way to encapsulate all this functionality in this class?

coral-common/src/main/java/com/linkedin/coral/common/utils/FunctionUtils.java
```diff
@@ -0,0 +1,61 @@
...
+      return className;
+    }
+  }
+
+  /**
+   * Generates a versioned function name based on the given function name and class name.
+   * For example, if the function name is "myFunction" and the class name is "coralversionedudf_1_0_0.com.linkedin.MyClass",
+   * the versioned function name will be "myFunction_1_0_0". If the class name is not shaded, such as "com.linkedin.MyClass",
+   * the versioned function name will be "myFunction".
+   */
```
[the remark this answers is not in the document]
wmoustafa: Also seeing the only use is in Spark (and the behavior is Spark specific), not sure if this fits in coral-common.

wmoustafa (commented):
First high-level review.

coral-common/src/main/java/com/linkedin/coral/common/utils/FunctionUtils.java
```diff
@@ -0,0 +1,61 @@
...
+
+import com.google.common.collect.ImmutableList;
+
+
+public class FunctionUtils {
+  public static final String CORAL_VERSIONED_UDF_PREFIX = "coralversionedudf_(\\d+|x)_(\\d+|x)_(\\d+|x)";
+
+  /**
+   * Checks if the given class name has a shading prefix.
+   * A class name is considered shaded if the prefix before the first dot
```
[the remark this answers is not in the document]
ljfgem: Sure, modified.

coral-common/src/main/java/com/linkedin/coral/common/utils/FunctionUtils.java
```diff
@@ -0,0 +1,61 @@
...
+package com.linkedin.coral.common.utils;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import com.google.common.collect.ImmutableList;
+
+
+public class FunctionUtils {
+  public static final String CORAL_VERSIONED_UDF_PREFIX = "coralversionedudf_(\\d+|x)_(\\d+|x)_(\\d+|x)";
```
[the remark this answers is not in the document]
ljfgem: Modified.

coral-common/src/main/java/com/linkedin/coral/common/utils/FunctionUtils.java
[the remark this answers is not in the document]
ljfgem: Moved it to `FallBackToLinkedInHiveUDFTransformer` in coral-spark.

coral-spark/src/main/java/com/linkedin/coral/spark/transformers/FallBackToLinkedInHiveUDFTransformer.java
ljfgem: Moved to a new method `getVersionedViewDependentFunctionName` in `VersionedSqlUserDefinedFunction`.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveFunctionResolver.java
```diff
@@ -182,24 +197,39 @@ public Collection<Function> tryResolveAsDaliFunction(String functionName, @Nonnu
...
-      if (dynamicResolvedFunctions.size() == 0) {
+      if (dynamicResolvedFunctions.isEmpty()) {
         // we want to see class name in the exception message for coverage testing
         // so throw exception here
         throw new UnknownSqlFunctionException(funcClassName);
       }
 
       return dynamicResolvedFunctions;
+    } else {
+      if (isVersioningUDF(funcClassName)) {
```
wmoustafa, line 215: The logic here should be:
* If it has the necessary properties (and hence Dali UDF)
  * If it is on the format `[name]+_version` or `db_t_+[name]`, then
    *  `short name = [name]`.
    * `function name = unversioned class name`
    * `long name = [name]+_version or db_t_+[name] or whatever original name is`
  * else a special case of the above
  
Ideally Dali views have `function-identifier` property that makes it possible not derive function name from the unversioned class name but the latter is acceptable in the interim, especially everything is encapsulated in this method.

wmoustafa: Also in case `short name = [name]` is not useful for Trino, we can have `short name` depend on `function name` through a hardcoded dependency. Intention of `short name` is to be used in engines that cannot leverage versioning like Trino.

ljfgem: Modified the code by adding a new method `getTrinoFunctionName` to `VersionedSqlUserDefinedFunction`, which returns the Trino function name as how we did before:
1. Try to retrieve the Trino function name from the map `TRINO_FUNC_NAME_MAP` based on the UDF class name
2. If the UDF class name doesn't exist in the map, follow the previous convention by converting the camel case class name to lowercase underscore function name.

Since we don't have a standard for version in view function name (we only have a standard for versioning prefix in class name), and view function name may not be related to the registered Trino function name, I would prefer to keep the current approach of generating Trino function name.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/VersionedSqlUserDefinedFunction.java
```diff
@@ -68,6 +87,27 @@ public String getViewDependentFunctionName() {
...
 
+  /**
+   * Retrieves the short function name based on the class name. If the class name is found
+   * in the predefined {@link VersionedSqlUserDefinedFunction#SHORT_FUNC_NAME_MAP},
+   * the corresponding short name is returned. Otherwise, the method converts the last
+   * segment of the class name from UPPER_CAMEL to LOWER_UNDERSCORE format to generate
+   * the short function name.
+   */
+  public String getShortFunctionName() {
+    String unversionedClassName = getName();
```
wmoustafa, line 98: Please add java doc about the nature of the function name (e.g., its format etc). You can say this is just a convention and other naming approaches are still valid as long as they identify the type inference.

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/HiveToRelConverterTest.java
```diff
@@ -356,6 +356,14 @@ public void testDaliUDFCall() {
     assertEquals(RelOptUtil.toString(rel), expectedPlan);
   }
 
+  @Test
+  public void testUDFWithVersioningClassName() {
```
wmoustafa, line 360: Please clarify the unit test name.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveFunctionResolver.java
```diff
@@ -44,6 +44,7 @@
  * Class to resolve hive function names in SQL to Function.
  */
 public class HiveFunctionResolver {
+  private static final String CORAL_VERSIONED_UDF_PREFIX = "coral_udf_version_(\\d+|x)_(\\d+|x)_(\\d+|x)";
```
wmoustafa, line 47: Remove `CORAL_` since `VERSIONED_UDF` is clear enough that it is related to VersionedUDF class?
Should use `CLASS_NAME_PREFIX`?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveFunctionResolver.java
```diff
@@ -218,7 +222,7 @@ public void addDynamicFunctionToTheRegistry(String funcClassName, Function funct
             new SqlUserDefinedFunction(new SqlIdentifier(funcClassName, ZERO),
                 new HiveGenericUDFReturnTypeInference(funcClassName, hiveTable.getDaliUdfDependencies()), null,
                 createSqlOperandTypeChecker(numOfOperands), null, null),
-            hiveTable.getDaliUdfDependencies(), functionName));
+            hiveTable.getDaliUdfDependencies(), functionName, funcClassName));
```
wmoustafa, line 225: At some point we deal with `functionName` as a class name (e.g., `String unversionedClassName = getName();`). Is there a way to stick to one terminology througout the whole code?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/VersionedSqlUserDefinedFunction.java
```diff
@@ -68,6 +87,27 @@ public String getViewDependentFunctionName() {
...
+      return SHORT_FUNC_NAME_MAP.get(unversionedClassName);
+    }
+    Converter<String, String> caseConverter = CaseFormat.UPPER_CAMEL.converterTo(CaseFormat.LOWER_UNDERSCORE);
+    String[] nameSplit = unversionedClassName.split("\\.");
+    return caseConverter.convert(nameSplit[nameSplit.length - 1]);
+  }
+
+  public String getUDFClassName() {
+    return udfClassName;
+  }
```
wmoustafa, line 109: FunctionName vs UDF Name.. can we fix all terminology and unify them to minimize confusion?

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/VersionedSqlUserDefinedFunction.java
ljfgem: Done

coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/HiveToRelConverterTest.java
ljfgem: Modified

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/VersionedSqlUserDefinedFunction.java
ljfgem: Modified

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveFunctionResolver.java
ljfgem: Modified to use `functionName` to refer to the function name in view text, and `funcClassName` to refer to the function class.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/HiveFunctionResolver.java
```diff
@@ -182,12 +183,14 @@ public Collection<Function> tryResolveAsDaliFunction(String functionName, @Nonnu
     if (funcClassName == null) {
       return ImmutableList.of();
     }
-    final Collection<Function> Functions = registry.lookup(funcClassName);
-    if (Functions.size() == 0) {
+    // If the UDF class name is versioned, remove the versioning prefix, which allows user to
+    // register the unversioned UDF once and use different versioning prefix in the view
+    final Collection<Function> functions = registry.lookup(removeVersioningPrefix(funcClassName));
```
wmoustafa, line 188: funcClassName ditto

wmoustafa (approved):
Thank you so much @ljfgem for working on this. Looks great now. Could you please update the title and description to match the new approach/sematics? I do not "shading" is a thing anymore.

It is probably along the lines of `Extend VersionedSqlUserDefinedFunction to support version-specific function names`.

Please make sure Trino tests pass both regression and server side.

ljfgem:
> Thank you so much @ljfgem for working on this. Looks great now. Could you please update the title and description to match the new approach/sematics? I do not "shading" is a thing anymore.
> 
> It is probably along the lines of `Extend VersionedSqlUserDefinedFunction to support version-specific function names`.
> 
> Please make sure Trino tests pass both regression and server side.

@wmoustafa Thank you for your reviews and suggestions on this PR! Modified the title and description, also finished the required tests.

PR #511, opened 2024-06-21
remove wrong CAST astnode in coral-hive

JiaLiangC:

There are two mechanisms in v3 for building abstract syntax trees (ASTs): operators and rewrite rules. 
-> Using the antlr3 rewrite symbol ^ tree construction operator

By inspecting all the AST nodes produced by the grammar files, we can see that KW_CAST is not used as a node in the Hive AST. The AbstractASTVisitor is designed to visit all non-DDL AST nodes, so KW_CAST needs to be removed.
I wrote a simple demo that prints the AST including the CAST syntax. It is transformed into a TOK_FUNCTION node, which validates the above points.
String src10 = "SELECT id,name,CAST(salary AS FLOAT) as salary_float FROM employees";
ASTNode root = parseDriver.parse(src10).getTree();
System.out.println(root.dump());
```
nil
   TOK_QUERY
      TOK_FROM
         TOK_TABREF
            TOK_TABNAME
               employees
      TOK_INSERT
         TOK_DESTINATION
            TOK_DIR
               TOK_TMP_FILE
         TOK_SELECT
            TOK_SELEXPR
               TOK_TABLE_OR_COL
                  id
            TOK_SELEXPR
               TOK_TABLE_OR_COL
                  name
            TOK_SELEXPR
               TOK_FUNCTION
                  TOK_FLOAT
                  TOK_TABLE_OR_COL
                     salary
               salary_float
   <EOF>
```

Test: The ParseTreeBuilderTest.java contains tests with SQL that includes the cast method. All Hive tests have passed. 
image

@wmoustafa Please help review it.

### DIFF
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/AbstractASTVisitor.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/parsetree/AbstractASTVisitor.java
@@ -180,7 +180,6 @@ public abstract class AbstractASTVisitor<R, C> {
       case HiveParser.TOK_ISNOTNULL:
       case HiveParser.TOK_ISNULL:
       case HiveParser.KW_CASE:
-      case HiveParser.KW_CAST:
       case HiveParser.KW_WHEN:
       case HiveParser.KW_IN:
       case HiveParser.KW_EXISTS:


wmoustafa:
> There are two mechanisms in v3 for building abstract syntax trees (ASTs): operators and rewrite rules.
> -> Using the antlr3 rewrite symbol ^ tree construction operator

Could you clarify what is the relationship between this intro and the PR? Especially the current coral-hive does not use Hive 3? Also, when stating v3 it is better to state it is v3 of which project, probably Hive.

wmoustafa:
While it is a good catch, I am not sure if it is sufficient to delete it based on the current tests passing. We probably need to dig deeper to find when this token is used since there might be other test cases that depend on it. Coral uses a suite of thousands of views to validate changes against internally and we might have to validate this through that framework. @KevinGe00 FYI.

JiaLiangC:
> > There are two mechanisms in v3 for building abstract syntax trees (ASTs): operators and rewrite rules.
> > -> Using the antlr3 rewrite symbol ^ tree construction operator
> 
> Could you clarify what is the relationship between this intro and the PR? Especially the current coral-hive does not use Hive 3? Also, when stating v3 it is better to state it is v3 of which project, probably Hive.

V3 is used by Hive1 with ANTLR v3. This fix is based on Hive1 and the description has been clarified.

PR #512, opened 2024-06-21
[Coral-Trino] Fix redundant transforms/casts from GenericProjectTransformer

KevinGe00:
#### Issue
Before #449, a (fuzzy union) view in hive with a column of type `users:array<struct<id:int,name:string>>`, would simply be selected in the final trino translation SQL as `users`.

After #449, the same column in the same view would be rewritten as in the trino translation:
```
transform(users, x -> cast(row("x"."id", "x"."name") as row("id" int, "name" varchar)))
```

The `transform` is redundant as all it's doing is transforming a column into it's own original type. Although it is harmless typing-wise, it could potentially degrade performance by wasting time/resources doing a no-op.

#### Why/How
#449 introduced a change where Coral IR (RelNodes) are generated using `StructKind.PEEK_FIELDS_NO_EXPAND` instead of `StructKind.FULLY_QUALIFIED`. This causes Coral's type derivation on RelNodes to also report back `StructKind.PEEK_FIELDS_NO_EXPAND`  instead of `StructKind.FULLY_QUALIFIED` (since derivation relies on `ToRelConverter` which [uses the new](https://github.com/linkedin/coral/blob/185b9acb7deafb1cc3c555adc2bc3b5cfb9561c4/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveToRelConverter.java#L55) `TypeFactory` from #449).

In `GenericProjectTransformer` when we now [compare](https://github.com/linkedin/coral/blob/185b9acb7deafb1cc3c555adc2bc3b5cfb9561c4/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/GenericProjectTransformer.java#L205) `fromDataType` vs `toDataType` to convert the Coral IR function `generic_project` to its Trino compatible representation.`fromDataType` is derived using Coral's type derivation and `toDataType` is derived [here](https://github.com/linkedin/coral/blob/185b9acb7deafb1cc3c555adc2bc3b5cfb9561c4/coral-common/src/main/java/com/linkedin/coral/common/functions/GenericProjectFunction.java#L46) in `GenericProjectFunction`. This causes a difference in `StructKind`s, comparisons here of 2 structs that are actually the same (same fields) then do not return equal. For the purposes of `GenericProjectTransformer`, we should consider the two StructKinds as equivalent since we only care about the struct fields (and whether or not they should be casted/transformed). Once the "inequivalent" structs are reported, the transformer will apply a redundant  `transform` or `cast`.

#### Fix
`fromDataType` is built in `FuzzyUnionSqlRewriter`, however, in this `RelDataType` building process we are missing a necessary declaration to override the default `StructKind` of the `Builder` object. Adding the declaration means `fromDataType` in `GenericProjectTransformer` will now use `StructKind.PEEK_FIELDS_NO_EXPAND`  instead of `StructKind.FULLY_QUALIFIED`, matching `toDataType`.

- Selectively tested on production views that were regressing for this reason
- i-tested, expected regressions (back to the state before 449)

### DIFF
--- a/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/GenericProjectTransformer.java
+++ b/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/GenericProjectTransformer.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2023-2023 LinkedIn Corporation. All rights reserved.
+ * Copyright 2023 LinkedIn Corporation. All rights reserved.
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
--- a/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/GenericProjectTransformer.java
+++ b/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/GenericProjectTransformer.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2023 LinkedIn Corporation. All rights reserved.
+ * Copyright 2023-2024 LinkedIn Corporation. All rights reserved.
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
@@ -361,7 +361,8 @@ public class GenericProjectTransformer extends SqlCallTransformer {
    * @param toDataType desired map RelDataType
    */
   private boolean areEqual(RelDataType fromDataType, RelDataType toDataType) {
-    String fromDataTypeFullyQualifiedStruct = fromDataType.getFullTypeString().replace("RecordType:peek_no_expand", "RecordType");
+    String fromDataTypeFullyQualifiedStruct =
+        fromDataType.getFullTypeString().replace("RecordType:peek_no_expand", "RecordType");
     return fromDataTypeFullyQualifiedStruct.equals(toDataType.getFullTypeString());
   }
 
--- a/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/GenericProjectTransformer.java
+++ b/coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/GenericProjectTransformer.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2023-2024 LinkedIn Corporation. All rights reserved.
+ * Copyright 2023-2023 LinkedIn Corporation. All rights reserved.
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
@@ -10,7 +10,6 @@ import java.util.List;
 
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeField;
-import org.apache.calcite.rel.type.RelDataTypeImpl;
 import org.apache.calcite.rel.type.RelRecordType;
 import org.apache.calcite.sql.SqlCall;
 import org.apache.calcite.sql.SqlIdentifier;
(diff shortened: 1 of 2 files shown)


coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/transformers/GenericProjectTransformer.java
```diff
@@ -348,6 +349,23 @@ private String relDataTypeFieldAccessString(RelDataType fromDataType, RelDataTyp
...
+   * StructKinds as equivalent for our purposes here as StructKind is used only as internal data representation.
+   *
+   * Note that this comparison using the full type string is derived from {@link RelDataTypeImpl#equals(Object)}.
+   *
+   * @param fromDataType given map RelDataType
+   * @param toDataType desired map RelDataType
+   */
+  private boolean areEqual(RelDataType fromDataType, RelDataType toDataType) {
+    String fromDataTypeFullyQualifiedStruct =
+        fromDataType.getFullTypeString().replace("RecordType:peek_no_expand", "RecordType");
```
wmoustafa, line 365: Can we make this check happen on the class members instead of the digest?

wmoustafa:
> In `GenericProjectTransformer` we compare ...

Can we always introduce which part of the pipeline this happens? Example: In LHS, when converting from Hive SqlNode to Coral SqlNode, etc. Heads up for others: @aastha25 @ljfgem @rzhang10

aastha25:
As discussed offline, please check if it is possible to use `StructKind.PEEK_FIELDS_NO_EXPAND` during all RelNode creation steps in the translation pipeline so that we dont have `StructKind.FULLY_QUALIFIED` type struct fields anywhere and hence no mismatch in the first place. String comparison for for different StructKind looks fragile. We could probably reuse `CoralJavaTypeFactoryImpl` and `HiveRelBuilder`.

wmoustafa:
> As discussed offline, please check if it is possible to use StructKind.PEEK_FIELDS_NO_EXPAND during all RelNode creation steps

What are the "all RelNode creation steps"? How to identify them?

aastha25:
> > As discussed offline, please check if it is possible to use StructKind.PEEK_FIELDS_NO_EXPAND during all RelNode creation steps
> 
> What are the "all RelNode creation steps"? How to identify them?

There's 2 places off the top of my head: 
a) on the LHS: Coral SqlNode1 ---> CoralRelNode. 
b) During [type derivation ](https://github.com/KevinGe00/coral/blob/master/coral-common/src/main/java/com/linkedin/coral/common/utils/TypeDerivationUtil.java#L74). 

I suspect this is the reason why `fromDataType` and `toDataType` don't match here https://github.com/linkedin/coral/pull/512/files#diff-5bd7b847ded0e80b31c380003182e5f91b9e8c2e202ac25d3ae2dff0a76c5b48R206

wmoustafa:
I think if we want to do this, we might want to always leverage a `RelBuilder` like [this](https://github.com/linkedin/coral/blob/bbf5f1587d7f6cb40dc396d6ff20a64136ea8306/coral-common/src/main/java/com/linkedin/coral/common/HiveRelBuilder.java#L59) (which actually triggers the `PEEK_FIELDS_NO_EXPAND` path). 

Also if we do it, I think we should do it only if it make sense semantically in the first place, with plausible explanation (as opposed to doing it solely to address the incompatibility check).

KevinGe00:
The issue has less to do with RelNode creation (which consistently uses the same `TypeFactory` across Coral). Issue lies in how we create the expected data type or `toDataType`, updated PR description to explain.

wmoustafa (approved):

PR #515, opened 2024-07-09
Use Sonatype token username and token password

ljfgem:

At present, we use Nexus UI sonatype username and password for release, which doesn't work now due to the recent maven central migration:
```
Caused by: org.gradle.internal.resource.transport.http.HttpErrorStatusCodeException: Could not PUT 'https://s01.oss.sonatype.org/service/local/staging/deployByRepositoryId/comlinkedincoral-1181/com/linkedin/coral/coral-common/2.2.34/coral-common-2.2.34.jar'. Received status code 401 from server: Content access is protected by token
```
To configure a publisher's plugin authentication we would need to update the plugin settings to use a user token instead of the Nexus UI username and password login. This PR modifies the configs with the newly added github secrets `SONATYPE_TOKEN_USERNAME` and `SONATYPE_TOKEN_PASSWORD`.

NA

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
         run: ./gradlew githubRelease publishToSonatype closeAndReleaseStagingRepository --stacktrace
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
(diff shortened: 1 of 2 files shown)


KevinGe00 (approved):
LGTM

PR #516, opened 2024-07-09
Add prev format in table name for Join

yyy1000:

### DIFF
--- a/coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeGenerationTransformer.java
+++ b/coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeGenerationTransformer.java
@@ -1,0 +1,164 @@
+package com.linkedin.coral.incremental;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import org.apache.calcite.plan.RelOptCluster;
+import org.apache.calcite.plan.RelOptTable;
+import org.apache.calcite.prepare.RelOptTableImpl;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.RelShuttle;
+import org.apache.calcite.rel.RelShuttleImpl;
+import org.apache.calcite.rel.core.TableScan;
+import org.apache.calcite.rel.logical.LogicalAggregate;
+import org.apache.calcite.rel.logical.LogicalFilter;
+import org.apache.calcite.rel.logical.LogicalJoin;
+import org.apache.calcite.rel.logical.LogicalProject;
+import org.apache.calcite.rel.logical.LogicalTableScan;
+import org.apache.calcite.rel.logical.LogicalUnion;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexNode;
+
+
+public class RelNodeGenerationTransformer {
+
+  static RelNode convertRelPrev(RelNode originalNode) {
+      RelShuttle converter = new RelShuttleImpl() {
+        @Override
+        public RelNode visit(TableScan scan) {
+          RelOptTable originalTable = scan.getTable();
+          List<String> incrementalNames = new ArrayList<>(originalTable.getQualifiedName());
+          String deltaTableName = incrementalNames.remove(incrementalNames.size() - 1) + "_prev";
+          incrementalNames.add(deltaTableName);
+          RelOptTable incrementalTable =
+              RelOptTableImpl.create(originalTable.getRelOptSchema(), originalTable.getRowType(), incrementalNames, null);
+          return LogicalTableScan.create(scan.getCluster(), incrementalTable);
+        }
+
+        @Override
+        public RelNode visit(LogicalJoin join) {
+          RelNode left = join.getLeft();
+          RelNode right = join.getRight();
+          RelNode prevLeft = convertRelPrev(left);
+          RelNode prevRight = convertRelPrev(right);
+          RexBuilder rexBuilder = join.getCluster().getRexBuilder();
+
+          LogicalProject p3 = createProjectOverJoin(join, prevLeft, prevRight, rexBuilder);
+
+          return p3;
+        }
+
+        @Override
+        public RelNode visit(LogicalFilter filter) {
+          RelNode transformedChild = convertRelPrev(filter.getInput());
+
+          return LogicalFilter.create(transformedChild, filter.getCondition());
+        }
+
+        @Override
+        public RelNode visit(LogicalProject project) {
+          RelNode transformedChild = convertRelPrev(project.getInput());
+          return LogicalProject.create(transformedChild, project.getProjects(), project.getRowType());
+        }
+
+        @Override
+        public RelNode visit(LogicalUnion union) {
+          List<RelNode> children = union.getInputs();
+          List<RelNode> transformedChildren =
+              children.stream().map(child -> convertRelPrev(child)).collect(Collectors.toList());
+          return LogicalUnion.create(transformedChildren, union.all);
+        }
+
+        @Override
+        public RelNode visit(LogicalAggregate aggregate) {
+          RelNode transformedChild = convertRelPrev(aggregate.getInput());
+          return LogicalAggregate.create(transformedChild, aggregate.getGroupSet(), aggregate.getGroupSets(),
+              aggregate.getAggCallList());
+        }
+      };
+      return originalNode.accept(converter);
+    }
+
+  private RelNodeGenerationTransformer() {
+  }
+
+  public static RelNode convertRelIncremental(RelNode originalNode) {
+    RelShuttle converter = new RelShuttleImpl() {
+      @Override
+      public RelNode visit(TableScan scan) {
+        RelOptTable originalTable = scan.getTable();
+        List<String> incrementalNames = new ArrayList<>(originalTable.getQualifiedName());
+        String deltaTableName = incrementalNames.remove(incrementalNames.size() - 1) + "_delta";
+        incrementalNames.add(deltaTableName);
+        RelOptTable incrementalTable =
+            RelOptTableImpl.create(originalTable.getRelOptSchema(), originalTable.getRowType(), incrementalNames, null);
+        return LogicalTableScan.create(scan.getCluster(), incrementalTable);
+      }
+
+      @Override
+      public RelNode visit(LogicalJoin join) {
+        RelNode left = join.getLeft();
+        RelNode right = join.getRight();
+        RelNode prevLeft = convertRelPrev(left);
+        RelNode prevRight = convertRelPrev(right);
+        RelNode incrementalLeft = convertRelIncremental(left);
+        RelNode incrementalRight = convertRelIncremental(right);
+
+        RexBuilder rexBuilder = join.getCluster().getRexBuilder();
+
+        LogicalProject p1 = createProjectOverJoin(join, prevLeft, incrementalRight, rexBuilder);
+        LogicalProject p2 = createProjectOverJoin(join, incrementalLeft, prevRight, rexBuilder);
+        LogicalProject p3 = createProjectOverJoin(join, incrementalLeft, incrementalRight, rexBuilder);
+
+        LogicalUnion unionAllJoins =
+            LogicalUnion.create(Arrays.asList(LogicalUnion.create(Arrays.asList(p1, p2), true), p3), true);
+        return unionAllJoins;
+      }
+
+      @Override
+      public RelNode visit(LogicalFilter filter) {
+        RelNode transformedChild = convertRelIncremental(filter.getInput());
+        return LogicalFilter.create(transformedChild, filter.getCondition());
+      }
+
+      @Override
+      public RelNode visit(LogicalProject project) {
+        RelNode transformedChild = convertRelIncremental(project.getInput());
+        return LogicalProject.create(transformedChild, project.getProjects(), project.getRowType());
+      }
+
+      @Override
+      public RelNode visit(LogicalUnion union) {
+        List<RelNode> children = union.getInputs();
+        List<RelNode> transformedChildren =
+            children.stream().map(child -> convertRelIncremental(child)).collect(Collectors.toList());
+        return LogicalUnion.create(transformedChildren, union.all);
+      }
+
+      @Override
+      public RelNode visit(LogicalAggregate aggregate) {
+        RelNode transformedChild = convertRelIncremental(aggregate.getInput());
+        return LogicalAggregate.create(transformedChild, aggregate.getGroupSet(), aggregate.getGroupSets(),
+            aggregate.getAggCallList());
+      }
+    };
+    return originalNode.accept(converter);
+  }
+
+  private static LogicalProject createProjectOverJoin(LogicalJoin join, RelNode left, RelNode right,
+      RexBuilder rexBuilder) {
+    LogicalJoin incrementalJoin =
+        LogicalJoin.create(left, right, join.getCondition(), join.getVariablesSet(), join.getJoinType());
+    ArrayList<RexNode> projects = new ArrayList<>();
+    ArrayList<String> names = new ArrayList<>();
+    IntStream.range(0, incrementalJoin.getRowType().getFieldList().size()).forEach(i -> {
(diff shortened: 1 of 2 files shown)


coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
```diff
@@ -0,0 +1,168 @@
...
+      return getExecutionCostTableScan((TableScan) rel);
+    } else if (rel instanceof LogicalJoin) {
+      return getExecutionCostJoin((LogicalJoin) rel);
+    } else if (rel instanceof LogicalUnion) {
+      return getExecutionCostUnion((LogicalUnion) rel);
+    } else if (rel instanceof LogicalProject) {
+      return getExecutionCostProject((LogicalProject) rel);
+    }
+    return new CostInfo(0.0, 0.0);
+  }
```
yyy1000, line 72: This is the core cost calculation framework.
It
1. Walk through the whole RelNode tree and get the cost
2. Keep the Row Count Information which will be used for IO cost

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
```diff
@@ -0,0 +1,168 @@
...
+  private List<JoinKey> findJoinKeys(LogicalJoin join) {
+    List<JoinKey> joinKeys = new ArrayList<>();
+    RexNode condition = join.getCondition();
+    if (condition instanceof RexCall) {
+      processRexCall((RexCall) condition, join, joinKeys);
+    }
+    return joinKeys;
+  }
+
+  private void processRexCall(RexCall call, LogicalJoin join, List<JoinKey> joinKeys) {
```
yyy1000, line 107: This function will from a Join condition to get all joinKeys, which will be used to get JoinSelectivity to estimate Join Size

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
```diff
@@ -0,0 +1,168 @@
...
+        RelDataTypeField rightField = rightType.getFieldList().get(rightIndex);
+        String rightTableName = getTableName(join.getRight().getTable());
+        String rightFieldName = rightField.getName();
+
+        joinKeys.add(new JoinKey(leftTableName, rightTableName, leftFieldName, rightFieldName));
+      }
+    }
+  }
+
+  private Double estimateJoinSelectivity(List<JoinKey> joinKeys) {
```
yyy1000, line 139: I didn't write the logic now, 
it will be from the join key, and from the statistic map info like distinct/row_count to get Selectivity

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
```diff
@@ -0,0 +1,237 @@
...
+
+    public JoinKey(String leftTableName, String rightTableName, String leftFieldName, String rightFieldName) {
+      this.leftTableName = leftTableName;
+      this.rightTableName = rightTableName;
+      this.leftFieldName = leftFieldName;
+      this.rightFieldName = rightFieldName;
+    }
+  }
+
+  private Map<String, Double> stat = new HashMap<>();
```
KevinGe00, line 63: Can you make the key to the stats map an `ENUM` called `CostStatistic` or just `Statistic`? You can follow `RewriteType` in coral-service to give each ENUM a value if needed.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
```diff
@@ -0,0 +1,237 @@
...
+    } else if (rel instanceof LogicalProject) {
+      return getExecutionCostProject((LogicalProject) rel);
+    }
+    return new CostInfo(0.0, 0.0);
+  }
+
+  private CostInfo getExecutionCostTableScan(TableScan scan) {
+    RelOptTable originalTable = scan.getTable();
+    String tableName = getTableName(originalTable);
+    Double row = stat.getOrDefault(tableName, 5.0);
```
KevinGe00, line 142: Would it be smarter to skip over the table if we have no statistics about it? Using a default value could throw off the final cost by a significant magnitude.

If we're keeping, please make `5.0` a default constant variable.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
```diff
@@ -0,0 +1,237 @@
...
+        for (Map.Entry<String, JsonElement> distinctEntry : distinctCounts.entrySet()) {
+          String columnName = distinctEntry.getKey();
+          Double distinctCount = distinctEntry.getValue().getAsDouble();
+
+          distinctStat.put(tableName + ":" + columnName, distinctCount);
+        }
+
+      }
+    } catch (IOException e) {
+      e.printStackTrace();
```
KevinGe00, line 115: Would be more graceful to handle this catch block by throwing an exception, can create a custom exception if need be.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+  }
+
+  public Double getCost(RelNode rel) {
+    CostInfo executionCostInfo = getExecutionCost(rel);
+    Double IOCost = executionCostInfo.row * IOCostParam;
+    return executionCostInfo.cost * shuffleCostParam + IOCost;
```
KevinGe00, line 123: Please add an explanation how you came to this equation

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
```diff
@@ -0,0 +1,237 @@
...
+public class RelNodeCostEstimator {
+
+  class CostInfo {
+    // TODO: we may also need to add TableName field.
+    Double cost;
+    Double row;
+
+    public CostInfo(Double cost, Double row) {
+      this.cost = cost;
+      this.row = row;
```
KevinGe00, line 45: `numRows` or `rowCount` would be more accurate.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
```diff
@@ -0,0 +1,237 @@
...
+  private CostInfo getExecutionCostJoin(LogicalJoin join) {
+    RelNode left = join.getLeft();
+    RelNode right = join.getRight();
+    if (!(left instanceof TableScan) || !(right instanceof TableScan)) {
+      return new CostInfo(0.0, 0.0);
+    }
+    CostInfo leftCost = getExecutionCost(left);
+    CostInfo rightCost = getExecutionCost(right);
+    Double joinSize = estimateJoinSize(join, leftCost.row, rightCost.row);
+    return new CostInfo(max(leftCost.cost, rightCost.cost), joinSize);
```
KevinGe00, line 159: Can you explain here and in comments why we are taking the max?

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  private List<JoinKey> findJoinKeys(LogicalJoin join) {
+    List<JoinKey> joinKeys = new ArrayList<>();
+    RexNode condition = join.getCondition();
+    if (condition instanceof RexCall) {
+      processRexCall((RexCall) condition, join, joinKeys);
+    }
+    return joinKeys;
```
KevinGe00, line 168: Could you add an assertion here that `joinKeys.size` is greater than or equal to 1, otherwise throw an exception?

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
KevinGe00: Could you name this something more descriptive like `getJoinKeysFromJoinCondition`?

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  private String getTableName(RelOptTable table) {
+    return String.join(".", table.getQualifiedName());
+  }
+
+  private CostInfo getExecutionCostJoin(LogicalJoin join) {
+    RelNode left = join.getLeft();
+    RelNode right = join.getRight();
+    if (!(left instanceof TableScan) || !(right instanceof TableScan)) {
```
KevinGe00, line 153: Are expanding the cost estimator to handle joins for more than 2 tables?

KevinGe00 (changes requested):
Thanks for the PR! Left some comments mainly about readability and clarity :)

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
```diff
@@ -0,0 +1,237 @@
...
+    this.stat = stat;
+  }
+
+  public void setDistinctStat(Map<String, Double> distinctStat) {
+    this.distinctStat = distinctStat;
+  }
+
+  private Double IOCostParam = 1.0;
+
+  private Double shuffleCostParam = 1.0;
```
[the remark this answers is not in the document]
yyy1000: That's fine!

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
yyy1000: Because in modern distributed systems like Spark, the shuffle cost is dominated by the largest shuffle, multicore execute the left and right at the same time.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
yyy1000: Discussed it offline and make the statistic data a structure like:
```
Class Statistic {
	Double rowcount;
	Map<string(column name), double> distinctCountByRow;
}
```

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
```diff
@@ -0,0 +1,281 @@
...
+
+  private CostInfo getExecutionCostTableScan(TableScan scan) {
+    RelOptTable originalTable = scan.getTable();
+    String tableName = getTableName(originalTable);
+    try {
+      TableStatistic tableStat = costStatistic.get(tableName);
+      Double rowCount = tableStat.rowCount;
+      return new CostInfo(rowCount, rowCount);
+    } catch (NullPointerException e) {
+      throw new IllegalArgumentException("Table statistics not found for table: " + tableName);
```
KevinGe00, line 176: Could you add a unit test that triggers this catch block?

coral-incremental/src/test/java/com/linkedin/coral/incremental/RelNodeCostEstimatorTest.java
```diff
@@ -0,0 +1,75 @@
...
+    RelNode relNode = hiveToRelConverter.convertSql(sql);
+    estimator.loadStatistic(TEST_JSON_FILE_DIR + "statistic.json");
+    assertEquals(estimator.getCost(relNode), 300.0);
+    estimator.setStat(fakeStatData());
+    assertEquals(estimator.getCost(relNode), 240.0);
+  }
+
+  @Test
+  public void testSimpleJoin() throws IOException {
+    String sql = "SELECT * FROM test.bar1 JOIN test.bar2 ON test.bar1.x = test.bar2.x";
```
KevinGe00, line 68: The idea is not to get rid of incremental computation from the cost estimator testing.

We want to test that given a correct list of  RelNodes representing the batch, partially incremental and fully incrementally of a 2-table join, does this estimator compute the right cost for each and eventual return the cheapest plan?

So you can hard-code the correct batch, partially incremental and fully incrementally queries and feed them into `hiveToRelConverter.convertSql`, then pass the resulting RelNodes to your cost estimator.

coral-incremental/src/main/java/com/linkedin/coral/incremental/TableStatistic.java
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright 2024 LinkedIn Corporation. All rights reserved.
+ * Licensed under the BSD-2 Clause license.
+ * See LICENSE in the project root for license information.
+ */
+package com.linkedin.coral.incremental;
+
+import java.util.Map;
+
+
+public class TableStatistic {
```
KevinGe00, line 11: Any reason this is not a nested class in `RelNodeCostEstimator`? Are we expecting to use this class outside of `RelNodeCostEstimator`?

KevinGe00 (commented):
Thanks for addressing the previous feedback :)

coral-incremental/src/test/java/com/linkedin/coral/incremental/RelNodeCostEstimatorTest.java
yyy1000: Yeah, I see. 
I used your method before but failed. It's because `hiveToRelConverter.convertSql` couldn't handle the sql for 'table_prev' and 'table_delta', (didn't register the table_prev like schema before), thus can't convert it to RelNode. 
I think for this PR, not using incremental computation is OK because the incremental format is just a union with three joins ? Could discuss this more. :)

coral-incremental/src/main/java/com/linkedin/coral/incremental/TableStatistic.java
yyy1000: good catch, will move it inside the class.

coral-incremental/src/test/java/com/linkedin/coral/incremental/RelNodeCostEstimatorTest.java
yyy1000: Will test table with 'prev' and 'delta' format once plan generation part is integrated.
Now only test raw format with multi-format query.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
```diff
@@ -0,0 +1,285 @@
...
+      this.shuffleCost = shuffleCost;
+      this.rowCount = row;
+    }
+  }
+
+  class TableStatistic {
+    // The number of rows in the table
+    Double rowCount;
+    // The number of distinct values in each column
+    Map<String, Double> distinctCountByRow;
```
wmoustafa, line 70: Does this work for nested columns? If yes, what is the spec of the key?

Does it work for complex types?

If the answer is no, we should be explicit about this at least in the Java doc.

coral-incremental/src/main/java/com/linkedin/coral/incremental/RelNodeCostEstimator.java
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  private CostInfo getExecutionCostJoin(LogicalJoin join) {
+    RelNode left = join.getLeft();
+    RelNode right = join.getRight();
+    CostInfo leftCost = getExecutionCost(left);
+    CostInfo rightCost = getExecutionCost(right);
+    Double joinSize = estimateJoinSize(join, leftCost.rowCount, rightCost.rowCount);
+    // The shuffle cost of a join is the maximum shuffle cost of its children because
+    // in modern distributed systems, the shuffle cost is dominated by the largest shuffle.
```
wmoustafa, line 195: Terminology is not clear here. What is the shuffle cost of children? We might need to discuss this over a doc.

wmoustafa:
Overall feedback is that I believe the PR sounds too specific and there is no general framework for cost estimation that is applicable to various operator by specializing some of aspects. I think we need to iterate over this PR to make it more abstract with various implementations. This may be better discussed over a doc.

PR #523, opened 2024-08-05
Replace FunctionFieldReferenceOperator with Calcite DOT operator

KevinGe00:
We can get rid of `FunctionFieldReferenceOperator` entirely and replace it with Calcite's default `SqlDotOperator`. Historically `FunctionFieldReferenceOperator` was introduced because "Calcite DOT operator by default does not work for referencing fields from struct return types of functions." However, that's no longer the case, thus rendering `FunctionFieldReferenceOperator` unneeded. If we can get rid of `FunctionFieldReferenceOperator`, we also no longer need to migrate it (as part of Coral IR upgrades).

- clean build
- ran regression tests for avro, trino and spark
Only regression I'm seeing is a cosmetic one that is actually desirable. Unnecessary parentheses introduced because FunctionFieldReferenceOperator is not a native Calcite operator go away:
```
old: (coalesce_struct(lithograph_etl.contentmetadata).tag_1).legacyhelpcenteranswerid
new: coalesce_struct(lithograph_etl.contentmetadata).tag_1.legacyhelpcenteranswerid
```


aastha25 (commented):
1) Have we verified that the cosmetic changes have no impact when queried on Spark/Trino? This should be a simple experiment to execute the two variants on Spark/Trino for a sample table. 

2) Could you also confirm the the RelNode representations have no differences apart from the operator name?

PR #533, opened 2024-09-17
Add EXTRACT_COLLECTION UDF to registry

felixqfu:
Added a new UDF to static Hive function registry

./gradlew clean build

### DIFF
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
@@ -682,7 +682,7 @@ public class StaticHiveFunctionRegistry implements FunctionRegistry {
     createAddUserDefinedFunction("com.linkedin.groot.runtime.udf.spark.GetMappedValueUDF", FunctionReturnTypes.STRING,
         family(SqlTypeFamily.STRING, SqlTypeFamily.STRING));
     createAddUserDefinedFunction("com.linkedin.groot.runtime.udf.spark.ExtractCollectionUDF",
-        FunctionReturnTypes.arrayOfType(SqlTypeName.ANY, true), family(SqlTypeFamily.ANY, SqlTypeFamily.STRING));
+        FunctionReturnTypes.arrayOfType(SqlTypeName.VARCHAR, true), family(SqlTypeFamily.ANY, SqlTypeFamily.STRING));
 
     // UDTFs
     addFunctionEntry("explode", new CoralSqlUnnestOperator(false));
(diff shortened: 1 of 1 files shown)


coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
```diff
@@ -681,6 +681,8 @@ public boolean isOptional(int i) {
         SqlTypeFamily.BOOLEAN, SqlTypeFamily.ANY, SqlTypeFamily.ARRAY, SqlTypeFamily.STRING, SqlTypeFamily.STRING));
     createAddUserDefinedFunction("com.linkedin.groot.runtime.udf.spark.GetMappedValueUDF", FunctionReturnTypes.STRING,
         family(SqlTypeFamily.STRING, SqlTypeFamily.STRING));
+    createAddUserDefinedFunction("com.linkedin.groot.runtime.udf.spark.ExtractCollectionUDF",
+        FunctionReturnTypes.arrayOfType(SqlTypeName.ANY, true), family(SqlTypeFamily.ANY, SqlTypeFamily.STRING));
```
ljfgem, line 685: return type shouldn't contain `SqlTypeName.ANY`, is this the same type as the first parameter's type?

rzhang10: @ljfgem  I discussed with Qiang offline, this UDF's return type is pretty complicated, it depends on its input arg type via some complex logic, and I'm not sure if it's even possible to express all that type in calcite. So I think using a broader arrayOfType of ANY type to express it should be fine here? I understand it will be a wider type than what this function strictly is, but it shouldn't cause any problems as long as the user of this udf knows what it's doing and uses it correctly with expected operand types.

ljfgem: IIRC, we encountered a similar issue while generating Avro schema for the UDF because `RelDataTypeToAvroType#basicSqlTypeToAvroType` doesn't support `ANY` type and would [throw exception](https://github.com/linkedin/coral/blob/e48ff5e6bf6c476b605a49075a1e7a55f4f3b0e1/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelDataTypeToAvroType.java#L132).

If coral-schema fails, the UDF couldn't be registered property and Spark query would fail.

felixqfu: The inputs are 1. an array field in dataframe which array contains struct and potentially nested arrays. 2. a string path reference from (1) to a primitive leaf field (int, long, string etc). 
The return type will be an array of the type of leaf field, which type is only available at runtime

felixqfu: @ljfgem If ANY is not possible, is there any way to set it to a union type? Something like `Array<Int, Long and String>`

felixqfu: Changed to `FunctionReturnTypes.arrayOfType(SqlTypeName.VARCHAR, true)`
Please review again. Thanks

dihu-linkedin (approved):

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/functions/StaticHiveFunctionRegistry.java
dihu-linkedin: Entry looks fine to me

ljfgem (approved):

PR #544, opened 2024-12-18
[Coral-Trino] use 'TRY_CAST' instead of 'CAST'

hqbhoho:
 Hive/Spark CAST is equivalent to Trino TRY_CAST. If CAST fails, a query using Hive/Spark will return null, while a query using Trino will throw error. So I think should  use 'TRY_CAST' instead of 'CAST'.
- **Hive/Spark**
select cast('a' as int)  return null 
- **Trino**
select cast('a' as int)   throw error like "Cannot cast 'a' to INT"
select try_cast('a' as int)   return null 

Modified unit tests.


hqbhoho:
@wmoustafa Could you help review it?

PR #548, opened 2025-01-30
[Coral-schema] Generalize operand schema inference on ordinal return type UDF calls

KevinGe00:
Currently, coral-schema is only set up to correctly infer schemas for UDFs with an ordinal return type, only if the ordinal operand is a reference to an input column. However, this is limiting and we have since encountered views with nested calls on UDFs with an ordinal return type. For example, `UDF_A('foo', UDF_A('foo', colA))`, where the return type of `UDF_A` is whatever the return type of it's second operand is.

We correct this by using the `SchemaRexShuttle` to visit the ordinal operand, which will infer the schema of the operand and add it to the final schema as a field. This way, we leverage the shuttle to categorize and process the operand for us so we don't have to manually categorize the operand.

 
- New unit test to schema inference on nested UDF calls with ordinal return types
- `./gradlew clean build`
- regression test passes
- tested schema produced on spark-shell for a view with previously incorrect schema (due to incorrect nullabilites)

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2019-2024 LinkedIn Corporation. All rights reserved.
+ * Copyright 2019-2025 LinkedIn Corporation. All rights reserved.
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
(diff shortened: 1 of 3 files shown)


simbadzina:
Did the regression tests introduce any changes to view text?

KevinGe00:
> Did the regression tests introduce any changes to view text?

No changes, I only ran the regression for avro schema translations since that is the only possible translation path that this change could introduce changes in.

simbadzina (approved):
LGTM. Secondary changes to the schema noted in other comments.

aastha25 (approved):

PR #552, opened 2025-05-20
Register ExtractCollectionUdf as Trino UDF

KevinGe00:
This PR maps the ExtractCollectionUDF UDF to its Trino counterpart, so users can query the Hive views using these UDFs after registering the corresponding Trino function names in Trino.

` ./gradlew clean build`

No user is currently using this UDF for now so it should not cause any regressions.


ljfgem (approved):
lgtm

PR #57, opened 2021-03-18
Use plain String concatenation for test setup

findepi:
This allows easy copying from IDE into e.g. `beeline`.

PR #556, opened 2025-10-24
Add native Apache Iceberg table support with CoralCatalog abstraction

aastha25:

**Summary**
This PR introduces native Apache Iceberg table support to Coral, enabling direct schema conversion from Iceberg to Calcite's RelDataType without lossy intermediate conversions through Hive's type system. The implementation preserves Iceberg-specific type semantics including timestamp precision and explicit nullability. 
Key architectural decision: `HiveMetastoreClient` remains unchanged and 
does NOT extend `CoralCatalog`. Integration classes use composition 
(storing both instances) with runtime dispatch.

**New Components**
- Catalog Abstraction (`coral-common/src/main/java/com/linkedin/coral/common/catalog/`)
- `CoralCatalog`: Format-agnostic catalog interface with `getTable()`, `getAllTables()`, `namespaceExists()`
- `CoralTable`: Unified table metadata interface (`name()`, `properties()`, `tableType()`)
- `HiveCoralTable` / `IcebergCoralTable`: Implementations wrapping native Hive/Iceberg table objects
- TableType`: Simple enum (TABLE or VIEW)

**Iceberg Integration**
- `IcebergTable`: Calcite ScannableTable implementation for Iceberg tables
- `IcebergTypeConverter`: Converts Iceberg Schema → Calcite RelDataType with precision preservation
- `IcebergHiveTableConverter`: Backward compatibility bridge for UDF resolution (converts Iceberg → Hive table object)

**Integration Pattern**
- `HiveSchema`, `HiveDbSchema`, `ToRelConverter`: Store both `CoralCatalog`  and `HiveMetastoreClient` instances
- Runtime dispatch: if `coralCatalog != null` use unified path; 
  else if `msc != null` use Hive-only path
- `HiveMetastoreClient` and `HiveMscAdapter` marked @Deprecated (still functional, prefer CoralCatalog)

**How Reviewers Should Read This**
Start here:
1. `CoralCatalog.java` - New abstraction layer interface
2. `CoralTable.java` - Unified table metadata interface
3. `IcebergCoralTable.java` - How Iceberg tables are wrapped
4. `IcebergTypeConverter.java` - Core schema conversion logic

Then review integration:
1. `HiveDbSchema.java` - Dispatch logic based on `CoralTable` type (Iceberg vs Hive)
2. `IcebergTable.java` - Calcite integration
3. `ToRelConverter.java` - Dual-path support (CoralCatalog vs HiveMetastoreClient)
4. `HiveMetastoreClient.java` - Backward compatibility

Test:
1. `IcebergTableConverterTest.java` - End-to-end Iceberg conversion test

New and existing tests pass
integration tests - WIP


coral-common/src/main/java/com/linkedin/coral/common/IcebergTypeConverter.java
```diff
@@ -0,0 +1,218 @@
...
+        break;
+      case DATE:
+        convertedType = typeFactory.createSqlType(SqlTypeName.DATE);
+        break;
+      case TIME:
+        convertedType = typeFactory.createSqlType(SqlTypeName.TIME);
+        break;
+      case TIMESTAMP:
+        // Iceberg timestamp type - microsecond precision (6 digits)
+        convertedType = typeFactory.createSqlType(SqlTypeName.TIMESTAMP, 6);
```
sumedhsakdeo: Can we handle timestamp with time zone for completeness?
```suggestion
        Types.TimestampType timestampType = (Types.TimestampType) icebergType;
        if (timestampType.shouldAdjustToUTC()) {
                // TIMESTAMP WITH TIME ZONE - stores instant in time
               convertedType = typeFactory.createSqlType(SqlTypeName.TIMESTAMP_WITH_LOCAL_TIME_ZONE, 6);
        } else {
               // TIMESTAMP - stores local datetime
               convertedType = typeFactory.createSqlType(SqlTypeName.TIMESTAMP, 6);
        }
```
Ref: https://github.com/apache/calcite/blob/95350ed1a449bbb2f008fcf2b704544e7d95c410/core/src/main/java/org/apache/calcite/sql/type/SqlTypeName.java#L73

coral-common/src/main/java/com/linkedin/coral/common/IcebergTypeConverter.java
```diff
@@ -0,0 +1,218 @@
...
+        convertedType = typeFactory.createSqlType(SqlTypeName.FLOAT);
+        break;
+      case DOUBLE:
+        convertedType = typeFactory.createSqlType(SqlTypeName.DOUBLE);
+        break;
+      case DATE:
+        convertedType = typeFactory.createSqlType(SqlTypeName.DATE);
+        break;
+      case TIME:
+        convertedType = typeFactory.createSqlType(SqlTypeName.TIME);
```
sumedhsakdeo:
```suggestion
        convertedType = typeFactory.createSqlType(SqlTypeName.TIME, 6);
```

coral-common/src/main/java/com/linkedin/coral/common/IcebergTypeConverter.java
```diff
@@ -0,0 +1,218 @@
...
+      case UUID:
+        // Represent UUID as CHAR(36) to preserve UUID semantics
+        convertedType = typeFactory.createSqlType(SqlTypeName.CHAR, 36);
+        break;
+      case FIXED:
+        Types.FixedType fixedType = (Types.FixedType) icebergType;
+        convertedType = typeFactory.createSqlType(SqlTypeName.BINARY, fixedType.length());
+        break;
+      case BINARY:
+        convertedType = typeFactory.createSqlType(SqlTypeName.VARBINARY, Integer.MAX_VALUE);
```
sumedhsakdeo: Any particular reason why we use VARBINARY over BINARY here? 
Unlike HiveTypeConverter https://github.com/linkedin/coral/blob/9f8dfce736e3b14aded7653cb340c75bf51dab7b/coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java#L86

coral-common/src/main/java/com/linkedin/coral/common/IcebergTypeConverter.java
```diff
@@ -0,0 +1,218 @@
...
+   * @param mapType Iceberg map type
+   * @param typeFactory Calcite type factory
+   * @return RelDataType representing the map
+   */
+  public static RelDataType convert(Types.MapType mapType, RelDataTypeFactory typeFactory) {
+    // Recursively convert key and value types
+    RelDataType keyType = convert(mapType.keyType(), typeFactory);
+    RelDataType valueType = convert(mapType.valueType(), typeFactory);
+
+    // Iceberg map values can be required or optional
```
sumedhsakdeo:
```suggestion
    keyType = typeFactory.createTypeWithNullability(keyType, false);
    // Iceberg map values can be required or optional
```
Nit

coral-common/src/main/java/com/linkedin/coral/common/IcebergTable.java
```diff
@@ -0,0 +1,96 @@
...
+  @Override
+  public Statistic getStatistic() {
+    // Future enhancement: Could use Iceberg statistics here
+    // Iceberg provides rich statistics: row count, file count, size, etc.
+    return Statistics.UNKNOWN;
+  }
+
+  @Override
+  public Schema.TableType getJdbcTableType() {
+    return dataset.tableType() == TableType.VIEW ? Schema.TableType.VIEW : Schema.TableType.TABLE;
```
sumedhsakdeo:
```suggestion
     return Schema.TableType.TABLE;
```
🤔, with an assert that  dataset.tableType() should be TableType.MANAGED_TABLE?

coral-common/src/main/java/com/linkedin/coral/common/IcebergTable.java
```diff
@@ -0,0 +1,96 @@
...
+
+  @Override
+  public Enumerable<Object[]> scan(DataContext root) {
+    throw new RuntimeException("Calcite runtime execution is not supported");
+  }
+
+  /**
+   * Returns the underlying Iceberg Table for advanced operations.
+   * 
+   * @return Iceberg Table object
```
sumedhsakdeo:
```suggestion
   * @return org.apache.iceberg.Table
```

coral-common/src/main/java/com/linkedin/coral/common/catalog/IcebergHiveTableConverter.java
```diff
@@ -0,0 +1,114 @@
...
+import org.apache.hadoop.hive.metastore.api.TableType;
+import org.apache.iceberg.hive.HiveSchemaUtil;
+
+
+/**
+ * Utility class to convert Iceberg datasets to Hive Table objects for backward compatibility.
+ *
+ * <p>This converter creates complete Hive Table objects from Iceberg tables, including schema conversion
+ * using {@code HiveSchemaUtil}. While the table object acts as "glue code" for backward compatibility,
+ * it populates all standard Hive table metadata to ensure broad compatibility with downstream code paths.
```
sumedhsakdeo: Do we expect to exercise this glue code in practice? If so, under what scenarios?

coral-common/src/main/java/com/linkedin/coral/common/catalog/IcebergHiveTableConverter.java
```diff
@@ -0,0 +1,114 @@
...
+    Map<String, String> serdeParams = new HashMap<>();
+    if (hiveParameters.containsKey("avro.schema.literal")) {
+      serdeParams.put("avro.schema.literal", hiveParameters.get("avro.schema.literal"));
+    }
+    serDeInfo.setParameters(serdeParams);
+    storageDescriptor.setSerdeInfo(serDeInfo);
+
+    // Convert Iceberg schema to Hive columns
+    try {
+      storageDescriptor.setCols(HiveSchemaUtil.convert(icebergTable.schema()));
```
sumedhsakdeo, line 100: Any particular reason why we choose to set storageDescriptor columns from HiveSchemaUtil.convert(icebergTable.schema()) and not from AvroSchemaUtil.convert(hiveParameters.get("avro.schema.literal"))?

coral-common/src/main/java/com/linkedin/coral/common/catalog/IcebergHiveTableConverter.java
```diff
@@ -0,0 +1,114 @@
...
+      // If schema conversion fails, set empty columns list
+      // This shouldn't break function resolution as it only needs properties
+      storageDescriptor.setCols(new ArrayList<>());
+    }
+
+    // Create Hive Table object with all metadata
+    Table hiveTable = new Table(tableName, dbName, StringUtils.EMPTY, // owner
+        0, // createTime
+        0, // lastModifiedTime
+        0, // retention
```
sumedhsakdeo, line 111: What are the side-effects of empty metadata here for owner .. retention?

coral-common/src/main/java/com/linkedin/coral/common/catalog/CoralCatalog.java
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * Checks if a namespace (database) exists in the catalog.
+   * This provides a lightweight way to verify namespace existence
+   * without retrieving full metadata.
+   *
+   * @param dbName Database or namespace name
+   * @return true if the namespace exists, false otherwise
+   */
+  boolean namespaceExists(String dbName);
```
wmoustafa: Let us fix the inconsistencies between `namespace`, `db`, `schema`.

wmoustafa:
Let us not use `Dataset` as it is not a standard term. `Table` maybe more appropriate, but I understand you want to use it elsewhere, but we can find an alternative. I think `Schema` is also closer to calcite terminolorgy than `Namespace`. As much as possible we should use standard terms, or when in ambiguity, we should be closer to Calcite terminology.

coral-common/src/main/java/com/linkedin/coral/common/catalog/CoralCatalog.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.List;
+
+
+/**
+ * Top-level catalog interface for accessing datasets in Coral.
+ * This interface provides a unified API for querying table metadata
+ * across different table formats (Hive, Iceberg, etc.).
+ *
+ * CoralCatalog abstracts away the differences between various table formats
+ * and provides a consistent way to access dataset information through
```
wmoustafa: Ditto on "Dataset" terminology.

coral-common/src/main/java/com/linkedin/coral/common/catalog/DatasetConverter.java
```diff
@@ -0,0 +1,68 @@
...
+   * This method always treats the table as a Hive table.
+   *
+   * This is a convenience method equivalent to {@link #fromHiveTable(Table)}.
+   *
+   * @param table Hive metastore Table object
+   * @return Dataset representing the table (HiveDataset)
+   */
+  public static Dataset autoConvert(Table table) {
+    return fromHiveTable(table);
+  }
```
wmoustafa: What is the objective of this one line method? Does not seem to be needed.

coral-common/src/main/java/com/linkedin/coral/common/catalog/HiveDataset.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  /**
+   * Returns the table type (TABLE or VIEW).
+   *
+   * @return TableType enum value
+   */
+  @Override
+  public TableType tableType() {
+    return TableType.fromHiveTableType(table.getTableType());
```
wmoustafa: Can you write the spec of conversion between Hive, Iceberg, and Coral representations? How does this expand for more table formats? Ideally we should have a Coral representation that is universal enough and everything can be converted to it. So I would expect methods like `toCoralType` as opposed to `fromHiveType`. Underlying table formats should not be hard coded in the universal catalog as well.

wmoustafa (changes requested):

coral-common/src/main/java/com/linkedin/coral/common/IcebergTypeConverter.java
aastha25: you're right, im changing this to be same as hive.

coral-common/src/main/java/com/linkedin/coral/common/IcebergTypeConverter.java
aastha25: key shouldnt be nullable and by default it will be created as non-nullable.

coral-common/src/main/java/com/linkedin/coral/common/catalog/IcebergHiveTableConverter.java
aastha25: This code path is used to read table properties when parsing the view SQL in @parsetreeBuilder. 

Yes, this glue code gets exercised primarily to retrieve eligible table properties on the base tables during parsing  
stage in ParseTreeBuilder/HiveFunctionResolver (no schema dependency). Without this glue, we would need larger scale refactoring in those classes to interpret `IcebergTable` natively.

coral-common/src/main/java/com/linkedin/coral/common/catalog/IcebergHiveTableConverter.java
aastha25: (a) iceberg schema is the SOT (b) avro literal may not always exist or could be stale & (c) this logic is ported from existing production code paths so as to have consistency in how we convert iceberg table -> hive table object across the stack. 
Practically, setting this one way or the other in this specific class has no bearing on view schema resolution.

coral-common/src/main/java/com/linkedin/coral/common/catalog/IcebergHiveTableConverter.java
aastha25: none, we practically only need tbl properties for backward compatibility with SQL parser logic in `ParseTreeBuilder`

coral-common/src/main/java/com/linkedin/coral/common/catalog/HiveDataset.java
aastha25: `TableType.fromHiveTableType` this method has been deleted. 
Also, as discussed, the spec of table formats -> coral IR is just schema conversion which is captured in classes `TypeConverter` for hive tables & `IcebergTypeConverter` for iceberg tables.

coral-common/src/main/java/com/linkedin/coral/common/catalog/DatasetConverter.java
aastha25: this class has been removed. resolving comment.

coral-common/src/main/java/com/linkedin/coral/common/catalog/CoralCatalog.java
aastha25: thanks for the feedback, I have refactored the PR to move away from `Dataset`. 
we now have
(1) `coralCatalog` (new catalog interface) & `HiveMetastoreClient` (old catalog interface) are independent and both work with Coral for translations. `HiveMetastoreClient` has been marked as deprecated in favor of coralCatalog.
(2) `getTable()` is the API in `coralCatalog`. It returns an interface of `CoralTable`. Currently, we have 2 impls of coralTable - `hiveCoralTable` & `icebergCoralTable`

coral-common/src/main/java/com/linkedin/coral/common/catalog/IcebergHiveTableConverter.java
```diff
@@ -0,0 +1,113 @@
...
+    // Utility class - prevent instantiation
+  }
+
+  /**
+   * Converts IcebergCoralTable to a Hive Table object for backward compatibility with function resolution.
+   *
+   * @param icebergCoralTable Iceberg coral table to convert
+   * @return Hive Table object with complete metadata and schema
+   */
+  public static Table toHiveTable(IcebergCoralTable icebergCoralTable) {
```
wmoustafa: Based on offline discussion, I understood we got rid of those, but seems we still leverage methods that hardcode the table formats. This is not good for extensibility.

wmoustafa:
Can we eliminate 
```java
if (coralCatalog != null) {
  ...
} else {
  ...
}
```
that we are currently using everywhere and use some adapter class instead?

coral-common/src/main/java/com/linkedin/coral/common/catalog/CoralCatalog.java
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * Checks if a namespace (database) exists in the catalog.
+   * This provides a lightweight way to verify namespace existence
+   * without retrieving full metadata.
+   *
+   * @param namespaceName Namespace (database) name
+   * @return true if the namespace exists, false otherwise
+   */
+  boolean namespaceExists(String namespaceName);
```
wmoustafa: Inconsistency between namespace, and schema elsewhere.

coral-hive/src/test/java/com/linkedin/coral/common/IcebergTableConverterTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+    // Verify column types
+    assertEquals(fields.get(0).getType().getSqlTypeName(), SqlTypeName.BIGINT);
+    assertEquals(fields.get(2).getType().getSqlTypeName(), SqlTypeName.VARCHAR);
+
+    // CRITICAL: Verify that TIMESTAMP has precision 6 (microseconds)
+    RelDataType timestampType = fields.get(1).getType();
+    assertEquals(timestampType.getSqlTypeName(), SqlTypeName.TIMESTAMP, "event_time should be TIMESTAMP");
+    assertEquals(timestampType.getPrecision(), 6, "Timestamp should have precision 6 (microseconds)");
+  }
```
wmoustafa, line 118: I found [here](https://github.com/apache/calcite/blob/63633ac9fba4bfc9538f850a08a283c1f7279fec/core/src/main/java/org/apache/calcite/rel/type/RelDataTypeSystemImpl.java#L204) that the max precision of `TIMESTAMP` in Calcite is `3`. I also wrote a similar test in my local and it failed for me. Curious how it worked here. Could you clarify which particular `RelDataTypeFactory` is being picked up in this test?

wmoustafa: Ok. I checked out the PR and found it was `HiveTypeSystem` which has max precision of `9`.

coral-common/src/main/java/com/linkedin/coral/common/catalog/CoralCatalog.java
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * Retrieves a table by namespace and table name.
+   * This method returns a unified CoralTable abstraction that works
+   * across different table formats (Hive, Iceberg, etc.).
+   *
+   * @param namespaceName Namespace (database) name
+   * @param tableName Table name
+   * @return CoralTable object representing the table, or null if not found
+   */
+  CoralTable getTable(String namespaceName, String tableName);
```
wmoustafa: `namespaceName` -> `namespace`.

coral-common/src/main/java/com/linkedin/coral/common/catalog/CoralTable.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.linkedin.coral.common.types.CoralDataType;
+
+
+/**
+ * A unified abstraction representing a table or view in Coral.
+ * This interface provides a common way to access table metadata regardless
+ * of the underlying table format (Hive, Iceberg, etc.).
+ *
+ * This abstraction is used by Calcite integration layer to dispatch to
```
wmoustafa: Let us not mention implementation details (e.g., Calcite) explicitly in Coral common module. Can you check the rest of the PR?

coral-common/src/main/java/com/linkedin/coral/common/catalog/CoralTable.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+public interface CoralTable {
+
+  /**
+   * Returns the fully qualified table name in the format "database.table".
+   *
+   * @return Fully qualified table name
+   */
+  String name();
+
```
wmoustafa, line 26: Does the method name need to reflect it is a fully qualified table name? Also, is this method used anywhere? Let us add things only when necessary.

coral-common/src/main/java/com/linkedin/coral/common/catalog/HiveCoralTable.java
```diff
@@ -0,0 +1,130 @@
...
+
+
+/**
+ * Implementation of {@link CoralTable} interface for Hive tables.
+ * This class wraps a Hive metastore Table object and provides
+ * a unified CoralTable API for accessing table metadata.
+ *
+ * Used by Calcite integration to dispatch to HiveTable.
+ */
+public class HiveCoralTable implements CoralTable {
```
wmoustafa: There are table implementations in `common` and table implementations in `catalog`. What is the basis for defining in each?

wmoustafa: Does this need to be implemented as a plugin to avoid class path issues? Also note that both Hive and Iceberg co-exist in the same module which is not quite normal.

coral-common/src/main/java/com/linkedin/coral/common/catalog/HiveTable.java
```diff
@@ -0,0 +1,130 @@
...
+
+    for (FieldSchema col : allCols) {
+      final String colName = col.getName();
+
+      // Skip duplicate columns (partition keys might overlap with regular columns)
+      if (!fieldNames.contains(colName)) {
+        // Convert Hive type string to TypeInfo, then to CoralDataType
+        final TypeInfo typeInfo = TypeInfoUtils.getTypeInfoFromTypeString(col.getType());
+        final CoralDataType coralType = HiveToCoralTypeConverter.convert(typeInfo);
+
```
wmoustafa, line 118: Example of package inconsistency. Utility method for this implementation is in the `common` package.

wmoustafa (dismissed):

coral-common/src/main/java/com/linkedin/coral/common/catalog/IcebergTable.java
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * Returns the underlying Iceberg Table object.
+   * Used by Calcite integration layer (IcebergTable).
+   *
+   * @return Iceberg Table object
+   */
+  public org.apache.iceberg.Table getIcebergTable() {
+    return table;
+  }
```
wmoustafa, line 82: Ideally you should not need this, and it might indicate a leak in the API. Can we avoid it?

coral-common/build.gradle
```diff
@@ -16,4 +16,12 @@ dependencies {
   }
 
   compile deps.'hadoop'.'hadoop-common'
+
+  // LinkedIn Iceberg dependencies
```
wmoustafa, line 20: Another reason why this needs to be a plugin. This should integrate with OSS Iceberg too.

coral-common/src/main/java/com/linkedin/coral/common/ToRelConverter.java
```diff
@@ -138,20 +181,76 @@ public SqlNode toSqlNode(String sql) {
...
+        // It's a Hive table
+        stringViewExpandedText = "SELECT * FROM " + dbName + "." + tableName;
+      }
+    } else if (coralTable instanceof IcebergCoralTable) {
+      // Iceberg coral table: always a table (Iceberg doesn't have views)
+      IcebergCoralTable icebergCoralTable = (IcebergCoralTable) coralTable;
+
+      // Convert Iceberg coral table to minimal Hive Table for backward compatibility
+      // This is needed because downstream code (ParseTreeBuilder, HiveFunctionResolver)
+      // expects a Hive Table object for Dali UDF resolution
```
wmoustafa, line 224: Can you elaborate more? Why not use `CoralTable` there?

coral-common/src/main/java/com/linkedin/coral/common/catalog/IcebergHiveTableConverter.java
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * <p><b>What gets converted:</b>
+ * <ul>
+ *   <li>Iceberg schema → Hive columns (via {@code HiveSchemaUtil.convert()})</li>
+ *   <li>All Iceberg table properties → Hive table parameters (including Dali UDF metadata)</li>
+ *   <li>Table metadata (name, owner, timestamps, table type)</li>
+ *   <li>Storage descriptor with SerDe info (for compatibility)</li>
+ * </ul>
+ */
+public class IcebergHiveTableConverter {
```
wmoustafa, line 50: The point of this change is not to do this anymore.

coral-trino/src/main/java/com/linkedin/coral/trino/rel2trino/DataTypeDerivedSqlCallConverter.java
```diff
@@ -48,6 +49,17 @@ public DataTypeDerivedSqlCallConverter(HiveMetastoreClient mscClient, SqlNode to
...
 
+  public DataTypeDerivedSqlCallConverter(CoralCatalog coralCatalog, SqlNode topSqlNode) {
+    toRelConverter = new HiveToRelConverter(coralCatalog);
+    topSqlNode.accept(new RegisterDynamicFunctionsForTypeDerivation());
+
+    TypeDerivationUtil typeDerivationUtil = new TypeDerivationUtil(toRelConverter.getSqlValidator(), topSqlNode);
+    operatorTransformerList = SqlCallTransformers.of(new FromUtcTimestampOperatorTransformer(typeDerivationUtil),
+        new GenericProjectTransformer(typeDerivationUtil), new NamedStructToCastTransformer(typeDerivationUtil),
+        new ConcatOperatorTransformer(typeDerivationUtil), new SubstrOperatorTransformer(typeDerivationUtil),
+        new CastOperatorTransformer(typeDerivationUtil), new UnionSqlCallTransformer(typeDerivationUtil));
```
wmoustafa, line 57: Refactor the code to avoid repitition?

coral-common/src/main/java/com/linkedin/coral/common/IcebergCalciteTableAdapter.java
```diff
@@ -0,0 +1,108 @@
...
+   *
+   * Uses two-stage conversion: Iceberg → Coral → Calcite.
+   * This provides a unified type system abstraction across table formats.
+   *
+   * @param typeFactory Calcite type factory
+   * @return RelDataType representing the table schema
+   */
+  @Override
+  public RelDataType getRowType(RelDataTypeFactory typeFactory) {
+    // Stage 1: Iceberg → Coral
```
wmoustafa: Nit: `Step`.

coral-common/src/main/java/com/linkedin/coral/common/IcebergTypeConverter.java
```diff
@@ -0,0 +1,218 @@
...
+/**
+ * Converts Iceberg Schema and Types to Calcite RelDataType.
+ * Preserves Iceberg type semantics including nullability and nested structures.
+ *
+ * This converter provides native Iceberg schema support for Calcite, avoiding
+ * lossy conversions through Hive type system.
+ *
+ * Copied structure from TypeConverter for consistency.
+ */
+public class IcebergTypeConverter {
```
wmoustafa: I do not see why this is required. Could you explain?

coral-common/src/main/java/com/linkedin/coral/common/HiveDbSchema.java
```diff
@@ -17,46 +17,86 @@
...
-    checkNotNull(msc);
-    checkNotNull(dbName);
+  /**
+   * Constructor for HiveDbSchema. Exactly one of coralCatalog or msc must be non-null.
+   * 
+   * @param coralCatalog Coral catalog for unified access (can be null if msc is provided)
+   * @param msc Hive metastore client for Hive-specific access (can be null if coralCatalog is provided)
+   * @param dbName Database name (must not be null)
+   */
+  HiveDbSchema(CoralCatalog coralCatalog, HiveMetastoreClient msc, @Nonnull String dbName) {
```
wmoustafa: Can we introduce the `coral` classes here, depending only on `CoralCatalog`, and mark the `Hive` ones as deprecated?

coral-common/src/main/java/com/linkedin/coral/common/catalog/IcebergHiveTableConverter.java
```diff
@@ -0,0 +1,119 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.hadoop.hive.metastore.api.SerDeInfo;
+import org.apache.hadoop.hive.metastore.api.StorageDescriptor;
+import org.apache.hadoop.hive.metastore.api.Table;
+import org.apache.iceberg.hive.HiveSchemaUtil;
+
+
+/**
+ * Utility class to convert Iceberg datasets to Hive Table objects for backward compatibility.
```
sumedhsakdeo, line 20: Does it also convert schema, if yes, is the converter is lossy?

coral-common/src/main/java/com/linkedin/coral/common/types/BinaryType.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.util.Objects;
+
+
+/**
+ * Represents a BINARY type with optional length in the Coral type system.
+ *
+ * Length indicates:
+ *  - -1: unbounded/variable-length (LENGTH_UNBOUNDED, default for BINARY/VARBINARY)
+ *  -  0: fixed-length (for FIXED types, e.g., Iceberg FIXED(16))
```
sumedhsakdeo:
```suggestion
 *  -  >0: fixed-length (for FIXED types, e.g., Iceberg FIXED(16))
```
?

coral-common/src/main/java/com/linkedin/coral/common/IcebergToCoralTypeConverter.java
```diff
@@ -0,0 +1,175 @@
...
+        return PrimitiveType.of(CoralTypeKind.FLOAT, nullable);
+      case DOUBLE:
+        return PrimitiveType.of(CoralTypeKind.DOUBLE, nullable);
+      case DATE:
+        return PrimitiveType.of(CoralTypeKind.DATE, nullable);
+      case TIME:
+        return PrimitiveType.of(CoralTypeKind.TIME, nullable);
+      case TIMESTAMP:
+        // Iceberg timestamp has microsecond precision (6 digits)
+        return TimestampType.of(6, nullable);
```
sumedhsakdeo, line 112: Do we need to distinguish between TIMESTAMP w/ timezone and w/o timezone?
https://github.com/linkedin/iceberg/blob/ebf4776724f346310105a58a6966a69dba2200c1/api/src/main/java/org/apache/iceberg/types/Types.java#L49

coral-common/src/main/java/com/linkedin/coral/common/IcebergToCoralTypeConverter.java
```diff
@@ -0,0 +1,175 @@
...
+        // Represent UUID as CHAR(36) to preserve UUID semantics
+        return CharType.of(36, nullable);
+      case FIXED:
+        Types.FixedType fixedType = (Types.FixedType) icebergType;
+        // Fixed-length binary - preserve length
+        return BinaryType.of(fixedType.length(), nullable);
+      case BINARY:
+        // Variable-length binary
+        return BinaryType.of(BinaryType.LENGTH_UNBOUNDED, nullable);
+      case DECIMAL:
```
sumedhsakdeo, line 126: Don't see DECIMAL in IcebergType https://github.com/linkedin/iceberg/blob/809534da01bc5a3002c5edd427e91eff09ec2443/api/src/main/java/org/apache/iceberg/types/Types.java#L54

coral-common/src/main/java/com/linkedin/coral/common/ToRelConverter.java
```diff
@@ -138,20 +180,78 @@ public SqlNode toSqlNode(String sql) {
...
   @VisibleForTesting
   public SqlNode processView(String dbName, String tableName) {
-    org.apache.hadoop.hive.metastore.api.Table table = hiveMetastoreClient.getTable(dbName, tableName);
-    if (table == null) {
-      throw new RuntimeException(String.format("Unknown table %s.%s", dbName, tableName));
-    }
-    String stringViewExpandedText = null;
-    if (table.getTableType().equals("VIRTUAL_VIEW")) {
-      stringViewExpandedText = table.getViewExpandedText();
+    if (coralCatalog != null) {
```
sumedhsakdeo, line 185: javadoc is missing updates for processView

coral-common/src/main/java/com/linkedin/coral/common/ToRelConverter.java
sumedhsakdeo: +1, also is the conversion lossy

coral-common/src/main/java/com/linkedin/coral/common/catalog/IcebergHiveTableConverter.java
aastha25: captured the reasoning in detail in the class doc and linked this ticket https://github.com/linkedin/coral/issues/575, specifying that this class will be deleted in a follow up patch.

coral-common/src/main/java/com/linkedin/coral/common/catalog/IcebergHiveTableConverter.java
aastha25: captured the reasoning in detail in the class doc and linked this ticket https://github.com/linkedin/coral/issues/575, specifying that this class will be deleted in a follow up patch.

coral-common/src/main/java/com/linkedin/coral/common/catalog/CoralCatalog.java
aastha25: resolved it is multiple places. lmk if i missed it someplace.

coral-common/src/main/java/com/linkedin/coral/common/catalog/CoralTable.java
aastha25: API called `name()` returning the fully qualified table name (db.tbl) seems a standard convention (also adopted in iceberg). It is used in code, so there is relevance to this method. 
Also, I have clarified in the javadoc what does this method expected to return.

coral-common/src/main/java/com/linkedin/coral/common/catalog/CoralTable.java
aastha25: resolved this comment.

coral-common/src/main/java/com/linkedin/coral/common/catalog/HiveCoralTable.java
aastha25: discussed offline and also tried to clarify in the class code + java docs.
TL;DR: `catalog` has user <-> coral interaction. `common` has coral <-> calcite interaction

aastha25: hive & iceberg dependencies co-existing in dynamic data lakes is an inevitable reality. On the feedback of injecting this dependency at runtime, that can be implemented in multiple way. We can do it in Coral or the downstream consumer of Coral can exclude iceberg brought in by Coral (upstream) & inject their own iceberg. 
I'm going to evaluate the feasibility of implementing as this a plugin in Coral outside the scope of this PR.

coral-common/src/main/java/com/linkedin/coral/common/catalog/IcebergTable.java
aastha25: I have marked this as an internal API which will be removed in future release. This should be clean up as part of https://github.com/linkedin/coral/issues/57.

coral-common/build.gradle
aastha25: specifically I want to bring in the custom shaded distribution of li-iceberg-hive-metastore here. But in general, I feel ok but Coral depending on li-iceberg. 
```
'iceberg-hive-metastore': "com.linkedin.iceberg:iceberg-hive-metastore:${versions['linkedin-iceberg']}:shaded"
```

coral-common/src/main/java/com/linkedin/coral/common/ToRelConverter.java
aastha25: the integration here requires only "view SQL definition" & view's "tbl properties". so these details are lossless in converting icebergTable (coralTable) ->  hiveTable. 
why we need this? that explain in detail in https://github.com/linkedin/coral/issues/575

coral-common/src/main/java/com/linkedin/coral/common/catalog/IcebergHiveTableConverter.java
aastha25: captured the reasoning in detail in the class doc and linked this ticket #575, specifying that this class will be deleted in a follow up patch.
the converted schema is not used.

coral-common/src/main/java/com/linkedin/coral/common/HiveDbSchema.java
aastha25: good idea, please take a look at the new code. I have added new parallel classes here which work with `CoralCatalog`

coral-common/src/main/java/com/linkedin/coral/common/IcebergToCoralTypeConverter.java
aastha25: You're correct that DECIMAL isn't in Iceberg's TYPES map, but that's by design. The TYPES map only contains singleton primitive types (types without parameters) like IntegerType.get(), StringType.get(), etc. DECIMAL is indeed a primitive type in Iceberg's type system (it's a leaf type, not a complex/nested type like List, Map, or Struct), but it's a parameterized primitive that requires precision and scale parameters. Since DECIMAL(10,2) is a different type from DECIMAL(20,5), you can't use a singleton pattern—each instance must be constructed with specific parameters using DecimalType.of(precision, scale). 
This is why our converter handles it in the switch statement but doesn't expect to find it in the singleton TYPES map. Similar to how FIXED (another parameterized primitive) also isn't in the map.

coral-common/src/main/java/com/linkedin/coral/common/IcebergToCoralTypeConverter.java
aastha25: Yes Iceberg has TIMESTAMP w/ timezone and w/o timezone. However, both variants return typeId() == TypeID.TIMESTAMP 1, so there's conversion coverage and they both go through the same case TIMESTAMP: branch in this converter.

**Current behavior:** Both are mapped to TimestampType.of(6, nullable) since Coral's TimestampType currently doesn't express timezone awareness.

**Future enhancement:** We will investigate how to express timezone in Coral IR (coral type system and also underlying calcite representation) to preserve this distinction. This would enable proper differentiation when converting to engines that support both types. For now, this is a known limitation but at least we have a fallback that works for both variants. I have added inline comments with this clarification.

PR #557, opened 2025-10-30
[Coral-Schema] handle single-element unions when the supplied column schema does not specify <uniontype>

ruolin59:
When Avro schemas define single-element unions (e.g., `["string"]` for primitives, `[{"type":"record",...}]` for array items/map values), the schema merge logic in `MergeHiveSchemaWithAvro` was incorrectly unwrapping them, changing non-nullable types to nullable types or losing the union structure entirely.

This issue occurs because some table systems (like OpenHouse) don't represent single-element unions as `uniontype<...>` in the Hive schema. When the Hive schema simply shows `string` or `struct<...>` instead of `uniontype<string>` or `uniontype<struct<...>>`, the merge logic treats them as regular types and unwraps the Avro union.

**Changes made:**

1. **`HiveSchemaWithPartnerVisitor.java`**: Added logic at the top level of the `visit()` method to:
   - Detect when the Avro partner schema has a single-element union but the Hive `TypeInfo` doesn't declare a union (via `shouldUnwrapPartner()`)
   - Temporarily unwrap the partner before the switch statement for proper type matching
   - Rewrap the result after processing to preserve the original Avro structure

2. **`SchemaUtilities.java`**: 
   - Added `wrapInSingleElementUnion()` utility method along with `extractSingleElementUnion()`
   - These methods provide a consistent API for single-element union handling

3. **Key behavior**: The fix only applies when there's a **mismatch** between Hive and Avro:
   - If Hive declares `uniontype<...>`, normal union processing occurs (no unwrap/rewrap)
   - If Hive doesn't declare `uniontype` but Avro has a single-element union, we unwrap for matching then rewrap to preserve structure

This ensures that Avro schemas with single-element unions (commonly found in `avro.schema.literal` properties) maintain their exact structure after merging, preventing unintended nullability changes.

Added two comprehensive unit tests in `MergeHiveSchemaWithAvroTests`:

1. **`shouldHandleSingleElementUnionsInAllTypes`**: 
   - Tests single-element unions in primitives (`["string"]`, `["boolean"]`), array items, and map values
   - Hive schema does NOT use `uniontype` (simulating OpenHouse-style schemas)
   - Verifies that all single-element unions are preserved in the output

2. **`shouldHandleSingleElementUnionsWithHiveUnionType`**:
   - Tests backward compatibility when Hive explicitly declares `uniontype<...>`
   - Includes both regular multi-type unions and single-element unions
   - Ensures the fix doesn't interfere with normal union processing

Both tests verify that:
- The input Avro schema structure is preserved (single-element unions remain wrapped)
- Field names, nullability, docs, and default values are maintained
- Complex nested structures (records within arrays/maps) work correctly

All existing tests continue to pass, confirming no regressions.

Also manually ran integration tests in spark, by generating a jar. Confirmed that the original symptoms are now fixed, ie. view on top of the table that had the single-element-union now show correct nullability for all fields

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/HiveSchemaWithPartnerVisitor.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/HiveSchemaWithPartnerVisitor.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2021-2022 LinkedIn Corporation. All rights reserved.
+ * Copyright 2021-2025 LinkedIn Corporation. All rights reserved.
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
@@ -115,12 +115,12 @@ public abstract class HiveSchemaWithPartnerVisitor<P, FP, R, FR> {
       default:
         throw new UnsupportedOperationException(typeInfo + " not supported");
     }
-    
+
     // Rewrap in single-element union if the partner was originally wrapped
     if (partnerWrapped) {
       resultSchema = (R) SchemaUtilities.wrapInSingleElementUnion((Schema) resultSchema);
     }
-    
+
     return resultSchema;
   }
 
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2019-2023 LinkedIn Corporation. All rights reserved.
+ * Copyright 2019-2025 LinkedIn Corporation. All rights reserved.
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
--- a/coral-schema/src/test/java/com/linkedin/coral/schema/avro/MergeHiveSchemaWithAvroTests.java
+++ b/coral-schema/src/test/java/com/linkedin/coral/schema/avro/MergeHiveSchemaWithAvroTests.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2020-2023 LinkedIn Corporation. All rights reserved.
+ * Copyright 2020-2025 LinkedIn Corporation. All rights reserved.
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
@@ -254,13 +254,12 @@ public class MergeHiveSchemaWithAvroTests {
     String hive = "struct<id:bigint,name:string,active:boolean,"
         + "items:array<struct<fooconfiguration:struct<name:string,urlvalue:string,source:string>,"
         + "barconfiguration:struct<name:string,domain:string>>>,"
-        + "metadata:map<string,struct<category:string,priority:int>>,"
-        + "tags:array<string>>";
+        + "metadata:map<string,struct<category:string,priority:int>>," + "tags:array<string>>";
 
     // Nested record schemas
-    Schema fooConfigSchema = struct("FooConfiguration", "doc-foo", "com.example.data",
-        required("name", Schema.Type.STRING), required("urlValue", Schema.Type.STRING),
-        required("source", Schema.Type.STRING));
+    Schema fooConfigSchema =
+        struct("FooConfiguration", "doc-foo", "com.example.data", required("name", Schema.Type.STRING),
+            required("urlValue", Schema.Type.STRING), required("source", Schema.Type.STRING));
 
     Schema barConfigSchema = struct("BarConfiguration", "doc-bar", "com.example.data",
         required("name", Schema.Type.STRING), required("domain", Schema.Type.STRING));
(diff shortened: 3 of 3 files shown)


coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeHiveSchemaWithAvro.java
```diff
@@ -177,13 +177,41 @@ public Schema mapKeyPartner(Schema partner) {
     @Override
     public Schema mapValuePartner(Schema partner) {
       Schema schema = SchemaUtilities.extractIfOption(partner);
-      return (schema.getType() == Schema.Type.MAP) ? schema.getValueType() : null;
+      if (schema.getType() != Schema.Type.MAP) {
+        return null;
+      }
+      Schema valueType = schema.getValueType();
```
pandaamit91: There seems to be a decent amount of repeated code in both the methods - mapValuePartner and listElementPartner. Can we abstract that out to a private method?

ruolin59: updated

pandaamit91 (approved):
LGTM! But would prefer to have another set of eyes (more experienced in the Coral repo) approve it as well.

wmoustafa:
The description is not quite adding up since "unions" are not a thing in Coral. Could you rewrite the description using "records" which is the actual primitive? That would help understand the problem better.

wmoustafa:
If you really need to use unions for the description, you may talk about unions in the Hive type system, then records in the Coral type system but you have to describe the whole flow and conversion chain with that segregation (but not mix unions with Coral type system description). If unions are not needed and we can state the problem just with records, then even better.

ruolin59:
@wmoustafa the "union" here isn't a union related to Hive, but union as related to avro literal schema definition. This is addressed in the description where it says

> This causes a type mismatch when comparing against Hive schemas, which don't represent these structures as unions, leading to incorrect schema generation where fields that should be non-nullable (like "type": "string") become nullable (like "type": ["null", "string"])

the "union" here is more referring to the fact that it's de-facto either a union of [element, null], or "single-union" where it's just a union of [element]. Not sure if that make sense

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeHiveSchemaWithAvro.java
```diff
@@ -177,13 +198,19 @@ public Schema mapKeyPartner(Schema partner) {
...
     }
 
     @Override
     public Schema listElementPartner(Schema partner) {
       Schema schema = SchemaUtilities.extractIfOption(partner);
-      return (schema.getType() == Schema.Type.ARRAY) ? schema.getElementType() : null;
+      if (schema.getType() != Schema.Type.ARRAY) {
+        return null;
+      }
+      return extractFromUnion(schema.getElementType());
```
wmoustafa: Sounds there is overlap with `extractIfOption` which is called above and also defined similarly:
```java
  static Schema extractIfOption(Schema schema) {
    if (isNullableType(schema)) {
      return getOtherTypeFromNullableType(schema);
    } else {
      return schema;
    }
  }
```

ruolin59: `extractIfOption` is completely different, in fact `extractFromUnion` calls `extractIfOption` (see line 178)

wmoustafa: Makes sense about the main logic being different. Do we need to call `extractIfOption` before `extractFromUnion` as well as inside it?

wmoustafa: Also since this fails for array elements and map values only, we must be doing this conversion correctly elsewhere. Do you know where that conversion happens (for other types)?

ruolin59: Re overlap: `extractIfOption` unwraps only nullable unions, while `extractFromUnion` normalizes nested types and already falls back to `extractIfOption` for `["null", T]`. We first call `extractIfOption` on the container (array/map) to unwrap nullable containers, then apply `extractFromUnion` to the element/value; no extra pre-unwrapping is needed for the nested type

Re where it’s handled elsewhere: container nullability is already normalized for structs/maps/lists via `extractIfOption(partner)`; union branches are handled in `unionObjectPartner` by discarding `null` before indexing; map keys are enforced as non-nullable string. The missing case was specifically array items and map values encoded as single-element unions

wmoustafa: I think we might be mixing between container nullability and unwrapping a union. It looks like the main objective of this PR is to address the case where unions of size one are not unwrapped to their underlying type. I was assuming this was handled elsewhere for cases other than map value and list element, but looking at the code, it seems that this handling is missing in most places, such as struct fields or even primitives, where unwrapping might also be needed.

ruolin59: discussed offline, this is now resolved with using a different approach to extract single-element unions

PR #558, opened 2025-11-01
Introduce Coral type system abstraction

wmoustafa:
## Summary

Introduces a **planner-agnostic Coral type system** integrated with **Calcite**, supporting all major SQL types, nested and parametric types, and precise **TIMESTAMP(p)** semantics.
Includes a full unit test suite validating nullability, nesting, and precision consistency across primitives and complex types.

###  Core Type System

**Path:** `coral-common/src/main/java/com/linkedin/coral/common/types/`

Defines:

* `CoralDataType` (base interface) and `CoralTypeKind`
* `CoralPrimitiveType`
* `CoralDecimalType` (precision, scale)
* `CoralCharType` (length)
* `CoralVarcharType` (length)
* `CoralArrayType` (element type)
* `CoralMapType` (key/value types)
* `CoralStructType`
* `CoralTimestampType` *(fractional seconds precision 0–9)*

Establishes Coral’s **independent type layer**.

### Calcite Integration

**File:** `CoralTypeToRelDataTypeConverter.java`

* Converts all Coral types to Calcite `RelDataType`
* Preserves nullability and nested ARRAY/MAP/STRUCT types
* Maps `STRING` → `VARCHAR` (Hive-style precision)
* Maps `TIMESTAMP(p)` correctly (0s, 3ms, 6µs)

### Testing

**File:** `CoralTypeSystemTest.java`

Covers:

* All primitives with nullability
* `DECIMAL` (precision/scale)
* `CHAR` / `VARCHAR` (length)
* `ARRAY` / `MAP` (element nullability)
* `STRUCT` (simple + nested)
* `TIMESTAMP` precision mapping (0, 3, 6)

### DIFF
--- a/coral-common/src/main/java/com/linkedin/coral/common/HiveToCoralTypeConverter.java
+++ b/coral-common/src/main/java/com/linkedin/coral/common/HiveToCoralTypeConverter.java
@@ -1,0 +1,134 @@
+/**
+ * Copyright 2024-2025 LinkedIn Corporation. All rights reserved.
+ * Licensed under the BSD-2 Clause license.
+ * See LICENSE in the project root for license information.
+ */
+package com.linkedin.coral.common;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.hadoop.hive.serde2.typeinfo.*;
+
+import com.linkedin.coral.common.types.*;
+
+
+/**
+ * Converts Hive TypeInfo objects to Coral data types.
+ * This enables integration between Hive's type system and Coral's type system.
+ */
+public final class HiveToCoralTypeConverter {
+
+  private HiveToCoralTypeConverter() {
+    // Utility class - prevent instantiation
+  }
+
+  /**
+   * Converts a Hive TypeInfo to a Coral data type.
+   * @param typeInfo the Hive type to convert
+   * @return the corresponding Coral data type
+   */
+  public static CoralDataType convert(TypeInfo typeInfo) {
+    if (typeInfo == null) {
+      throw new IllegalArgumentException("TypeInfo cannot be null");
+    }
+
+    switch (typeInfo.getCategory()) {
+      case PRIMITIVE:
+        return convertPrimitive((PrimitiveTypeInfo) typeInfo);
+      case LIST:
+        return convertList((ListTypeInfo) typeInfo);
+      case MAP:
+        return convertMap((MapTypeInfo) typeInfo);
+      case STRUCT:
+        return convertStruct((StructTypeInfo) typeInfo);
+      case UNION:
+        return convertUnion((UnionTypeInfo) typeInfo);
+      default:
+        throw new UnsupportedOperationException("Unsupported type category: " + typeInfo.getCategory());
+    }
+  }
+
+  private static CoralDataType convertPrimitive(PrimitiveTypeInfo type) {
+    boolean nullable = true; // Hive types are generally nullable
+
+    switch (type.getPrimitiveCategory()) {
+      case BOOLEAN:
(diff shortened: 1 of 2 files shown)


coral-common/src/main/java/com/linkedin/coral/common/types/CoralTypeKind.java
```diff
@@ -0,0 +1,41 @@
...
+package com.linkedin.coral.common.types;
+
+/**
+ * Enumeration of all supported data type kinds in the Coral type system.
+ * This provides a comprehensive set of primitive and complex types that
+ * can be mapped to various execution engines.
+ */
+public enum CoralTypeKind {
+  // Primitive numeric types
+  BOOLEAN,
```
aastha25, line 15: for backward compatibility (& a default mapping for unkown types), maybe we should also create NULL / OTHER as it exists today in [TypeConverter](https://github.com/linkedin/coral/blob/9f8dfce736e3b14aded7653cb340c75bf51dab7b/coral-common/src/main/java/com/linkedin/coral/common/TypeConverter.java#L99).

wmoustafa: I do not know if it is used in practice. Also coral conversion heavily relies on deterministic type inference. We can always add it in the future if the use case arises?

wmoustafa:
> > Yes. Probably in a separate PR, but I think this PR is all what we need for #556?
> 
> to make #556 functional, we would need the converter between hive/iceberg type system -> CoralTypeSystem since `coralTable` (hiveCoralTable, IcbergCoralTable) would now expose API - `CoralDataType getSchema()`

Oh you meant in that way. I thought that would be part of #556. Let us coordinate how to go about that.

aastha25:
Code changes look good to me. A couple of minor notes that I discovered through i-testing and will be fixing as part of a small follow up PR:

(1) `timestamp` in hive type system maps to `timestamp` in Calcite with -1 precision. For backward compatibility, especially with trino engine, we need to keep it that way. 
(2) This PR expands `uniontype' into a struct with 'n' fields. However, previously `uniontype` was being expanded into 'n+1' fields, where the extra field is a "tag" field first (INTEGER), then "field0", "field1", etc. 
(3) Following is a widely used pattern in view definitions:
```
SELECT ..., NULL AS dummyField, ... FROM t
```
and this means `dummyField` is a 'void' hive type which needs to be represented as  Coral type : NULL and subsequently Calcite type : NULL (backward compatible behavior). I will add NULL as a new coralTypeKind and add the glue logic around it.

 Rest, LGTM

coral-common/src/main/java/com/linkedin/coral/common/HiveToCoralTypeConverter.java
```diff
@@ -0,0 +1,134 @@
...
+        return DecimalType.of(decimalType.precision(), decimalType.scale(), nullable);
+      case VARCHAR:
+        VarcharTypeInfo varcharType = (VarcharTypeInfo) type;
+        return VarcharType.of(varcharType.getLength(), nullable);
+      case CHAR:
+        CharTypeInfo charType = (CharTypeInfo) type;
+        return CharType.of(charType.getLength(), nullable);
+      case VOID:
+      case UNKNOWN:
+        return PrimitiveType.of(CoralTypeKind.STRING, true); // Map to nullable string as a fallback
```
aastha25, line 90: minor nit to use `nullable` field here instead of hardcoding 'true', but not a blocker.

aastha25 (approved):
thanks for pushing this through, major milestone for Coral!

PR #560, opened 2025-11-06
[Coral-Schema] default value lowercasing

ruolin59:
This PR fixes a bug in `ToLowercaseSchemaVisitor` where field names within complex default values were not being lowercased, causing schema validation errors when creating fields with lowercased schemas.

Given an input schema with a record field that has a default value:
```
{
  "name": "Struct_Field",
  "type": {
    "type": "record",
    "name": "NestedRecord",
    "fields": [
      {"name": "firstName", "type": "string"},
      {"name": "Age", "type": "int"}
    ]
  },
  "default": {
    "firstName": "John",
    "Age": 30
  }
}
```
Before the fix:
```
{
  "name": "struct_field",
  "type": {
    "type": "record",
    "name": "nestedrecord",
    "fields": [
      {"name": "firstname", "type": "string"},
      {"name": "age", "type": "int"}
    ]
  },
  "default": {
    "firstName": "John",  // Mismatched casing
    "Age": 30             // Mismatched casing
  }
}
```
Result: `AvroTypeException: Invalid default for field struct_field`

After the fix:
```
{
  "name": "struct_field",
  "type": {
    "type": "record",
    "name": "nestedrecord",
    "fields": [
      {"name": "firstname", "type": "string"},
      {"name": "age", "type": "int"}
    ]
  },
  "default": {
    "firstname": "John",  // Correctly lowercased
    "age": 30             // Correctly lowercased
  }
}
```

The Solution:
Added `lowercaseDefaultValue()` method that recursively transforms default values based on the schema type:
- **RECORD types**: Lowercases field names in record default values
- **MAP types**: Lowercases all keys in map default values
- **ARRAY types**: Recursively processes each array element
- **Primitives**: Returns unchanged

The implementation handles both `GenericData.Record` and `Map`-based default value representations

Added a new test `testLowercaseSchemaWithComplexDefaultValues()` in `SchemaUtilitiesTests` that verifies the fix handles:
1. Simple primitive default values (no change needed)
2. Nested record default values with mixed-case field names
3. Map default values with mixed-case keys
4. Array default values containing records with mixed-case field names

The test uses input/expected schema files:
- `testLowercaseSchemaWithDefaultValues-input.avsc` - Contains fields with various default values using mixed case
- `testLowercaseSchemaWithDefaultValues-expected.avsc` - Contains the fully lowercased expected output

All existing tests continue to pass, confirming backward compatibility.

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/ToLowercaseSchemaVisitor.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/ToLowercaseSchemaVisitor.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2022-2023 LinkedIn Corporation. All rights reserved.
+ * Copyright 2022-2026 LinkedIn Corporation. All rights reserved.
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
@@ -143,7 +143,7 @@ public class ToLowercaseSchemaVisitor extends AvroSchemaVisitor<Schema> {
           Map<?, ?> mapValue = (Map<?, ?>) fieldDefaultValue; // Use wildcards to handle Utf8 keys
           Map<String, Object> lowercasedMap = new LinkedHashMap<>();
           Schema valueSchema = actualSchema.getValueType();
-          
+
           for (Map.Entry<?, ?> entry : mapValue.entrySet()) {
             String originalKey = entry.getKey().toString(); // Handle both String and Utf8
             String lowercasedKey = originalKey.toLowerCase();
@@ -159,9 +159,8 @@ public class ToLowercaseSchemaVisitor extends AvroSchemaVisitor<Schema> {
         if (fieldDefaultValue instanceof List) {
           List<Object> arrayValue = (List<Object>) fieldDefaultValue;
           Schema elementSchema = actualSchema.getElementType();
-          
-          return arrayValue.stream()
-              .map(element -> lowercaseDefaultValue(element, elementSchema))
+
+          return arrayValue.stream().map(element -> lowercaseDefaultValue(element, elementSchema))
               .collect(Collectors.toList());
         }
         return fieldDefaultValue;
@@ -193,18 +192,18 @@ public class ToLowercaseSchemaVisitor extends AvroSchemaVisitor<Schema> {
   private Map<String, Object> lowercaseRecordDefaultValue(Schema actualSchema,
       Function<String, Object> valueExtractor) {
     Map<String, Object> lowercasedRecordMap = new LinkedHashMap<>();
-    
+
     // Iterate through the lowercased schema fields
     for (Schema.Field field : actualSchema.getFields()) {
       String lowercasedFieldName = field.name();
       Object fieldValue = valueExtractor.apply(lowercasedFieldName);
-      
+
       if (fieldValue != null) {
         Object lowercasedFieldValue = lowercaseDefaultValue(fieldValue, field.schema());
         lowercasedRecordMap.put(lowercasedFieldName, lowercasedFieldValue);
       }
     }
-    
+
     return lowercasedRecordMap;
   }
 
@@ -224,7 +223,7 @@ public class ToLowercaseSchemaVisitor extends AvroSchemaVisitor<Schema> {
         return key;
       }
     }
-    
+
     return null;
   }
 }
--- a/coral-schema/src/test/java/com/linkedin/coral/schema/avro/SchemaUtilitiesTests.java
+++ b/coral-schema/src/test/java/com/linkedin/coral/schema/avro/SchemaUtilitiesTests.java
@@ -286,7 +286,7 @@ public class SchemaUtilitiesTests {
     // Compare with expected output, trimming whitespace for comparison
     String expected = TestUtils.loadSchema("testLowercaseSchemaWithDefaultValues-expected.avsc").trim();
     String actual = outputSchema.toString(true).trim();
-    
+
     Assert.assertEquals(actual, expected);
   }
 }


coral-schema/src/main/java/com/linkedin/coral/schema/avro/ToLowercaseSchemaVisitor.java
```diff
@@ -62,12 +67,163 @@ public Schema primitive(Schema primitive) {
...
+  /**
+   * Recursively lowercases field names within default values based on the schema structure.
+   * This handles complex types like records, maps, and arrays where field names appear in default values.
+   * 
+   * @param defaultValue The original default value (can be null, primitive, Map, List, etc.)
+   * @param schema The schema that describes the structure of this default value
+   * @return The default value with all field names lowercased
+   */
+  @SuppressWarnings("unchecked")
+  private Object lowercaseDefaultValue(Object defaultValue, Schema schema) {
```
aastha25: nit: rename the arguments to (`fieldDefaultValue`, `fieldSchema`) to capture that this is a per field level method.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ToLowercaseSchemaVisitor.java
```diff
@@ -62,12 +67,163 @@ public Schema primitive(Schema primitive) {
...
+    if (defaultValue == null) {
+      return null;
+    }
+
+    Schema actualSchema = schema;
+    
+    // Handle union types - get the actual schema based on the default value type
+    if (schema.getType() == Schema.Type.UNION) {
+      // For unions, the default value corresponds to the first type in the union
+      actualSchema = schema.getTypes().get(0);
```
aastha25: this assumption is not true. we 'expect' it to be the first type, but it can either. 
There's actually some utility method in coral-schema which is specifically designed to fetch the 'underlying schema' from the union type - that can be used here.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ToLowercaseSchemaVisitor.java
```diff
@@ -62,12 +67,163 @@ public Schema primitive(Schema primitive) {
...
+            String matchingKey = findMatchingKeyForLowercased(recordMap, lowercasedFieldName);
+            return matchingKey != null ? recordMap.get(matchingKey) : null;
+          });
+        }
+        // If neither Map nor GenericData.Record, return as-is
+        return defaultValue;
+
+      case MAP:
+        // For maps, lowercase the keys and recursively process values
+        if (defaultValue instanceof Map) {
```
aastha25: why do we need this if condition if its already inside the CASE? can this be simplified here and other places?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ToLowercaseSchemaVisitor.java
```diff
@@ -62,12 +67,163 @@ public Schema primitive(Schema primitive) {
...
+          Map<String, Object> lowercasedMap = new LinkedHashMap<>();
+          Schema valueSchema = actualSchema.getValueType();
+          
+          for (Map.Entry<?, ?> entry : mapValue.entrySet()) {
+            String originalKey = entry.getKey().toString(); // Handle both String and Utf8
+            String lowercasedKey = originalKey.toLowerCase();
+            Object lowercasedValue = lowercaseDefaultValue(entry.getValue(), valueSchema);
+            lowercasedMap.put(lowercasedKey, lowercasedValue);
+          }
+          return lowercasedMap;
```
aastha25, line 153: if the default value was [null, map], are we returning map by any chance? where's the logic to construct the union schema back? let's check this for other types too and ensure the spec of this method is returning the same schema as the input, just lower-cased.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ToLowercaseSchemaVisitor.java
```diff
@@ -62,12 +67,163 @@ public Schema primitive(Schema primitive) {
...
+    Schema actualSchema = schema;
+    
+    // Handle union types - get the actual schema based on the default value type
+    if (schema.getType() == Schema.Type.UNION) {
+      // For unions, the default value corresponds to the first type in the union
+      actualSchema = schema.getTypes().get(0);
+    }
+
+    switch (actualSchema.getType()) {
+      case RECORD:
```
aastha25, line 112: do we need the handling for complex union also? I understand we're stripping away the nulls for the union, but we could still have the a nullable- complex union?

aastha25 (commented):
thanks for picking this up! seems like an interesting bug!

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ToLowercaseSchemaVisitor.java
ruolin59: this is actually according to Avro specifications: https://avro.apache.org/docs/1.11.2/specification/

under "Unions":
> (Note that when a [default value](https://avro.apache.org/docs/1.11.2/specification/#schema-record) is specified for a record field whose type is a union, the type of the default value must match the first element of the union. Thus, for unions containing “null”, the “null” is usually listed first, since the default value of such unions is typically null.)

I'll update it to at least handle nullable defaults though

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ToLowercaseSchemaVisitor.java
ruolin59: the case statement is checking the case according to the schema, but this is just to make sure that the `defaultValue` is actually in accordance to the schema

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ToLowercaseSchemaVisitor.java
ruolin59: > if the default value was [null, map]
do you mean "if the field type was [null, map]"?

According to avro specs, in this case the default value SHOULD be `null`, but I'm going to add something under `case UNION` so that it can still parse properly if the default value is the non-null type. However, I'm not going to implement the complex union version of it

Not sure what you mean by "where's the logic to construct the union schema back", we aren't constructing any schemas, this is just using the schema to figure out how to parse the default values properly so the relevant parts of the default values can be lowercased

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ToLowercaseSchemaVisitor.java
ruolin59: for complex unions, we rely on the fact that the schema supplied is valid according to specs, see https://github.com/linkedin/coral/pull/560#discussion_r2677067646

PR #563, opened 2025-11-19
Integrate Coral Type System for Hive Tables

aastha25:

This PR completes the integration of the Coral type system (introduced in #558) by implementing the glue logic for two-stage schema conversion: Hive → Coral → Calcite. This architectural change enables better type system abstraction while maintaining 100% backward compatibility with existing behavior.

**Key Changes**
1. New Type Converters
  `HiveToCoralTypeConverter`: Converts Hive TypeInfo → Coral data types 
  `CoralTypeToRelDataTypeConverter`: Converts Coral data types → Calcite RelDataType
  Both converters have comprehensive unit test coverage in `HiveToCoralTypeConverterTest`
2. Refactored `HiveTable.getRowType()`
  New path (shadow mode): `getRowTypeViaCoralTypeSystem()` implements Hive → Coral → Calcite pipeline
  Legacy path (production path): `getRowTypeDirectConversion()` preserves original TypeConverter logic
  Logging: Warns on mismatch or failure (zero production impact)
New method: getCoralSchema() for Hive → Coral schema conversion
3. Critical Fixes in Coral Type System for Legacy Compatibility. Called out as pending in #558 [here](https://github.com/linkedin/coral/pull/558#issuecomment-3539593543)
  `TIMESTAMP` precision: Uses `PRECISION_NOT_SPECIFIED` (-1) to match legacy behavior (no explicit precision)
  `UNION` types: Generates {tag, field0, field1, ...} struct matching Trino's pattern (required by coalesce_struct UDF)
  `VOID` type: Maps to `SqlTypeName.NULL` (not STRING)
  Nullability: All types default to nullable per Hive semantics

 - Comprehensive unit tests in HiveToCoralTypeConverterTest covering all type categories
- Existing Integration tests pass 
- Tested against some production grade datasets

**Rollout**
This PR: Shadow mode - validate with zero risk
Follow up PR: Switch to Coral as primary after log validation

### DIFF
--- a/coral-common/src/main/java/com/linkedin/coral/common/HiveTable.java
+++ b/coral-common/src/main/java/com/linkedin/coral/common/HiveTable.java
@@ -142,27 +142,39 @@ public class HiveTable implements ScannableTable {
   /**
    * Returns the row type (schema) for this table.
    *
-   * Two conversion paths are supported:
-   * 1. Two-stage (preferred): Hive → Coral → Calcite
-   * 2. Direct (legacy): Hive → Calcite (for backward compatibility)
+   * Current behavior (validation/shadow mode):
+   * - Always returns the legacy Hive → Calcite direct conversion
+   * - Validates against the new Hive → Coral → Calcite two-stage conversion
+   * - Logs warnings if conversions don't match or if validation fails
    *
-   * The two-stage conversion enables using Coral type system as an intermediary,
-   * allowing better type system unification and testing.
+   * This allows safe validation of the new conversion path in production
+   * before switching to use it as the primary path.
    *
    * @param typeFactory Calcite type factory
    * @return RelDataType representing the table schema
    */
   @Override
   public RelDataType getRowType(RelDataTypeFactory typeFactory) {
-    // Use two-stage conversion if HiveCoralTable is available
+    // Always compute and return the legacy Hive direct conversion (production path)
+    RelDataType hiveType = getRowTypeFromHiveType(typeFactory);
+
+    // Validate against new two-stage Coral conversion (shadow/validation mode)
     try {
-      return getRowTypeFromCoralType(typeFactory);
+      RelDataType coralType = getRowTypeFromCoralType(typeFactory);
+
+      // Compare the two type representations
+      if (!hiveType.equals(coralType)) {
+        LOG.warn("Hive and Coral type conversion mismatch for table {}.{}. Hive: {}, Coral: {}", hiveTable.getDbName(),
+            hiveTable.getTableName(), hiveType, coralType);
+      }
     } catch (Exception e) {
-      // Fall back to direct conversion if two-stage conversion fails
-      LOG.warn("Two-stage type conversion failed for table {}, falling back to direct conversion. Error: {}",
-          hiveTable.getTableName(), e.getMessage(), e);
-      return getRowTypeFromHiveType(typeFactory);
+      // Log validation failure but continue with Hive type (zero production impact)
+      LOG.warn("Coral type validation failed for table {}.{}. Proceeding with Hive type. Error: {}",
+          hiveTable.getDbName(), hiveTable.getTableName(), e.getMessage(), e);
     }
+
+    // Always return the battle-tested Hive conversion result
+    return hiveType;
   }
 
   /**


coral-common/src/main/java/com/linkedin/coral/common/HiveTable.java
```diff
@@ -134,12 +139,70 @@ private void checkDaliTable() {
...
+          hiveTable.getTableName(), e.getMessage(), e);
+      return getRowTypeDirectConversion(typeFactory);
+    }
+  }
+
+  /**
+   * Two-stage conversion: Hive → Coral → Calcite.
+   * This is the preferred path when using CoralCatalog.
+   */
+  private RelDataType getRowTypeViaCoralTypeSystem(RelDataTypeFactory typeFactory) {
```
wmoustafa: Not sure we need this. I would imagine both Hive and Iceberg tables convert their types to Coral type as part as part their table API implementation. Internally we have common conversion between Coral and Calcite. IN this step, it does not matter we are coming from Hive or Iceberg.

aastha25: we are checking in hive table -> coral -> calcite type system prior to introducing the iceberg type system to roll out the coral type system incrementally. so this intermediate state is required. 

a future refactor will be applied in #556 when we introduce iceberg type system and those conversions.

wmoustafa: Won't the radius of impact of this be too large? I would keep current constant as much as possible and introduce new features on new things, then we can flip the flag for "existing" things.

If the objective is to validate, can we we continue to use the old implementation, but also compare in the background and log a non-fatal error when there is a mismatch? This way we will know from the log instead of from the user :)

coral-common/src/main/java/com/linkedin/coral/common/HiveTable.java
```diff
@@ -134,12 +139,70 @@ private void checkDaliTable() {
...
+   * allowing better type system unification and testing.
+   *
+   * @param typeFactory Calcite type factory
+   * @return RelDataType representing the table schema
+   */
   @Override
   public RelDataType getRowType(RelDataTypeFactory typeFactory) {
+    // Use two-stage conversion if HiveCoralTable is available
+    try {
+      return getRowTypeViaCoralTypeSystem(typeFactory);
```
wmoustafa: `getRowTypeFromCoralType`

coral-common/src/main/java/com/linkedin/coral/common/HiveTable.java
```diff
@@ -134,12 +139,70 @@ private void checkDaliTable() {
...
+    } catch (Exception e) {
+      // Fall back to direct conversion if two-stage conversion fails
+      LOG.warn("Two-stage type conversion failed for table {}, falling back to direct conversion. Error: {}",
+          hiveTable.getTableName(), e.getMessage(), e);
+      return getRowTypeDirectConversion(typeFactory);
```
wmoustafa: `getRowTypeFromHiveType`

coral-common/src/main/java/com/linkedin/coral/common/HiveTable.java
aastha25: this layer of coral, or for that matter coral by itself, is unaware of "new" vs "old", so it needs to be a switch at some point. Building out awareness of "new features" in coral seems like a good product investment in general, but let's pursue that outside of this PR. These changes will get rolled out with #556, Im trying to establish a checkpoint in the middle for closer monitoring. 

We're rolling out this change with a fallback, regression testing across all code paths and i-tested.

wmoustafa: I am not sure what you mean by Coral is not aware of "new" vs "old". The PR's objective is to introduce "old" and "new", so both are known to Coral. Here is what I meant.
```java
public RelDataType getRowType(RelDataTypeFactory typeFactory) {
  RelDataType hiveType = getRowTypeFromHiveType(typeFactory);

  try {
    RelDataType coralType = getRowTypeFromCoralType(typeFactory);
    if (!hiveType.equals(coralType)) {
      LOG.warn(
          "Hive and Coral type conversion mismatch for table {}. " +
          "Hive: {}, Coral: {}",
          hiveTable.getTableName(),
          hiveType,
          coralType
      );
      // Optional: throw a non-fatal exception
      throw new RuntimeException(
          "Type conversion mismatch between Hive and Coral for table " +
          hiveTable.getTableName()
      );
    }
  } catch (Exception e) {
    LOG.warn(
        "Coral validation failed for table {}. Proceeding with Hive type. Error: {}",
        hiveTable.getTableName(),
        e.getMessage(),
        e
    );
  }

  return hiveType; // Always return Hive path
}
```

aastha25: Sure I can accommodate this change. just one additional clarification, how are we thinking of accommodating this approach with `HiveCoralTable` & `IcebergCoralTable` with #556 ? 

```
HiveCoralTable implements CoralTable {

public CoralSchema getSchema() {
// integrate coral type system
}
```

and internally in `HiveTable.java`, for hive tables, we use the same approach - prefer `getRowTypeFromHiveType()` over `getRowTypeFromCoralType()` . but for iceberg tables, we prefer `getRowTypeFromCoralType()` ?

wmoustafa: Let us discuss the strategy offline and update here. There are a few options.

aastha25: I have updated this PR, PTAL

PR #587, opened 2026-03-25
Fix duplicate field error when partition column already exists in schema

simbadzina:
## Summary

- `addPartitionColsToSchema()` blindly appends partition columns without checking if they already exist in the schema, causing `AvroRuntimeException: Duplicate field X in record`
- This can happen when a Hive view projects a partition column as a regular field — the schema already contains the field, and `addPartitionColsToSchema()` tries to add it again
- The fix collects existing field names into a `Set` and skips partition columns that are already present

## Changes

- `SchemaUtilities.addPartitionColsToSchema()`: Added deduplication check before adding partition columns
- `SchemaUtilitiesTests`: Added two tests — one verifying duplicates are skipped, one verifying normal partition column addition still works

## Test plan

- [x] New test `testAddPartitionColsToSchemaSkipsDuplicates` fails without fix, passes with fix
- [x] New test `testAddPartitionColsToSchemaAddsNewPartitionCol` confirms normal behavior unchanged
- [x] Existing partition tests pass (`testBaseTableWithPartition`, `testSelectStarWithPartition`, `testSelectPartitionColumn`, `testUnionSelectStarFromPartitionTable`)


ruolin59:
This shouldn't be needed because partition columns should not exist inside the schema to begin with. Instead of here, it should be fixed on the caller side: if the schema contains the columns, no partition columns should be passed into this method in the first place

simbadzina:
> This shouldn't be needed because partition columns should not exist inside the schema to begin with. Instead of here, it should be fixed on the caller side: if the schema contains the columns, no partition columns should be passed into this method in the first place

The two callers of addPartitionColsToSchema are both inside SchemaUtilities (getCasePreservedSchemaForTable:79 and the merge branch of getAvroSchemaForTable:118-124)

Deduplication in here fit the contract "ensure partition columns are present on the schema".
For some tables the partition columns so already exist in the avro files and then independently in HMS.

Issue #575, opened 2026-01-08
Refactor core components to Use CoralTable Instead of Hive Table


aastha25:
`ParseTreeBuilder` is tightly coupled to Hive's `org.apache.hadoop.hive.metastore.api.Table` class, preventing it from working with other table formats (e.g., Iceberg). This needs to be refactored to use the unified `CoralTable` interface to enable multi-format support.

**TL;DR**: 
(1) Replace all usages of `org.apache.hadoop.hive.metastore.api.Table` in `ParseTreeBuilder`, `ParseContext`, and `HiveFunctionResolver` with the format-agnostic `CoralTable` interface. 
(2) Additionally, the methods `getDaliFunctionParams()` and `getDaliUdfDependencies()` in `HiveTable.java` parse specific table properties:
- `getDaliFunctionParams()`: Parses the `"functions"` table property (format: `"f:c1 g:c2"` mapping function name to class)
- `getDaliUdfDependencies()`: Parses the `"dependencies"` table property (format: `"o1:m1:v1 o2:m2:v2?transitive=false"`)

This will enable Coral to parse and translate SQL from both Hive and Iceberg tables using a unified API.

**Background:**
Coral is introducing `CoralTable` as a unified abstraction for representing tables across different formats (Hive, Iceberg, etc.) as part of https://github.com/linkedin/coral/pull/556.  However, `ParseTreeBuilder` - a core component that converts view SQL -> Hive AST (no schema dependency, only view metadatata -  definition & properties),  - is still deeply integrated with Hive's metastore `Table` class throughout its API surface and internal implementation.

**Minimal Example:**
### Before (Current - Hive-specific):
```java
// Client code
import org.apache.hadoop.hive.metastore.api.Table;

Table hiveTable = hiveMetastore.getTable("mydb", "mytable");
ParseTreeBuilder builder = new ParseTreeBuilder(functionResolver);
SqlNode sqlNode = builder.process("SELECT * FROM mytable", hiveTable);
```

### After (Desired - Format-agnostic):
```java
// Client code - works with both Hive and Iceberg
import com.linkedin.coral.common.catalog.CoralTable;
import com.linkedin.coral.common.catalog.HiveCoralTable;

// For Hive tables
org.apache.hadoop.hive.metastore.api.Table hiveTable = hiveMetastore.getTable("mydb", "mytable");
CoralTable coralTable = new HiveCoralTable(hiveTable);

// OR for Iceberg tables (when available)
[...]

PR #590, opened 2026-04-03
Add catalog SPI modules and refactor coral-common to SPI dispatch

wmoustafa:

This PR introduces a 3-module catalog SPI architecture that decouples Hive and Iceberg implementations from the core classpath. Currently, `coral-common` bundles `hive-metastore`, `hadoop-common`, and `iceberg-*` as `api` dependencies, which means every consumer transitively pulls all of them. This refactor isolates catalog-specific code behind an SPI boundary.

**New modules:**

- **`coral-catalog-spi`** — Pure interfaces (`CoralCatalog`, `CoralTable`, `TableType`), the `com.linkedin.coral.common.types` package, and new SPI contracts (`CoralCalciteTableAdapterFactory`, `CoralCalciteTableAdapterRegistry`). Zero Hive/Iceberg dependencies.
- **`coral-catalog-hive`** — Hive catalog implementations (`HiveTable`, `HiveCalciteTableAdapter`, `HiveCalciteViewAdapter`, `HiveToCoralTypeConverter`, `TypeConverter`) with `HiveCalciteTableAdapterFactory` registered via `META-INF/services`.
- **`coral-catalog-iceberg`** — Iceberg catalog implementations (`IcebergTable`, `IcebergCalciteTableAdapter`, `IcebergToCoralTypeConverter`, `IcebergHiveTableConverter`) with `IcebergCalciteTableAdapterFactory` registered via `META-INF/services`.

**`coral-common` changes:**

- `CoralDatabaseSchema.getTable()` now uses `CoralCalciteTableAdapterRegistry` for SPI-based dispatch instead of `instanceof` checks.
- 9 implementation classes converted to `@Deprecated` wrappers delegating to the new modules.
- Hive/Hadoop/Iceberg dependencies demoted from `api` to `implementation` scope.

**Backward compatibility:** All original FQCNs still resolve — interfaces/types kept the same package, and implementation classes have `@Deprecated` wrappers at the old locations. Consumers relying on transitive `api` exposure of `hive-metastore` or `iceberg-*` from `coral-common` may need to add explicit dependencies.

**Known limitations:**
- `ToRelConverter.toSqlNode(String, Table)` still couples the core to Hive's `Table` type (issue #575).
- `IcebergHiveTableConverter` in `coral-catalog-iceberg` retains `implementation` deps on Hive as a temporary bridge.

- `./gradlew compileJava` — all modules compile successfully
- `./gradlew :coral-common:test` — existing tests pass
- Verified `ServiceLoader` discovers both `HiveCalciteTableAdapterFactory` and `IcebergCalciteTableAdapterFactory`
- Confirmed `coral-catalog-spi` has no Hive/Iceberg on compile classpath
- Verified deprecated wrapper classes at old FQCNs still resolve and delegate correctly


coral-common/src/main/java/com/linkedin/coral/common/CoralDatabaseSchema.java
```diff
@@ -94,20 +84,8 @@ public Table getTable(String name) {
...
-      // Check if it's a view
-      if (hiveTable.tableType() == VIEW) {
-        return new HiveCalciteViewAdapter(hiveTable, ImmutableList.of(CoralRootSchema.ROOT_SCHEMA, dbName));
-      } else {
-        return new HiveCalciteTableAdapter(hiveTable);
-      }
-    }
-
-    return null;
+    return CoralCalciteTableAdapterRegistry.createAdapter(coralTable,
```
ruolin59, line 87: this method throws `UnsupportedOperationException`, I'd just like to confirm that this is the new behavior we want (so no catching an return null, correct?), if so, may be a good idea to at least update the javadoc

coral-common/build.gradle
```diff
@@ -10,19 +10,25 @@ dependencies {
     }
   }
 
-  api(deps.'hive'.'hive-metastore') {
+  // New catalog SPI and implementation modules
```
ruolin59, line 13: `coral-catalog-hive` and `coral-catalog-iceberg` appear to still have api dependencies on `hive-metastore` and `iceberg-core` respectively, so the consumer would still be pulling these deps transitively right? To clarify,
1. Was the intent to demote coral-common's project deps to implementation, or is keeping them as api a deliberate compatibility choice?
2. If it's deliberate, which downstream consumers are expected to migrate to depending on `coral-catalog-spi` directly?
3. Can the redundant implementation block at lines 21-34 be dropped?

ruolin59 (commented):
actually appears to mostly be moving things around and refactoring, but have a few questions about the module rewiring, and also, it would be good to add some new unit tests to cover the new items introduced in this pr, including
- `CoralCalciteTableAdapterRegistryTest`: priority ordering, register() merging with ServiceLoader, UnsupportedOperationException on no-match.
- One smoke test per new factory confirming supports() / createAdapter() return the expected adapter class for a representative CoralTable.
- One test per deprecated wrapper confirming new <old-FQCN>(...) produces an instance assignable to the new type and exposes the same public surface.

coral-catalog-hive/src/main/java/com/linkedin/coral/catalog/hive/HiveTable.java
```diff
@@ -0,0 +1,99 @@
...
+    String hiveTableType = table.getTableType();
+    if (hiveTableType != null && hiveTableType.toUpperCase().contains("VIEW")) {
+      return TableType.VIEW;
+    }
+    return TableType.TABLE;
+  }
+
+  /**
+   * INTERNAL API
+   * @deprecated This method is for internal use only and will be removed in a future release.
```
simbadzina, line 68: `@deprecated` annotation missing.

Also consider `@InternalApi`

coral-catalog-spi/src/main/java/com/linkedin/coral/common/catalog/CoralCalciteTableAdapterFactory.java
```diff
@@ -0,0 +1,54 @@
...
+  /**
+   * Returns the priority of this factory. Lower values indicate higher priority.
+   * When multiple factories support the same CoralTable type, the one with the
+   * lowest priority value is used.
+   *
+   * @return Priority value (default 100)
+   */
+  default int priority() {
+    return 100;
+  }
```
simbadzina, line 53: If more factories are added, how are the priorities set?
I don't understand the value of the priority.

coral-catalog-spi/src/main/java/com/linkedin/coral/common/catalog/CoralCalciteTableAdapterRegistry.java
```diff
@@ -0,0 +1,77 @@
...
+    List<CoralCalciteTableAdapterFactory> factories = new ArrayList<>(EXPLICIT_FACTORIES);
+
+    ServiceLoader<CoralCalciteTableAdapterFactory> loader = ServiceLoader.load(CoralCalciteTableAdapterFactory.class);
+    for (CoralCalciteTableAdapterFactory factory : loader) {
+      factories.add(factory);
+    }
+
+    return factories.stream().sorted(Comparator.comparingInt(CoralCalciteTableAdapterFactory::priority))
+        .collect(Collectors.toList());
+  }
```
simbadzina, line 76: Should explicit factories be made to take precedence over the ServiceLoaded ones? Or we rely on the priority always?

coral-catalog-spi/src/main/java/com/linkedin/coral/common/catalog/CoralCalciteTableAdapterFactory.java
simbadzina: Looks like just for explicit factories is where the priority override can be used?
https://github.com/linkedin/coral/pull/590/changes#r3283485632

coral-catalog-spi/src/main/java/com/linkedin/coral/common/catalog/CoralCalciteTableAdapterRegistry.java
```diff
@@ -0,0 +1,77 @@
...
+   * @return A Calcite Table adapter
+   * @throws UnsupportedOperationException if no factory supports the given table type
+   */
+  public static Table createAdapter(CoralTable coralTable, List<String> schemaPath) {
+    List<CoralCalciteTableAdapterFactory> allFactories = getAllFactories();
+    for (CoralCalciteTableAdapterFactory factory : allFactories) {
+      if (factory.supports(coralTable)) {
+        return factory.createAdapter(coralTable, schemaPath);
+      }
+    }
```
simbadzina, line 61: Can we cache the already discovered factories and try those first. So that the Service loader is not being invoked for each getTable.

coral-catalog-spi/src/main/java/com/linkedin/coral/common/catalog/CoralCalciteTableAdapterRegistry.java
```diff
@@ -0,0 +1,77 @@
...
+ * {@link ServiceLoader} and dispatches table adapter creation to the appropriate factory.
+ *
+ * <p>Factories are ordered by {@link CoralCalciteTableAdapterFactory#priority()} (lowest first).
+ * The first factory that {@link CoralCalciteTableAdapterFactory#supports(CoralTable) supports}
+ * a given CoralTable is used to create the adapter.
+ *
+ * <p>In environments with complex classloader hierarchies (e.g., Spark), factories can also
+ * be registered explicitly via {@link #register(CoralCalciteTableAdapterFactory)}.
+ */
+public final class CoralCalciteTableAdapterRegistry {
```
simbadzina, line 29: Can we add tests for this?

PR #593, opened 2026-04-08
[Coral-Spark] Rewrite VARBINARY to BINARY in SparkSqlRewriter

c-h-afzal:
Calcite's type system uses VARBINARY for binary data, but Spark's SQL parser does not recognize VARBINARY as a valid cast target — it uses BINARY instead. When Hive views call base64() on string columns, Calcite inserts an implicit CAST(... AS VARBINARY) which produces unparseable Spark SQL, causing CoralSpark translation failures and downstream function registry poisoning via the DaliSpark Hive fallback path.

This follows the same pattern already used by TrinoSqlRewriter, which rewrites BINARY/VARBINARY to Trino's VARBINARY in its convertTypeSpec method. Here we do the equivalent for Spark: rewrite VARBINARY to BINARY in SparkSqlRewriter.visit(SqlDataTypeSpec).

- Verified existing test cases pass and added a new test case.
- Also, ran the regression DAGs and found zero regressions. The baseline version tested against was 2.3.9

### DIFF
--- a/coral-spark/src/test/java/com/linkedin/coral/spark/CoralSparkTest.java
+++ b/coral-spark/src/test/java/com/linkedin/coral/spark/CoralSparkTest.java
@@ -412,6 +412,16 @@ public class CoralSparkTest {
   }
 
   @Test
+  public void testBase64OnStringProducesBinaryCast() {
+    // base64() expects BINARY input, so Calcite inserts an implicit CAST(... AS VARBINARY)
+    // when the argument is a STRING. The SparkSqlRewriter must rewrite VARBINARY to BINARY
+    // since Spark's parser does not recognize VARBINARY.
+    RelNode relNode = TestUtils.toRelNode(String.join("\n", "", "SELECT base64(b)", "FROM complex"));
+    String targetSql = "SELECT base64(CAST(complex.b AS BINARY))\n" + "FROM default.complex complex";
+    assertEquals(createCoralSpark(relNode).getSparkSql(), targetSql);
+  }
+
+  @Test
   public void testInterval() {
     RelNode relNode = TestUtils.toRelNode("SELECT CAST('2021-08-31' AS DATE) + INTERVAL '7' DAY FROM default.complex");
     String targetSql = "SELECT (CAST('2021-08-31' AS DATE) + INTERVAL '7' DAY)\n" + "FROM default.complex complex";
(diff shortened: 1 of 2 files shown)


coral-spark/src/test/java/com/linkedin/coral/spark/CoralSparkTest.java
```diff
@@ -411,6 +411,16 @@ public void testCastAsBinary() {
...
   }
 
+  @Test
+  public void testBase64OnStringProducesBinaryCast() {
+    // base64() expects BINARY input, so Calcite inserts an implicit CAST(... AS VARBINARY)
+    // when the argument is a STRING. The SparkSqlRewriter must rewrite VARBINARY to BINARY
+    // since Spark's parser does not recognize VARBINARY.
+    RelNode relNode = TestUtils.toRelNode(String.join("\n", "", "SELECT base64(b)", "FROM complex"));
+    String sparkSql = createCoralSpark(relNode).getSparkSql();
+    assertFalse(sparkSql.contains("VARBINARY"), "Spark SQL should not contain VARBINARY");
```
ruolin59: minor nit: The surrounding tests use `assertEquals` for a stronger validation, might be a good idea to do that here as well

coral-spark/src/test/java/com/linkedin/coral/spark/CoralSparkTest.java
```diff
@@ -411,6 +411,16 @@ public void testCastAsBinary() {
     assertEquals(createCoralSpark(relNode).getSparkSql(), targetSql);
   }
 
+  @Test
+  public void testBase64OnStringProducesBinaryCast() {
+    // base64() expects BINARY input, so Calcite inserts an implicit CAST(... AS VARBINARY)
+    // when the argument is a STRING. The SparkSqlRewriter must rewrite VARBINARY to BINARY
+    // since Spark's parser does not recognize VARBINARY.
+    RelNode relNode = TestUtils.toRelNode(String.join("\n", "", "SELECT base64(b)", "FROM complex"));
```
ruolin59, line 419: minor nit: would be good to test explicit `CAST(b AS VARBINARY)` here instead of the implicit VARBINARY cast

aastha25:
Congrats on your first Coral PR @c-h-afzal ! 

code changes look good and thanks for running regression testing. Please polish the PR description to match the template adopted by other merged PR such as - https://github.com/linkedin/coral/pull/597. 

(1) Add [Coral-Spark] at the beginning of the PR title
(2) add a clear 'motivation for the change' section and a clear testing done heading and testing evidence under.

aastha25:
can you also amend the commit history to not have claude as a co-author? its fine to use LLM for code writing but the commit history for the project should have core contributors. This seems to be the norm for other OSS projects as well.

aastha25 (approved):

PR #597, opened 2026-04-14
[Coral-Schema] Fix NoSuchMethodError for DECIMAL columns: replace Jackson1Utils with String.valueOf

ruolin59:
In `coral-schema`, three classes handle the `DECIMAL` type by constructing a Jackson 1.x `JsonNode` via `JsonNodeFactory` and passing it to `avroutil1`'s `Jackson1Utils.toJsonString`:
  - `MergeHiveSchemaWithAvro`
  - `RelDataTypeToAvroType`
  - `TypeInfoToAvroSchemaConverter`

When a downstream consumer has its own `avroutil1` jar on the classpath, the classloader resolves `Jackson1Utils` from that jar — which only has `toJsonString(org.codehaus.jackson.JsonNode)`. If the `JsonNode` was created from a different (e.g. relocated) copy of Jackson 1, the signatures do not match and the JVM throws `NoSuchMethodError` at runtime.

The fix replaces `Jackson1Utils.toJsonString(factory.numberNode(int))` with `String.valueOf(int)`. Since precision and scale are primitive `int` values, `String.valueOf` produces the identical JSON integer string with no Jackson or `avroutil1` dependency. This removes all three affected `Jackson1Utils` and `JsonNodeFactory` imports.

- Extended `MergeHiveSchemaWithAvroTests` with two new DECIMAL cases: zero scale (`DECIMAL(10,0)`) and max Hive precision (`DECIMAL(38,10)`), in addition to the existing `DECIMAL(4,2)` case.
- Added `testDecimalTypeField` to `RelDataTypeToAvroTypeTests` using the existing `basedecimal` table (`DECIMAL(2,1)`), with a new expected schema resource file asserting that `precision` and `scale` are encoded as JSON integers.
- Added a new `TypeInfoToAvroSchemaConverterTests` class with three direct unit tests (no Hive metastore required): `DECIMAL(10,5)`, `DECIMAL(10,0)`, and `DECIMAL(38,10)`.
- All tests pass via `./gradlew :coral-schema:test`.

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeHiveSchemaWithAvro.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeHiveSchemaWithAvro.java
@@ -10,7 +10,6 @@ import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
-import com.linkedin.avroutil1.compatibility.Jackson1Utils;
 
 import org.apache.avro.Schema;
 import org.apache.hadoop.hive.serde2.avro.AvroSerDe;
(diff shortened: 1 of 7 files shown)


aastha25 (approved):

aastha25:
please also check if it would be good to exclude jackson from the coral artifacts. This might need some downstream assessment as well - for ex: are downstream workflows relying on coral packaged jackson. Can be explored in a follow up PR, not a blocker to merge this.

PR #598, opened 2026-04-14
Add isIcebergBacked() capability hook to CoralTable

yyy1000:
## Motivation
Schema converters need to distinguish Iceberg-backed tables from Hive tables to choose the correct merge path (Iceberg-aware `MergeCoralSchemaWithAvro` vs Hive-based `MergeHiveSchemaWithAvro`). Without a capability hook, callers would need `instanceof` checks, which break when Spark-side wrappers like `LiOpenHouseCoralTable` are introduced — those implement `CoralTable` directly and are not instances of `IcebergTable`.

## Summary
- Add `default boolean isIcebergBacked()` on `CoralTable` returning `false`
- Override in `IcebergTable` to return `true`
- `HiveTable` inherits the default (`false`)

## How this will be used
In future PRs, `SchemaUtilities.getAvroSchemaForTable(CoralTable, strictMode)` and `RelToAvroSchemaConverter.getTableScanSchema()` will use `coralTable.isIcebergBacked()` to route to the Iceberg-aware merge path. Spark-side wrappers like `LiOpenHouseCoralTable` will also override this to return `true`.

## Test plan
- [x] New `CoralTableIsIcebergBackedTest` verifying: default interface method returns `false`, `HiveTable` returns `false`
- [x] `coral-common` test suite passes

### DIFF
--- a/coral-common/src/main/java/com/linkedin/coral/common/catalog/CoralTable.java
+++ b/coral-common/src/main/java/com/linkedin/coral/common/catalog/CoralTable.java
@@ -48,4 +48,15 @@ public interface CoralTable {
    * @return CoralDataType representing the table schema (typically a StructType)
    */
   CoralDataType getSchema();
+
+  /**
+   * Returns whether this table is backed by Iceberg.
+   * Used by schema converters to choose the Iceberg-aware merge path
+   * instead of the Hive merge path.
+   *
+   * @return true if the table is Iceberg-backed, false otherwise
+   */
+  default boolean isIcebergBacked() {
+    return false;
+  }
 }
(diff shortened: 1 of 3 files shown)


wmoustafa:
I would be strongly against leaking table format details to the catalog API. The sole reason for the API is to abstract those details out. Let us find another way to achieve the requirement.

wmoustafa (changes requested):
We should not expose implementation/format specific APIs in the catalog interface.

yyy1000:
> We should not expose implementation/format specific APIs in the catalog interface.

The catalog inferace already contains HiveTable and Iceberg table these classes, which is the same as what the PR contains. You can use reflection to get whether a table is iceberg table or hive table.
Wdyt?

aastha25:
discussed offline, this PR is not required. Junhao will close it.

PR #599, opened 2026-04-15
Add coral-benchmark module for cross-dialect translation testing

wmoustafa:
# Add coral-benchmark module for cross-dialect translation testing

## Summary

Adds a new `coral-benchmark` module that provides a framework for testing Coral translations end-to-end between any pair of supported dialects (Hive, Spark, Trino). This is the API design — interfaces, data types, and the orchestration layer — ready for implementation to be filled in.

## Motivation

Coral's existing tests verify individual dialect converters in isolation (e.g., Hive-to-Trino, Hive-to-Spark). There is no unified way to test arbitrary dialect-to-dialect translations, validate that translated queries are syntactically valid on the target engine, or compare query results across engines for semantic equivalence. This module closes that gap.

## Design

The framework is built around two SPIs and three verification levels:

**SPIs:**
- `DialectPlugin` — wraps existing Coral converters (`HiveToRelConverter`, `TrinoToRelConverter`, `RelToTrinoConverter`, `CoralSpark`) behind a uniform `toRelNode()` / `toDialectSql()` interface.
- `EnginePlugin` — provides execution capabilities (`createTable`, `loadData`, `explain`, `execute`) for embedded engines (Spark, Trino, Hive).

**Verification levels (escalating):**
1. **TRANSLATION** — source SQL -> Coral IR -> target SQL completes without error.
2. **EXPLAIN** — the translated SQL passes the target engine's EXPLAIN (syntax + planning validation).
3. **RESULT_SET** — query results from source and target engines are compared for semantic equivalence, with configurable tolerances for floating-point precision, row ordering, NULL handling, and type widening.

**Supporting components:**
- `InMemoryCatalog` — an in-memory `CoralCatalog` implementation using the Coral type system (`StructType`, `PrimitiveType`, etc.) so tests have no external metastore dependency.
- `RowSet` / `ResultSet` — typed tabular data containers for loading test data and capturing query results.
- `ResultSetComparator` with `ComparisonConfig` — configurable comparison logic for cross-engine result equivalence.
- `TranslationTestSuite` — builder-configured orchestrator that reads `.sql` files from a query directory and runs each through the configured pipeline.
- `TestReport` / `QueryTestResult` — structured reporting with per-query outcomes and aggregate failure categorization (translation error, explain failure, result mismatch).

## What's included

- 15 Java source files across 5 packages (`spi`, `catalog`, `data`, `comparison`, `suite`)
- `build.gradle` with `api` dependency on `:coral-common`
- The testing spec document (`coral-benchmark-spec.md`)
- Module registered in `settings.gradle`
- Compiles cleanly via `./gradlew :coral-benchmark:compileJava`

## What's not included (yet)

- Concrete `DialectPlugin` implementations (Hive, Spark, Trino)
- Concrete `EnginePlugin` implementations (embedded Spark, Trino)
- `ResultSetComparator.compare()` implementation
- `TranslationTestSuite.run()` implementation
- Query corpus (`.sql` test files)

### DIFF
--- a/coral-benchmark/build.gradle
+++ b/coral-benchmark/build.gradle
@@ -1,0 +1,5 @@
+apply plugin: 'java-library'
+
+dependencies {
+  api project(path: ':coral-common')
+}
--- a/coral-benchmark/coral-benchmark-spec.md
+++ b/coral-benchmark/coral-benchmark-spec.md
@@ -1,0 +1,281 @@
+# Coral Benchmark: Cross-Dialect Integration Testing Framework
+
+## 1. Purpose
+
+A new module (`coral-benchmark`) that tests Coral translations end-to-end: from any supported source dialect to any supported target dialect. The framework verifies both **syntactic correctness** (the translated query is valid in the target dialect) and **semantic correctness** (the translated query produces equivalent results on real engine execution).
+
+## 2. Design Principles
+
+- **Grounded in existing APIs.** The framework builds on `CoralCatalog`, `CoralTable`, and the Coral type system (`CoralDataType`, `CoralTypeKind`). It does not invent parallel abstractions for catalog or types.
+- **In-memory by default.** Tests run against an in-memory `CoralCatalog` implementation with no external metastore dependency.
+- **Dialect-agnostic core, dialect-specific plugins.** The core framework knows nothing about Hive, Spark, or Trino. Each dialect contributes an implementation of a small SPI that the core orchestrates.
+- **Incremental verification levels.** Users choose the level of verification appropriate to their needs, from pure IR round-trip checks up to full result-set comparison on live engines.
+
+## 3. Core Concepts
+
+### 3.1 Catalog Setup
+
+Tests declare their table schemas using the Coral type system and register them in an in-memory catalog.
+
+**In-memory catalog.** A concrete `InMemoryCatalog implements CoralCatalog` that holds tables in a `Map<namespace, Map<tableName, CoralTable>>`. Provides a builder API for test ergonomics:
+
+```java
+InMemoryCatalog catalog = InMemoryCatalog.builder()
+    .createNamespace("db")
+    .addTable("db", "users", StructType.of(Arrays.asList(
+        StructField.of("id", PrimitiveType.of(CoralTypeKind.INT, true)),
+        StructField.of("name", PrimitiveType.of(CoralTypeKind.STRING, true)),
+        StructField.of("created", TimestampType.of(3, true)),
+        StructField.of("tags", ArrayType.of(PrimitiveType.of(CoralTypeKind.STRING, true), true))
+    ), true))
+    .addTable("db", "events", StructType.of(Arrays.asList(
+        StructField.of("user_id", PrimitiveType.of(CoralTypeKind.INT, true)),
+        StructField.of("event_type", PrimitiveType.of(CoralTypeKind.STRING, true)),
+        StructField.of("payload", MapType.of(
+            PrimitiveType.of(CoralTypeKind.STRING, true),
+            PrimitiveType.of(CoralTypeKind.STRING, true), true)),
+        StructField.of("ts", TimestampType.of(3, true))
+    ), true))
+    .build();
+```
+
+All column types are expressed through the existing Coral type hierarchy (`PrimitiveType`, `StructType`, `ArrayType`, `MapType`, `DecimalType`, `TimestampType`, etc.). No raw strings for types.
+
+### 3.2 Query Corpus
+
+Queries are plain SQL SELECT statements stored as individual `.sql` files organized by source dialect:
+
+```
+coral-benchmark/src/test/resources/
+  queries/
+    hive/
+      group_by_count.sql
+      join_with_filter.sql
+      nested_struct_access.sql
+    trino/
+      group_by_count.sql
+      ...
+    spark/
+      ...
+```
+
+Each file contains a single SELECT statement written in the source dialect's syntax. File names are descriptive of the query pattern being tested. The same logical query may appear under multiple dialect directories, written in each dialect's native syntax.
+
+### 3.3 Dialect SPI
+
+Each dialect plugs into the framework by implementing a single interface:
+
+```java
+public interface DialectPlugin {
+
+    /** Identifier for this dialect (e.g., HIVE, SPARK, TRINO). */
+    Dialect dialect();
+
+    /** Parse a SQL string in this dialect and produce a Coral IR RelNode. */
+    RelNode toRelNode(String sql, CoralCatalog catalog);
+
+    /** Convert a Coral IR RelNode to a SQL string in this dialect. */
+    String toDialectSql(RelNode relNode, CoralCatalog catalog);
+}
+```
+
+Implementations wrap the existing converters:
+- **Hive**: `HiveToRelConverter` / `CoralRelToSqlNodeConverter` (Hive dialect)
+- **Trino**: `TrinoToRelConverter` / `RelToTrinoConverter`
+- **Spark**: `HiveToRelConverter` (Spark SQL parses as Hive) / `CoralSpark`
+
+Plugins are discovered by `ServiceLoader` or registered explicitly at suite construction time.
+
+### 3.4 Engine SPI (for execution-level verification)
+
+For tests that go beyond syntactic validation and actually execute queries, each engine provides:
+
+```java
+public interface EnginePlugin {
+
+    /** Which dialect this engine natively executes. */
+    Dialect dialect();
+
+    /** Start the engine (may be a no-op for remote engines). */
+    void start();
+
+    /** Create the given table schema in the engine's catalog, ready for data loading. */
+    void createTable(String namespace, String tableName, CoralDataType schema);
+
+    /** Load row data into a previously created table. */
+    void loadData(String namespace, String tableName, RowSet data);
+
+    /** Run EXPLAIN on a query and return success/failure. Validates syntax + planning. */
+    ExplainResult explain(String sql);
+
+    /** Execute a query and return its result set. */
+    ResultSet execute(String sql);
+
+    /** Tear down the engine. */
+    void stop();
+}
+```
+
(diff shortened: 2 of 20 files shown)


coral-benchmark/src/main/java/com/linkedin/coral/benchmark/spi/DialectPlugin.java
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+   * Parses a SQL string written in this dialect and converts it to Coral IR.
+   *
+   * @param sql    a SELECT statement in this dialect's syntax
+   * @param catalog the catalog providing table metadata for query resolution
+   * @return the Calcite {@link RelNode} representing the Coral intermediate representation
+   * @throws IllegalArgumentException if the SQL cannot be parsed in this dialect
+   */
+  RelNode toRelNode(String sql, CoralCatalog catalog);
```
ljfgem, line 44: What about passing catalog at construction or via `init(CoralCatalog)`. The SPI methods simplify to `toRelNode(String sql)` / `toDialectSql(RelNode relNode)`.

coral-benchmark/coral-benchmark-spec.md
```diff
@@ -0,0 +1,281 @@
...
+```java
+RowSet userData = RowSet.builder(usersSchema)   // usersSchema is the StructType from catalog setup
+    .addRow(1, "alice", Timestamp.valueOf("2024-01-15 10:00:00"), Arrays.asList("admin", "user"))
+    .addRow(2, "bob",   Timestamp.valueOf("2024-03-20 14:30:00"), Arrays.asList("user"))
+    .build();
+```
+
+Values are Java objects matching the Coral type mapping (INT -> Integer, STRING -> String, ARRAY -> List, MAP -> Map, STRUCT -> Object[], etc.).
+
+## 4. Verification Levels
```
ljfgem, line 139: What about adding lightweight ASCII diagrams (that Claude Code can generate) to make things clearer?

coral-benchmark/src/main/java/com/linkedin/coral/benchmark/spi/DialectPlugin.java
```diff
@@ -0,0 +1,57 @@
...
+   */
+  Dialect dialect();
+
+  /**
+   * Parses a SQL string written in this dialect and converts it to Coral IR.
+   *
+   * @param sql    a SELECT statement in this dialect's syntax
+   * @param catalog the catalog providing table metadata for query resolution
+   * @return the Calcite {@link RelNode} representing the Coral intermediate representation
+   * @throws IllegalArgumentException if the SQL cannot be parsed in this dialect
```
ljfgem, line 42: The orchestrator in `TranslationTestSuite.run()` will need to catch different exception types at different stages and map them to `FailureCategory`. What about defining a `BenchmarkException` hierarchy (`TranslationException`, `EngineException`) that plugins are contracted to throw.

coral-benchmark/coral-benchmark-spec.md
```diff
@@ -0,0 +1,281 @@
+# Coral Benchmark: Cross-Dialect Integration Testing Framework
```
ljfgem, line 1: Let's also add this new module (WIP) to the README.md file?

ljfgem (commented):
Thanks for the PR!

coral-benchmark/src/main/java/com/linkedin/coral/benchmark/spi/DialectPlugin.java
1fanwang, line 44: Curious how you're thinking about config/context flowing through the SPI. Looking at the existing converters:

- `RelToTrinoConverter` consumes a `Map<String, Boolean>` of `CoralTrinoConfigKeys` (e.g. `SUPPORT_LEGACY_UNNEST_ARRAY_OF_STRUCT`) — exactly the kind of semantic knob a correctness suite would want to toggle per test.
- `CoralSpark.create` threads an HMS client through.

Plugins could own those internally, but then the benchmark caller can't exercise a given converter across its config surface. Wondering if something like `DialectPlugin<Ctx>` or a small per-plugin config object earns its keep once the first real impls land — or whether you'd rather keep the core SPI narrow and handle the variance via plugin-specific sub-interfaces. No strong view, just flagging before it ossifies.

coral-benchmark/build.gradle
```diff
@@ -0,0 +1,5 @@
+apply plugin: 'java-library'
+
+dependencies {
+  api project(path: ':coral-common')
```
1fanwang, line 4: Small topology thought — today the module only depends on `:coral-common`, which keeps it dialect-agnostic. §9 of the spec suggests it'll grow deps on `coral-hive`, `coral-trino`, `coral-spark` once the concrete plugins land.

Have you considered hosting each `DialectPlugin` in its native module (e.g. `coral-hive` shipping `HiveDialectPlugin` with a `META-INF/services` entry, picked up by the `ServiceLoader` path you've already sketched)? Keeps this module thin, and a future contributor adding a new dialect wouldn't need to touch `coral-benchmark`. Defer if you've already weighed it — just wanted to surface while the topology is still fluid.

1fanwang (commented):
The overall direction looks good, thanks for continuing this effort!

In-memory catalog, SPI-based dialects, and escalating verification levels (especially RESULT_SET for semantic equivalence against real engines) — I think this is the shape that's been missing on the OSS side for exactly the kinds of subtle cross-engine differences that bite in practice: NULL handling, UNION coercion, timestamp precision. Skeleton-first with a spec doc is a nice way to surface the API before anyone invests in plugin plumbing.

One forward-looking thought: the long-term value of this framework probably lives in the query corpus more than the framework itself — curious if you have a mental model for how the corpus grows over time (in-repo seed vs. per-consumer contributions, which translation-divergence categories to prioritize first). Happy to pick that up in a follow-up.

coral-benchmark/src/main/java/com/linkedin/coral/benchmark/spi/DialectPlugin.java
wmoustafa: Went with a `DialectPluginProvider` / `DialectPlugin` split. The provider is the no-arg `ServiceLoader` entry point with `create(CoralCatalog)`; the plugin holds the catalog as a final field, so `toRelNode(sql)` and `toDialectSql(rel)` no longer need it threaded through. Keeps plugins immutable while still supporting standard SPI discovery.

coral-benchmark/coral-benchmark-spec.md
wmoustafa: Thanks for the suggestion. Done.

coral-benchmark/src/main/java/com/linkedin/coral/benchmark/spi/DialectPlugin.java
wmoustafa: Considered this. With the assumption that queries are pre-vetted before admission to the benchmark, every failure is attributable to translation gap or to engine result mismatch; there is no “is this the query’s fault?” classification for plugins to make.

That means the orchestrator can map failures to `FailureCategory` purely from the call site of each SPI invocation:

* A throw out of `toRelNode` / `toDialectSql` is `TRANSLATION_ERROR`.
* `EXPLAIN_FAILURE` and `RESULT_MISMATCH` already come from result types.

So we don’t think a typed exception hierarchy is adding much beyond the enums. Plugins can let underlying converter exceptions bubble naturally, and the orchestrator can map those to the above failure categories.

coral-benchmark/coral-benchmark-spec.md
wmoustafa: Done.

coral-benchmark/src/main/java/com/linkedin/coral/benchmark/spi/DialectPlugin.java
wmoustafa: Good point. With the factory split (check the most recent revision of the PR), my current thinking is that dialect-specific config (`CoralTrinoConfigKeys`, HMS clients, etc.) lives in concrete provider constructors, e.g., `new HiveDialectPluginProvider(hmsClient)`, `new TrinoDialectPluginProvider(configKeys)` so the SPI itself stays narrow and the config variance lives in the dialect's own provider. That seems to absorb the cases you mentioned without baking a `Ctx` parameter into `DialectPlugin` itself.

coral-benchmark/build.gradle
wmoustafa: Good catch. Trimmed the spec sections that committed to plugin/engine hosting in coral-benchmark. 8 no longer names plugin classes inside this module, and 9 no longer lists future deps on `:coral-hive` / `:coral-trino` / `:coral-spark`. The hosting decision (this module vs. native modules vs. per-dialect bridge modules) is intentionally left open so it can be made against the first real plugin rather than against a sketch. Will pick it up when the first concrete plugin is on the table.

wmoustafa:
> The overall direction looks good, thanks for continuing this effort!
> 
> In-memory catalog, SPI-based dialects, and escalating verification levels (especially RESULT_SET for semantic equivalence against real engines) — I think this is the shape that's been missing on the OSS side for exactly the kinds of subtle cross-engine differences that bite in practice: NULL handling, UNION coercion, timestamp precision. Skeleton-first with a spec doc is a nice way to surface the API before anyone invests in plugin plumbing.
> 
> One forward-looking thought: the long-term value of this framework probably lives in the query corpus more than the framework itself — curious if you have a mental model for how the corpus grows over time (in-repo seed vs. per-consumer contributions, which translation-divergence categories to prioritize first). Happy to pick that up in a follow-up.

Thanks. Fully agree the long-term value is in the corpus more than the framework. The framework PR is the first step towards that. Yes, I have some rough ideas. Mainly we want to maximize diversity in the categories we test (e.g., nested queries, CTEs, lateral joins, null handling, etc), achieve high coverage of operators, and address varying shapes of schemas. Happy to discuss further with you and plan the work.

1fanwang (approved):
lgtm overall, we can get this in and keep iterating on it

PR #600, opened 2026-04-15
Add MergeCoralSchemaWithAvro merge engine

yyy1000:
## Motivation
The existing `MergeHiveSchemaWithAvro` uses Hive `TypeInfo` as the structural source of truth, which is lossy for Iceberg tables (e.g. `TIMESTAMP(6)` loses microsecond precision, `FIXED` binary degrades to `BYTES`, UUID degrades to `STRING`).

This PR adds a new merge engine that uses Coral types (derived from Iceberg schema) as the source of truth while preserving partner Avro metadata — implementing the Iceberg-first semantics described in the design doc.

## Summary
- New `MergeCoralSchemaWithAvro` class with entry point `visit(StructType, Schema, String, String)`
- Coral/Iceberg is source of truth for: field existence, name, nullability, and primitive/logical types
- Partner Avro is source of truth for: defaults, docs, field props, aliases, enum/fixed/uuid materialization, union envelope shape, null placement
- Field matching: exact case-sensitive first, then unique case-insensitive (ambiguous = no match)
- Type mappings:
  - `TimestampType(6)` → `timestamp-micros`, `TimestampType(3)` → `timestamp-millis`
  - `BinaryType(n)` → Avro `fixed`, unbounded `BINARY` → `bytes`
  - `DecimalType(p,s)` → Avro `bytes` with decimal logical type
  - `DATE` → `int` with date logical type, `TIME` → `long` with time-micros
- Partner can be `null` for pure Coral-derived Avro schema generation
- Keeps `MergeHiveSchemaWithAvro` completely untouched

## How this will be used
In future PRs, `SchemaUtilities.getAvroSchemaForTable(CoralTable, strictMode)` will call this merge engine for Iceberg-backed tables:
```java
MergeCoralSchemaWithAvro.visit(
    (StructType) coralTable.getSchema(),  // Coral types from Iceberg
    partnerAvroSchema,                     // from getCasePreservedSchemaFromPropertyMaps (PR1)
    recordName,
    recordNamespace
);
```

## Test plan
- [x] 26 new tests in `MergeCoralSchemaWithAvroTests` covering:
  - Field name, nullability, and type sourced from Coral
  - Extra fields from Avro ignored, extra fields from Coral retained as optional
  - Doc strings, defaults, field props preserved from partner Avro
  - Arrays, maps, nested structs
  - Timestamp micros/millis/unspecified precision
  - Date, decimal, fixed binary, unbounded binary
  - BYTES→FIXED, STRING→ENUM, STRING→UUID promotion from partner
  - Null partner (pure Coral-derived schema)
  - Case-sensitive-first matching, ambiguous case-insensitive = no match
  - Partner null placement order preserved
  - Record name/namespace preserved from partner
- [x] Full `coral-schema` test suite passes (pre-existing JDO failures unrelated)

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
@@ -32,21 +32,18 @@ import static com.linkedin.coral.schema.avro.AvroSerdeUtils.*;
  * Merges a Coral schema (from Iceberg) with a partner Avro schema, using Iceberg-first semantics:
  *
  * <ul>
- *   <li>Coral/Iceberg is the source of truth for field existence, name, nullability, and type</li>
- *   <li>Partner Avro always contributes (when the attribute exists in the source): field name casing,
- *       default values, and field docs. The view schema spec is limited to {name, type, default, doc},
- *       so when the partner is derived from a view schema, only these attributes survive.</li>
- *   <li>When the partner is a full Avro table schema, it additionally contributes: field props, aliases,
- *       union envelope shape and null placement, and enum/fixed/uuid materialization. When the partner
- *       is derived from a view schema, these attributes are not present in the source and are dropped.</li>
+ *   <li>CoralDataType is the source of truth for field existence, name, nullability, and type.</li>
+ *   <li>Partner Avro contributes whatever metadata it carries for a matched field — defaults, docs,
+ *       field props, aliases, union envelope shape and null placement, enum/fixed/uuid materialization.
+ *       Nothing is fabricated: attributes the partner does not carry are absent from the output. When
+ *       the partner is derived from a view schema spec ({name, type, default, doc}), only those four
+ *       survive; when the partner is a full Avro table schema, the complete set is preserved.</li>
  *   <li>Fields only in the CoralDataType schema are included as optional; fields only in the partner
- *       Avro schema are dropped. Count mismatches do not fail, matching the behavior of
- *       {@link MergeHiveSchemaWithAvro}.</li>
+ *       are dropped. Count mismatches do not fail, matching {@link MergeHiveSchemaWithAvro}.</li>
  * </ul>
  *
- * Field matching: case-insensitive by field name. Output field names use the CoralDataType casing
- * (Iceberg-first). Iceberg's schema spec disallows sibling fields that differ only in case, so
- * case-insensitive matching against Iceberg-sourced schemas is unambiguous.
+ * Field matching is case-insensitive. Output field names use the CoralDataType casing. Iceberg's
+ * schema spec disallows sibling fields that differ only in case, so this is unambiguous.
  */
 class MergeCoralSchemaWithAvro {
 
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
@@ -1,0 +1,327 @@
+/**
+ * Copyright 2024-2026 LinkedIn Corporation. All rights reserved.
+ * Licensed under the BSD-2 Clause license.
+ * See LICENSE in the project root for license information.
+ */
+package com.linkedin.coral.schema.avro;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import javax.annotation.Nullable;
+
+import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
+
+import org.apache.avro.Schema;
+import org.apache.hadoop.hive.serde2.avro.AvroSerDe;
+
+import com.linkedin.coral.common.types.ArrayType;
+import com.linkedin.coral.common.types.BinaryType;
+import com.linkedin.coral.common.types.CoralDataType;
+import com.linkedin.coral.common.types.DecimalType;
+import com.linkedin.coral.common.types.MapType;
+import com.linkedin.coral.common.types.StructField;
+import com.linkedin.coral.common.types.StructType;
+import com.linkedin.coral.common.types.TimestampType;
+
+import static com.linkedin.coral.schema.avro.AvroSerdeUtils.*;
+
+
+/**
+ * Merges a Coral schema (from Iceberg) with a partner Avro schema, using Iceberg-first semantics:
+ *
+ * <ul>
+ *   <li>Coral/Iceberg is the source of truth for field existence, name, nullability, and type</li>
+ *   <li>Partner Avro provides metadata: defaults, docs, field props, aliases, enum/fixed/uuid materialization</li>
+ *   <li>If partner Avro says a field is a union, the union envelope shape and null placement are preserved from Avro</li>
+ *   <li>Fields only in Coral are included as optional; fields only in Avro are dropped</li>
+ * </ul>
+ *
+ * Field matching: exact case-sensitive match first, then unique case-insensitive match.
+ */
+class MergeCoralSchemaWithAvro {
+
+  private final AtomicInteger recordCounter = new AtomicInteger(0);
+
+  /**
+   * Entry point: merge a top-level Coral StructType with a partner Avro schema.
+   *
+   * @param structType top-level Coral schema from CoralTable.getSchema()
+   * @param partner partner Avro schema, or null if unavailable
+   * @param recordName Avro record name for the top-level record
+   * @param recordNamespace Avro record namespace for the top-level record
+   * @return merged Avro schema
+   */
+  static Schema visit(StructType structType, @Nullable Schema partner, String recordName, String recordNamespace) {
+    return new MergeCoralSchemaWithAvro().mergeTopLevelStruct(structType, partner, recordName, recordNamespace);
+  }
+
+  private Schema mergeTopLevelStruct(StructType structType, @Nullable Schema partner, String recordName,
+      String recordNamespace) {
+    Schema partnerRecord = extractPartnerRecord(partner);
+    List<Schema.Field> fields = new ArrayList<>();
+
+    for (StructField coralField : structType.getFields()) {
+      Schema.Field partnerField = findPartnerField(partnerRecord, coralField.getName());
+      Schema fieldSchema = mergeType(coralField.getType(), partnerField != null ? partnerField.schema() : null);
+      fields.add(mergeField(coralField.getName(), coralField.getType(), partnerField, fieldSchema));
+    }
+
+    Schema result;
+    if (partnerRecord != null) {
+      result = SchemaUtilities.copyRecord(partnerRecord, fields);
+    } else {
+      result = Schema.createRecord(recordName, null, recordNamespace, false);
+      result.setFields(fields);
+    }
+
+    // Top-level record is never wrapped in a nullable union — callers expect a record schema
+    return result;
+  }
+
+  private Schema mergeType(CoralDataType coralType, @Nullable Schema partner) {
+    switch (coralType.getKind()) {
+      case STRUCT:
+        return mergeStruct((StructType) coralType, partner);
+      case ARRAY:
+        return mergeArray((ArrayType) coralType, partner);
+      case MAP:
+        return mergeMap((MapType) coralType, partner);
+      default:
+        return mergeLeaf(coralType, partner);
+    }
+  }
+
+  private Schema mergeStruct(StructType structType, @Nullable Schema partner) {
+    Schema partnerRecord = extractPartnerRecord(partner);
+    List<Schema.Field> fields = new ArrayList<>();
+
+    for (StructField coralField : structType.getFields()) {
+      Schema.Field partnerField = findPartnerField(partnerRecord, coralField.getName());
+      Schema fieldSchema = mergeType(coralField.getType(), partnerField != null ? partnerField.schema() : null);
+      fields.add(mergeField(coralField.getName(), coralField.getType(), partnerField, fieldSchema));
+    }
+
+    Schema result;
+    if (partnerRecord != null) {
+      result = SchemaUtilities.copyRecord(partnerRecord, fields);
+    } else {
+      int recordNum = recordCounter.incrementAndGet();
+      result = Schema.createRecord("record" + recordNum, null, "namespace" + recordNum, false);
+      result.setFields(fields);
+    }
+
+    return applyCoralNullability(result, structType.isNullable(), partner);
+  }
+
+  private Schema mergeArray(ArrayType arrayType, @Nullable Schema partner) {
+    Schema partnerElement = null;
+    if (partner != null) {
+      Schema extracted = SchemaUtilities.extractIfOption(partner);
+      if (extracted.getType() == Schema.Type.ARRAY) {
+        partnerElement = extracted.getElementType();
+      }
+    }
+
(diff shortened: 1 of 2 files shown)


coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
```diff
@@ -0,0 +1,327 @@
...
+
+import static com.linkedin.coral.schema.avro.AvroSerdeUtils.*;
+
+
+/**
+ * Merges a Coral schema (from Iceberg) with a partner Avro schema, using Iceberg-first semantics:
+ *
+ * <ul>
+ *   <li>Coral/Iceberg is the source of truth for field existence, name, nullability, and type</li>
+ *   <li>Partner Avro provides metadata: defaults, docs, field props, aliases, enum/fixed/uuid materialization</li>
```
aastha25, line 36: since the view schema spec only contains name,type,default & doc, any additional attributes available in avro schema will be dropped.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
```diff
@@ -0,0 +1,327 @@
...
+ *   <li>If partner Avro says a field is a union, the union envelope shape and null placement are preserved from Avro</li>
+ *   <li>Fields only in Coral are included as optional; fields only in Avro are dropped</li>
```
aastha25, line 38: `Fields only in Coral` - ? please be more descriptive : `fields only in CoralDataType`. 

Additionally, if there's a mismatch of field count, what is the current behaviour in `MergeHiveSchemaWithAvro` ? I expect it to fail. Lets retain the existing behaviour in case of a mismatch.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
```diff
@@ -0,0 +1,327 @@
...
+
+    // 1. Exact case-sensitive match
+    Schema.Field exact = partnerRecord.getField(coralFieldName);
+    if (exact != null) {
+      return exact;
+    }
+
+    // 2. Unique case-insensitive match
+    Schema.Field caseInsensitiveMatch = null;
+    int matchCount = 0;
```
aastha25, line 316: unsure about the design decision here, are we allowing duplicate fully qualified field names with different casing? If yes, we need to compare by iceberg field Ids. Or, make a case insensitive comparison and prefer the casing of the field from iceberg schema. Internally, we've adopted the latter.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
```diff
@@ -0,0 +1,327 @@
...
+  /**
+   * Merge a single field: use Coral field name, apply partner metadata (docs, defaults, props).
+   */
+  private Schema.Field mergeField(String coralFieldName, CoralDataType coralFieldType, @Nullable Schema.Field partner,
+      Schema fieldSchema) {
+    if (partner == null) {
+      return AvroCompatibilityHelper.createSchemaField(SchemaUtilities.makeCompatibleName(coralFieldName), fieldSchema,
+          null, null);
+    }
+    Schema reordered = SchemaUtilities.reorderOptionIfRequired(fieldSchema, SchemaUtilities.defaultValue(partner));
```
aastha25, line 265: could you also explain why could we potentially need a 'reorder' here? whats the usecase? lets add the reasoning as an inline comment too

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
yyy1000: Current behavior:
```
  ▎ "4. Retains fields found only in the Hive schema; Ignores  
  ▎ fields found only in the Avro schema
  ▎ 5. Fields found only in Hive schema are represented as     
  ▎ optional fields in the resultant Avro schema"        
```      
  
  MergeHiveSchemaWithAvro does not fail on mismatch — it keeps 
  source-of-truth-only fields as optional and drops
  partner-only fields. That's exactly what our PR does. So we  
  already retain existing behavior.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
yyy1000: Good call — switched to pure case-insensitive (first match wins), same as `MergeHiveSchemaWithAvro#findCaseInsensitive`. Iceberg disallows case-differing siblings so there's no ambiguity. Output names already use Iceberg casing since `mergeField` takes `coralField.getName()`. Replaced the two casing tests with a single `shouldMatchPartnerFieldCaseInsensitively`.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
yyy1000: Added a comment. Avro requires the default to match the first union branch — so if Coral gives `[null, int]` but the partner's default is `1`, we swap to `[int, null]` before copying the default over.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
yyy1000: Good point — tightened the Javadoc to make it explicit that nothing is fabricated: the partner contributes only what it carries, so view-schema partners naturally drop to just `{name, type, default, doc}` and full Avro table partners carry everything through. No code change needed since the engine only preserves.

coral-schema/src/test/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvroTests.java
```diff
@@ -0,0 +1,384 @@
...
+  @Test
+  public void shouldUseFieldNamesFromCoral() {
+    StructType coral = struct(field("fA", intType(true)), field("fB", struct(field("gA", intType(true)))));
+    Schema avro = avroStruct("r1", optionalField("fa", Schema.Type.INT),
+        optionalField("fb", avroStruct("r2", optionalField("ga", Schema.Type.INT))));
+
+    Schema result = merge(coral, avro);
+    // Coral field names are used (via partner copyField which preserves partner names,
+    // but matching is case-insensitive so partner field "fa" matches Coral "fA")
+    assertNotNull(result.getField("fa")); // partner name preserved via copyField
```
simbadzina: Shouldn't the casing here be `fA`, from the Coral type?

simbadzina: [ * Field matching is case-insensitive. Output field names use the CoralDataType casing. Iceberg's](https://github.com/linkedin/coral/pull/600/changes#diff-a9e3efb69e74ca5841eccab244f8d649aa2583fb1c67ad81f5fe0b6c84daf10cR45)

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
```diff
@@ -0,0 +1,321 @@
...
+      Schema fieldSchema) {
+    if (partner == null) {
+      return AvroCompatibilityHelper.createSchemaField(SchemaUtilities.makeCompatibleName(coralFieldName), fieldSchema,
+          null, null);
+    }
+    // Avro requires the default value to match the first type in the option, reorder option if required.
+    // e.g. fieldSchema is [null, int] and the partner's default value is 1 → reorder to [int, null] so
+    // the default is compatible with the first branch.
+    Schema reordered = SchemaUtilities.reorderOptionIfRequired(fieldSchema, SchemaUtilities.defaultValue(partner));
+    return SchemaUtilities.copyField(partner, reordered);
```
aastha25: very surprised to learn that we're already returning field_props. but we'll see in regression testing if there's any issue.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
```diff
@@ -0,0 +1,321 @@
...
+      case CHAR:
+      case VARCHAR:
+        return Schema.create(Schema.Type.STRING);
+      case BINARY:
+        return binaryToAvro((BinaryType) coralType);
+      case NULL:
+        return Schema.create(Schema.Type.NULL);
+      case DATE: {
+        Schema dateSchema = Schema.create(Schema.Type.INT);
+        dateSchema.addProp(AvroSerDe.AVRO_PROP_LOGICAL_TYPE, AvroSerDe.DATE_TYPE_NAME);
```
aastha25: can you check if these need to be imported from `org.apache.hadoop.hive.serde2.avro.AvroSerDe` or is there some avro library which carries this components? I was expecting that merging (iceberg+avro) would have net 0 dependency on hive modules , for example hive-serde that you have imported here

coral-schema/src/test/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvroTests.java
aastha25: +1 . 
all the other tests has 'fA' assertion but there the avro schema has that casing as well

aastha25 (commented):
overall LGTM, some minor (future) iterations may be required in `MergeCoralSchemaWithAvro` as we run regression testing with production views. Please create todo tickets for[ comment#1](https://github.com/linkedin/coral/pull/600/changes#r3148865482).

[this](https://github.com/linkedin/coral/pull/600/changes#r3147434432) is the only open issue. 
rest, LGTM. thanks for working on this!

coral-schema/src/test/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvroTests.java
simbadzina: I got confused. IIUC now both `fA` and `fa` should be present. `fa` being an alias for `fA`. If so, can you add a  assertion that `fA` is also there.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
yyy1000: I checked and find it can be imported in avro directly. Alrady made the change.

coral-schema/src/test/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvroTests.java
yyy1000: Thanks for catching this — the implementation didn't match the design claim. I already updated the code and the assertion.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
```diff
@@ -259,19 +252,29 @@ private Schema checkCompatibilityAndPromote(Schema coralSchema, @Nullable Schema
...
     // the default is compatible with the first branch.
     Schema reordered = SchemaUtilities.reorderOptionIfRequired(fieldSchema, SchemaUtilities.defaultValue(partner));
-    return SchemaUtilities.copyField(partner, reordered);
+    Schema.Field merged = AvroCompatibilityHelper.createSchemaField(safeCoralName, reordered, partner.doc(),
+        SchemaUtilities.defaultValue(partner), partner.order());
+    SchemaUtilities.replicateFieldProps(partner, merged);
+    partner.aliases().forEach(merged::addAlias);
+    if (!partner.name().equals(safeCoralName)) {
+      merged.addAlias(partner.name());
+    }
```
aastha25: is `alias` going to a be a new key-value pair in the avro schema generated? 
let's stick to keeping the avro schema same as before where every field only contains name, schema, doc & default. 
the read/write APIs are responsible for performing case insensntive checks.

coral-schema/src/test/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvroTests.java
```diff
@@ -36,10 +36,16 @@ public void shouldUseFieldNamesFromCoral() {
...
 
     Schema result = merge(coral, avro);
-    // Coral field names are used (via partner copyField which preserves partner names,
-    // but matching is case-insensitive so partner field "fa" matches Coral "fA")
-    assertNotNull(result.getField("fa")); // partner name preserved via copyField
-    assertNotNull(result.getField("fb"));
+    // Iceberg-first: canonical field name is the Coral casing; the partner's casing is preserved
+    // as an alias so readers using the old name still resolve.
+    assertNotNull(result.getField("fA"));
+    assertTrue(result.getField("fA").aliases().contains("fa"));
```
aastha25: please change assertion to 
`assertNotNull(result.getField("fA"));` 
&
`assertNull(result.getField("fa"));`

and lets remove the aliases from the generated schema. we can introduce it later if there's a broader need for it.

yyy1000: done

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
yyy1000: ok, I removed it.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
```diff
@@ -0,0 +1,312 @@
...
+  /**
+   * Entry point: merge a top-level Coral StructType with a partner Avro schema.
+   *
+   * @param structType top-level Coral schema from CoralTable.getSchema()
+   * @param partner partner Avro schema, or null if unavailable
+   * @param recordName Avro record name for the top-level record
+   * @param recordNamespace Avro record namespace for the top-level record
+   * @return merged Avro schema
+   */
+  static Schema visit(StructType structType, @Nullable Schema partner, String recordName, String recordNamespace) {
```
simbadzina: Can you change this method name to `merge` or something else other than `visit`?

`MergeHiveSchemaWithAvro.visit` uses a visitor pattern, while `MergeCoralSchemaWithAvro` does not. I was wondering for a while whether we should only use the same pattern but seems it is not necessary.

Renaming will help clarify that is it an explicit choice to not use the visitor pattern here.

yyy1000: cool, I changed the name.

aastha25: hmm good observation. the standard is to use the visitor pattern, it is more robust. 
but lets revisit that now in a follow up PR to unblock the iceberg-first schema inference work.

aastha25 (approved):

PR #602, opened 2026-04-22
Fix Coral parse failure for non-reserved keywords used as table aliases

c-h-afzal:
When Hive stores a view, its Cost-Based Optimizer (CBO) automatically
    appends the table name as an unquoted alias to every table reference:

      -- You write:
      CREATE VIEW v AS SELECT * FROM db.collection WHERE x > 0
      -- Hive stores:
      SELECT * FROM db.collection collection WHERE x > 0

When the table name is a Hive non-reserved keyword (e.g., collection, comment, role — listed under the nonReserved rule in the Hive grammar), Coral's parser fails because the tableSource grammar rule only accepted Identifier lexer tokens as aliases. ANTLR tokenizes keywords like "collection" as KW_COLLECTION, not Identifier, so they could not match.

The fix introduces a tableAliasCandidate parser rule that accepts Identifier | nonReserved keywords, and uses it for implicit (non-AS) table aliases. The nonReserved set includes keywords like collection, comment, and role that are safe to use as identifiers.

Using the full identifier rule was not safe because Coral sets useSQL11ReservedKeywordsForIdentifier() to true, which makes identifier also match structural SQL keywords (UNION, INNER, GROUP, LEFT, etc.).
This caused the parser to consume these keywords as table aliases instead of recognizing them as SQL operators. For example:

      SELECT * FROM test.bar1 UNION SELECT * FROM test.bar2
                              ^^^^^ consumed as alias -> parse failure

      SELECT * FROM test.bar1 INNER JOIN test.bar2 ON ...
                              ^^^^^ consumed as alias -> parse failure

The explicit KW_AS alias branch still uses the full identifier rule since AS provides unambiguous disambiguation:
      SELECT * FROM db.group AS group WHERE ...  -- works

- New unit tests added
- Tested the change with LinkedIn's regression DAGs against the branch v2.3.9 with zero regressions.

### DIFF
--- a/coral-hive/src/main/antlr/imports/FromClauseParser.g
+++ b/coral-hive/src/main/antlr/imports/FromClauseParser.g
@@ -186,10 +186,16 @@ tableSource
     : tabname=tableName
     ((tableProperties) => props=tableProperties)?
     ((tableSample) => ts=tableSample)?
-    ((KW_AS) => (KW_AS alias=Identifier)
+    ((KW_AS) => (KW_AS asAlias=identifier)
     |
-    (Identifier) => (alias=Identifier))?
-    -> ^(TOK_TABREF $tabname $props? $ts? $alias?)
+    (tableAliasCandidate) => (implicitAlias=tableAliasCandidate))?
+    -> ^(TOK_TABREF $tabname $props? $ts? $asAlias? $implicitAlias?)
+    ;
+
+tableAliasCandidate
+    :
+    Identifier
+    | nonReserved -> Identifier[$nonReserved.text]
     ;
 
 tableName
--- a/coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilderTest.java
+++ b/coral-hive/src/test/java/com/linkedin/coral/hive/hive2rel/parsetree/ParseTreeBuilderTest.java
@@ -248,4 +248,61 @@ public class ParseTreeBuilderTest {
     // Validate if the translation is successful
     assertEquals(sqlNode.toString().replaceAll("\\r?\\n", " "), table.getViewExpandedText());
   }
+
+  /**
+   * Validates that non-reserved keywords (e.g., collection, comment, role — listed under the
+   * {@code nonReserved} rule in the Hive grammar) can be used as unquoted table aliases in FROM
+   * clauses. This is needed because Hive's CBO automatically adds the table name as an unquoted
+   * alias when persisting view definitions.
+   * See BDP-70589.
+   */
+  @DataProvider(name = "nonReservedKeywordImplicitAlias")
+  public Object[][] getNonReservedKeywordImplicitAliasSql() {
+    return new Object[][] { { "SELECT collection.a FROM test.tableOne collection WHERE collection.a > 0", "SELECT `collection`.`a`\nFROM `test`.`tableOne` AS `collection`\nWHERE `collection`.`a` > 0" }, { "SELECT comment.a FROM test.tableOne comment WHERE comment.a > 0", "SELECT `comment`.`a`\nFROM `test`.`tableOne` AS `comment`\nWHERE `comment`.`a` > 0" }, { "SELECT role.a FROM test.tableOne role WHERE role.a > 0", "SELECT `role`.`a`\nFROM `test`.`tableOne` AS `role`\nWHERE `role`.`a` > 0" }, { "SELECT data.a FROM test.tableOne data WHERE data.a > 0", "SELECT `data`.`a`\nFROM `test`.`tableOne` AS `data`\nWHERE `data`.`a` > 0" }, { "SELECT schema.a FROM test.tableOne schema WHERE schema.a > 0", "SELECT `schema`.`a`\nFROM `test`.`tableOne` AS `schema`\nWHERE `schema`.`a` > 0" }, { "SELECT view.a FROM test.tableOne view WHERE view.a > 0", "SELECT `view`.`a`\nFROM `test`.`tableOne` AS `view`\nWHERE `view`.`a` > 0" }, { "SELECT index.a FROM test.tableOne index WHERE index.a > 0", "SELECT `index`.`a`\nFROM `test`.`tableOne` AS `index`\nWHERE `index`.`a` > 0" }, { "SELECT enable.a FROM test.tableOne enable WHERE enable.a > 0", "SELECT `enable`.`a`\nFROM `test`.`tableOne` AS `enable`\nWHERE `enable`.`a` > 0" }, };
+  }
+
+  @Test(dataProvider = "nonReservedKeywordImplicitAlias")
+  public void testNonReservedKeywordAsImplicitAlias(String input, String expected) {
+    SqlNode sqlNode = convert(input);
+    assertNotNull(sqlNode, "Failed to parse SQL with nonReserved keyword as implicit alias: " + input);
+    assertEquals(sqlNode.toString(), expected);
+  }
+
+  @DataProvider(name = "nonReservedKeywordExplicitAlias")
+  public Object[][] getNonReservedKeywordExplicitAliasSql() {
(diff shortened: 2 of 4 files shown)


aastha25 (commented):
LGTM, please consider adding a UT in coral-spark as well where we create a view V1 as `SELECT * FROM db.collection WHERE ...` and then checking if querying the view is successful for an e2e test case.

The initial error reported from reading this type of view might be due to a parsing error but there could be a follow up issue from spark side as well. In the coral translation pipeline triggered while reading the view, the parsing is now successful and the coral IR is created, but is the coral IR convertible to spark SQL and is that spark SQL interpretable by spark?

wmoustafa:
Could you clarify why this commit did not address this use case? https://github.com/linkedin/coral/commit/a83b648216f7429ca5b67c264a1c44ba4ae77be8

Also, could you please check the most recent Hive and Spark upstream behavior in this situation?

One risk this presents is the ambiguity in queries like:
```
SELECT * FROM db.mytable group BY col
```
which would be a regression.

aastha25:
> Could you clarify why this commit did not address this use case? [a83b648](https://github.com/linkedin/coral/commit/a83b648216f7429ca5b67c264a1c44ba4ae77be8)
> 
> Also, could you please check the most recent Hive and Spark upstream behavior in this situation?
> 
> One risk this presents is the ambiguity in queries like:
> 
> ```
> SELECT * FROM db.mytable group BY col
> ```
> 
> which would be a regression.

my expectation is that in this case it would not get parsed as an identifier. some existing unit test would have failed. @c-h-afzal could you please try this out and confirm?

c-h-afzal:
Hey @wmoustafa  @aastha25  - appreciate your help on the review and feedback. I have added an extra test case and here's the explanation for the questions raised. 

**1. Why commit a83b648 didn't address this use case**

Commit a83b648 solves a different problem: Spark-created views that contain unquoted reserved keywords as column names (e.g., SELECT group FROM ...). It detects Spark-created views by checking the spark.sql.create.version table property and then sets hive.support.sql11.reserved.keywords=false on the parser, which enables the sql11ReservedKeywordsUsedAsIdentifier rule for column resolution.

Our issue is about reserved keywords used as table aliases in the tableSource grammar rule. The root cause is that tableSource used the lexer rule Identifier (uppercase — accepts only raw identifier tokens) instead of the parser rule identifier (lowercase — accepts Identifier + nonReserved keywords). This meant table aliases like collection, comment, or role — which are nonReserved keywords in the grammar — were rejected. Commit a83b648's sql11ReservedKeywordsUsedAsIdentifier mechanism doesn't help here because it operates at the identifier parser rule level, and the old code never reached that rule for table aliases.

**2. Upstream Hive and Spark behavior**

Coral's FromClauseParser.g was forked from Hive 1.1.0, which used alias=Identifier (uppercase). Upstream Hive changed this:
  - Hive 2.0.0: Still alias=Identifier (uppercase), but added syntactic predicates for KW_AS
  - Hive 2.3.0: ~~Changed to alias=identifier (lowercase) — exactly what this PR does~~  [update:We don't use identifier anymore; we use tableAliasCandidate which is a restricted subset (Identifier | nonReserved)]
  - Hive 3.1.0, 4.0.0, master: All use alias=identifier (lowercase)

~~This PR aligns Coral's grammar with upstream Hive 2.3.0+.~~ [update: We intentionally diverge from upstream because Coral's useSQL11ReservedKeywordsForIdentifier() returns true unconditionally, making the full identifier rule unsafe.]

**3. Ambiguity risk with SELECT * FROM db.mytable group BY col**

This is not a regression. The identifier parser rule accepts three categories:
  1. Identifier (raw identifier tokens)
  2. nonReserved keywords (e.g., KW_COLLECTION, KW_COMMENT, KW_ROLE)
  3. sql11ReservedKeywordsUsedAsIdentifier (e.g., KW_GROUP, KW_ORDER, KW_BY) — but only when useSQL11ReservedKeywordsForIdentifier() returns true

KW_GROUP is in category 3, not category 2. ~~The function useSQL11ReservedKeywordsForIdentifier() defaults to false in CoralParseDriver and is only set to true for Spark-created views (via commit a83b648).~~ [update:This was wrong. It returns true unconditionally in HiveParser.g. That's
  the whole reason we couldn't use identifier and had to create tableAliasCandidate.] ~~So under normal parsing, group in SELECT * FROM db.mytable group BY col will not be consumed as a table alias~~[update: The conclusion is still correct, but the reasoning is wrong. It's not because the flag defaults to false — it's because
  tableAliasCandidate excludes KW_GROUP entirely (since KW_GROUP is in sql11ReservedKeywordsUsedAsIdentifier, not in nonReserved)] — it will correctly be parsed as the GROUP BY clause.

I've added a test (testGroupByNotMistakenForAlias) that verifies SELECT a FROM test.tableOne group BY a correctly parses group as the GROUP BY keyword, not as a table alias.

wmoustafa:
@c-h-afzal My expectation from the linked commit is if it solves table names, then it should solve aliases to by induction? In the description of this PR you stated that aliases are added automatically; then keyword aliases must have been added in the commit I linked too.

c-h-afzal:
> @c-h-afzal My expectation from the linked commit is if it solves table names, then it should solve aliases to by induction? In the description of this PR you stated that aliases are added automatically; then keyword aliases must have been added in the commit I linked too.

@wmoustafa  Gotcha! — you concern is valid and also I made some updates sicne your comment. Let me explain why a83b648 doesn't fix for aliases: 

 First, a83b648 didn't actually solve table names for keywords like collection, comment, or role. Those are nonReserved in the grammar, so the tableName rule (which uses the identifier parser rule) already accepted them before a83b648. Table names like db.collection never had a problem.

  What a83b648 solved is a different issue: Spark-created views with sql11ReservedKeywords (like group, timestamp) used as column names. It sets hive.support.sql11.reserved.keywords=false for Spark views, enabling the
  sql11ReservedKeywordsUsedAsIdentifier alternative in the identifier parser rule.

  The alias issue existed because tableSource used the Identifier lexer token (uppercase) for implicit aliases, not the identifier parser rule:

  (Identifier) => (alias=Identifier)    // ← lexer token, not parser rule

  The lexer sees collection and emits KW_COLLECTION, not Identifier. No parser-level flag can change that — a83b648 has zero effect here.

  The natural fix is to switch to the identifier parser rule (which is what upstream Hive did in 2.3.0+), but in Coral useSQL11ReservedKeywordsForIdentifier() returns true unconditionally for Spark views, so identifier also matches structural
  SQL keywords (UNION, INNER, GROUP, LEFT, etc.). This breaks queries like:

  SELECT * FROM test.t1 UNION SELECT * FROM test.t2
  &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;^^^^^ consumed as alias → parse failure

  The fix introduces tableAliasCandidate which accepts Identifier | nonReserved only — excluding sql11ReservedKeywordsUsedAsIdentifier. This covers the CBO alias case (collection, comment, role are all nonReserved) while keeping structural
  keywords safe. The explicit AS branch still uses the full identifier rule since AS provides unambiguous disambiguation.

  Tests cover all three scenarios: nonReserved implicit aliases, sql11 explicit AS aliases, and sql11 keywords correctly parsed as SQL operators (GROUP BY, INNER JOIN, LEFT JOIN, UNION, etc.).

aastha25:
The intention is to support 'non-Reserved Keyword' and thats the right terminology but the PR description, and UTs have inconsistent language.
instance#1 - `Validates that reserved keywords can be used as unquoted table aliases in FROM clauses.` [here in UT](https://github.com/linkedin/coral/pull/602/changes#diff-0c7d509dd5ee66adb2f850eabd219a062d0cdf87a36b49d95ec5bed495fee741R253). this should be changed to "non-reserved keyword" ?
instance#2 - `When the table name is a Hive keyword (e.g., collection, comment, role), ...` lets be explicit about this list being non-reserved keywords and what are the said keywords. otherwise it makes it hard to follow and look up in retro when the context is lost. 
 instance#3 - the pr title. 

code changes & test coverage - LGTM

aastha25 (approved):

PR #589, opened 2026-03-30
Refactor partner-Avro parsing into reusable property-map helper

yyy1000:
## Motivation
Today, partner-Avro parsing in `SchemaUtilities` is tightly coupled to the Hive `Table` object — `getCasePreservedSchemaFromTblProperties(Table)` and `readSchemaFromSchemaLiteral(Table)` both require a Hive metastore `Table` to read `avro.schema.literal` and `dali.row.schema`.

This is a blocker for the Iceberg-first Avro rollout, where `CoralTable`-backed flows (e.g. Iceberg tables) need to read partner Avro from property maps (`CoralTable.properties()`) without having a Hive `Table` available. Without this refactor, every new schema path would need to duplicate the parsing logic.

## Summary
- Extract schema-reading logic from `getCasePreservedSchemaFromTblProperties` and `readSchemaFromSchemaLiteral` into a new `getCasePreservedSchemaFromPropertyMaps(Map<String, String>, Map<String, String>, String)` method
- Enables `CoralTable`-backed flows (e.g. Iceberg tables) to reuse partner-Avro parsing without requiring a Hive metastore `Table` object
- Refactor existing `getCasePreservedSchemaFromTblProperties(Table)` to delegate to the new method; remove now-unused `readSchemaFromSchemaLiteral`

## How the new method will be used

**Hive tables (unchanged):** `getCasePreservedSchemaFromTblProperties(Table)` is still called at all existing call sites. It now delegates internally to `getCasePreservedSchemaFromPropertyMaps`, extracting `table.getParameters()` and `table.getSd().getSerdeInfo().getParameters()` as the two maps.

**Iceberg/OpenHouse tables (future PRs):** When `SchemaUtilities.getAvroSchemaForTable(CoralTable, strictMode)` is added, it will call `getCasePreservedSchemaFromPropertyMaps` directly:

```java
Schema partnerAvro = getCasePreservedSchemaFromPropertyMaps(
    coralTable.properties(),  // table-level properties from CoralTable
    null,                     // no serde properties — Iceberg doesn't have a Hive SerDe layer
    coralTable.name()         // e.g. "db@tableName" for logging
);
```

The `serdeProperties` parameter is `null` because Iceberg/OpenHouse tables don't go through a Hive SerDe — they store `avro.schema.literal` (if at all) in their table properties directly. The method handles `null` serde properties gracefully by skipping that lookup.

## Test plan
[...]

PR #604, opened 2026-04-27
[Coral-Schema] Add CoralCatalog base-table and table-scan paths for Avro schema generation

yyy1000:
## Motivation
PR #600 added the `MergeCoralSchemaWithAvro` engine but did not wire it into any call site. As a result, base tables and table scans inside views still resolve through the Hive-only path, even when the table is Iceberg-backed.

This PR adds the Coral-side `CoralCatalog` plumbing so base tables and table scans can resolve through the new merge engine. Hive- and Iceberg-backed base tables share a single dispatch path; Hive behavior is preserved unchanged for callers that don't opt in.

## Summary
- `SchemaUtilities.getAvroSchemaForTable(CoralTable, strictMode)` — new overload. `HiveTable` delegates to the existing `getAvroSchemaForTable(Table, strictMode)`. Other CoralTables (Iceberg-backed) parse partner Avro from `CoralTable.properties()` via `getCasePreservedSchemaFromPropertyMaps` (PR #589) and merge with the Coral schema using `MergeCoralSchemaWithAvro.merge`. Strict mode throws `SchemaNotFoundException` when partner Avro is absent; non-strict mode emits a pure Coral-derived schema.
- `ViewToAvroSchemaConverter.create(CoralCatalog, HiveMetastoreClient)` — new factory + matching private constructor + optional `CoralCatalog` field. The existing `create(HiveMetastoreClient)` factory and constructor are unchanged. `inferAvroSchema()` dispatches base-table resolution through the `CoralCatalog` when one is provided; views fall through to the existing path because view conversion still requires `HiveToRelConverter` on a Hive metastore. `coralCatalog` is also threaded into the inner `RelToAvroSchemaConverter`.
- `RelToAvroSchemaConverter` — new optional `CoralCatalog` field + `(CoralCatalog, HiveMetastoreClient)` constructor. The existing `(HiveMetastoreClient)` constructor delegates to it with `null`. `SchemaRelShuttle.getTableScanSchema()` resolves through `coralCatalog.getTable()` first when set, falling back to `hiveMetastoreClient.getTable()` if the CoralCatalog does not surface the table.

## Behavior preservation
Every existing constructor/factory still exists. The new constructors delegate to the legacy ones with a `null coralCatalog`, so the legacy code path is byte-for-byte identical. Callers that don't opt in see no behavior change.

## Test plan
- [x] New `SchemaUtilitiesTests` cases for the Iceberg path:
  - `testGetAvroSchemaForCoralTableIcebergWithPartner` — partner-provided record name preserved, fields populated.
  - `testGetAvroSchemaForCoralTableIcebergNoPartnerNonStrict` — pure Coral-derived schema with name/namespace from the CoralTable name.
  - `testGetAvroSchemaForCoralTableIcebergNoPartnerStrictThrows` — `SchemaNotFoundException` raised.
  - `testGetAvroSchemaForCoralTableUnqualifiedNameYieldsEmptyNamespace` — unqualified name handling.
- [x] The Hive-delegation branch and the `ViewToAvroSchemaConverter` / `RelToAvroSchemaConverter` dispatches are covered transitively by existing tests on the `HiveMetastoreClient`-only constructors — the new constructors and factories are additive, and `null coralCatalog` reproduces prior behavior.
- [x] Full `coral-schema:test` suite passes; `spotlessJavaCheck` clean.

## Follow-up
Complex multi-branch unions in partner Avro are not yet handled by `MergeCoralSchemaWithAvro` (only `[null, T]` shapes are). A follow-up PR will extend the merge engine for union/fixed/UUID/Avro-only edge cases. Pure Iceberg base tables rarely carry multi-branch unions in their partner Avro, so this PR is safe to ship before that work; production regression testing on this path will inform the exact scope of the follow-up.

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
@@ -13,6 +13,7 @@ import java.util.Map;
 import java.util.Queue;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 import org.apache.avro.Schema;
 import org.apache.avro.SchemaBuilder;
@@ -59,6 +60,8 @@ import org.slf4j.LoggerFactory;
 import com.linkedin.coral.com.google.common.base.Preconditions;
 import com.linkedin.coral.common.HiveMetastoreClient;
 import com.linkedin.coral.common.HiveUncollect;
+import com.linkedin.coral.common.catalog.CoralCatalog;
+import com.linkedin.coral.common.catalog.CoralTable;
 import com.linkedin.coral.hive.hive2rel.functions.OrdinalReturnTypeInferenceV2;
 
 
@@ -104,11 +107,35 @@ import com.linkedin.coral.hive.hive2rel.functions.OrdinalReturnTypeInferenceV2;
  *
  */
 public class RelToAvroSchemaConverter {
+  @Nullable
   private final HiveMetastoreClient hiveMetastoreClient;
+  @Nullable
+  private final CoralCatalog coralCatalog;
   private static final Logger LOG = LoggerFactory.getLogger(RelToAvroSchemaConverter.class);
 
+  /**
+   * Legacy constructor wired against {@link HiveMetastoreClient}. Prefer
+   * {@link #RelToAvroSchemaConverter(CoralCatalog)} for new callers.
+   *
+   * @deprecated use {@link #RelToAvroSchemaConverter(CoralCatalog)} instead.
+   */
+  @Deprecated
   public RelToAvroSchemaConverter(HiveMetastoreClient hiveMetastoreClient) {
     this.hiveMetastoreClient = hiveMetastoreClient;
+    this.coralCatalog = null;
+  }
+
+  /**
+   * CoralCatalog-backed constructor. Base-table resolution inside
+   * {@link SchemaRelShuttle#getTableScanSchema} is delegated to the CoralCatalog (and
+   * {@link SchemaUtilities#getAvroSchemaForTable(CoralTable, boolean)}), so Hive- and
+   * Iceberg-backed tables share a single resolution path. This constructor and the legacy
+   * {@link #RelToAvroSchemaConverter(HiveMetastoreClient)} are mutually exclusive — pick one
+   * abstraction; instances do not mix.
+   */
+  public RelToAvroSchemaConverter(CoralCatalog coralCatalog) {
+    this.coralCatalog = coralCatalog;
+    this.hiveMetastoreClient = null;
   }
 
   /**
@@ -125,7 +152,7 @@ public class RelToAvroSchemaConverter {
     Preconditions.checkNotNull(relNode, "RelNode to convert cannot be null");
 
     Map<RelNode, Schema> schemaMap = new HashMap<>();
-    relNode.accept(new SchemaRelShuttle(hiveMetastoreClient, schemaMap, strictMode, forceLowercase));
+    relNode.accept(new SchemaRelShuttle(coralCatalog, hiveMetastoreClient, schemaMap, strictMode, forceLowercase));
     Schema viewSchema = schemaMap.get(relNode);
 
     return viewSchema;
@@ -171,10 +198,17 @@ public class RelToAvroSchemaConverter {
     private final boolean strictMode;
     private final boolean forceLowercase;
 
+    // Exactly one of these is non-null per instance, mirroring the public constructor split on
+    // RelToAvroSchemaConverter. The two paths never mix: a CoralCatalog-backed shuttle never
+    // resolves a table through HiveMetastoreClient, and vice versa.
+    @Nullable
     private final HiveMetastoreClient hiveMetastoreClient;
+    @Nullable
+    private final CoralCatalog coralCatalog;
 
-    public SchemaRelShuttle(HiveMetastoreClient hiveMetastoreClient, Map<RelNode, Schema> schemaMap, boolean strictMode,
-        boolean forceLowercase) {
+    public SchemaRelShuttle(@Nullable CoralCatalog coralCatalog, @Nullable HiveMetastoreClient hiveMetastoreClient,
+        Map<RelNode, Schema> schemaMap, boolean strictMode, boolean forceLowercase) {
+      this.coralCatalog = coralCatalog;
       this.hiveMetastoreClient = hiveMetastoreClient;
       this.schemaMap = schemaMap;
       this.strictMode = strictMode;
@@ -361,25 +395,39 @@ public class RelToAvroSchemaConverter {
     }
 
     /**
-     * This method retrieves avro schema for tableScan
+     * Retrieves the avro schema for a tableScan.
      *
-     * @param tableScan
-     * @return avro schema for tableScan. The schema includes partition columns if table is partitioned
-     * @throws RuntimeException if cannot find table in Hive metastore
-     * @throws RuntimeException if cannot determine avro schema for tableScan
+     * Resolution stays inside whichever abstraction the converter was constructed with:
+     * <ul>
+     *   <li>CoralCatalog-backed instances resolve via {@link CoralCatalog#getTable} and
+     *       {@link SchemaUtilities#getAvroSchemaForTable(CoralTable, boolean)}.</li>
+     *   <li>HiveMetastoreClient-backed instances resolve via {@link HiveMetastoreClient#getTable}
+     *       and {@link SchemaUtilities#getAvroSchemaForTable(Table, boolean)}.</li>
+     * </ul>
+     * The two paths do not fall back to each other — if the configured abstraction does not
+     * surface the table, this method throws.
+     *
+     * @return avro schema for tableScan. The schema includes partition columns if the table is partitioned.
+     * @throws RuntimeException if the configured catalog/metastore cannot find the table or produce its schema.
      */
     private Schema getTableScanSchema(TableScan tableScan) {
       List<String> qualifiedName = tableScan.getTable().getQualifiedName();
       String dbName = qualifiedName.get(1);
       String tableName = qualifiedName.get(2);
+
+      if (coralCatalog != null) {
+        CoralTable coralTable = coralCatalog.getTable(dbName, tableName);
+        if (coralTable == null) {
+          throw new RuntimeException("Cannot find table " + dbName + "." + tableName + " in CoralCatalog");
+        }
+        return SchemaUtilities.getAvroSchemaForTable(coralTable, strictMode);
+      }
+
       Table baseTable = hiveMetastoreClient.getTable(dbName, tableName);
       if (baseTable == null) {
         throw new RuntimeException("Cannot find table " + dbName + "." + tableName + " in Hive metastore");
       }
-
-      Schema tableSchema = SchemaUtilities.getAvroSchemaForTable(baseTable, strictMode);
-
-      return tableSchema;
+      return SchemaUtilities.getAvroSchemaForTable(baseTable, strictMode);
     }
   }
 
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
@@ -42,6 +42,9 @@ import org.slf4j.LoggerFactory;
 
 import com.linkedin.coral.com.google.common.base.Preconditions;
 import com.linkedin.coral.com.google.common.base.Strings;
+import com.linkedin.coral.common.catalog.CoralTable;
+import com.linkedin.coral.common.catalog.HiveTable;
+import com.linkedin.coral.common.types.StructType;
 import com.linkedin.coral.schema.avro.exceptions.SchemaNotFoundException;
 
 import static com.linkedin.coral.schema.avro.AvroSerdeUtils.*;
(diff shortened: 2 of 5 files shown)


coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
```diff
@@ -55,6 +73,20 @@ public static ViewToAvroSchemaConverter create(HiveMetastoreClient hiveMetastore
...
 
+  /**
+   * Creates a converter that uses {@code coralCatalog} for base-table resolution and falls back
+   * to {@code hiveMetastoreClient} for views and for tables not visible through the CoralCatalog.
+   *
+   * @param coralCatalog catalog used for base-table resolution; must not be null
+   * @param hiveMetastoreClient client used for view conversion and as a fallback; must not be null
+   * @return a new converter wired to both
+   */
+  public static ViewToAvroSchemaConverter create(CoralCatalog coralCatalog, HiveMetastoreClient hiveMetastoreClient) {
```
aastha25: not aligned with this user facing method. from a UX perspective, if a user is looking to generate a view's avro schema , they will either pass  `CoralCatalog` or `HiveMetastoreClient` depending on whether they are on new or old variant.
we should - 
(1) expose a public static - `public static ViewToAvroSchemaConverter create(CoralCatalog coralCatalog) {..}` method. 
(2) mark `public static ViewToAvroSchemaConverter create(HiveMetastoreClient hiveMetastoreClient) {..)` deprecated.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
```diff
@@ -372,6 +392,16 @@ private Schema getTableScanSchema(TableScan tableScan) {
...
+      // Hive- and Iceberg-backed base tables share the same downstream merge engine.
+      if (coralCatalog != null) {
+        CoralTable coralTable = coralCatalog.getTable(dbName, tableName);
+        if (coralTable != null) {
+          return SchemaUtilities.getAvroSchemaForTable(coralTable, strictMode);
+        }
+      }
+
       Table baseTable = hiveMetastoreClient.getTable(dbName, tableName);
       if (baseTable == null) {
```
aastha25, line 427: this codepath makes we wonder that we are potentially allowing users to create a `CoralRelNode` using the old abstractions - `HiveMetastoreClient` and then create a view schema by merging `coralRelNode` generated using `HiveMetastoreClient` with `CoralTable` generated using `CoralCatalog`. that can give some weird bugs and else cases. 

we need to decouple the two code paths a bit more.
the user's entry point is the class - `ViewToAvroSchemaConverter` (is that right?) . if yes, we need to wire it up such that there are 2 independent code paths (pseudo code example) - 
(1) 
```
viewToAvroSchemaConverter1 = new ViewToAvroSchemaConverter(HiveMetastoreClient);
viewToAvroSchemaConverter1.getAvroSchema(db,tbl)

-- this uses HiveToRelConverter(HiveMetastoreClient) & RelToAvroConverter(HiveMetastoreClient)
```

and 

(2) 
```
viewToAvroSchemaConverter2 = new ViewToAvroSchemaConverter(CoralCatalog);
viewToAvroSchemaConverter2.getAvroSchema(db,tbl)

-- this uses HiveToRelConverter(CoralCatalog) & RelToAvroConverter(CoralCatalog)
```

please check if we're enabling a code path which inter mixes the two interactions.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
aastha25: public facing API should only take either `CoralCatalog` or `HiveMetastoreClient`. please refactor the code.

aastha25:
@yyy1000 thanks for the PR! keep the momentum going  :) 

feel free to introduce new class variants , if needed, to enable the decoupling of new & old codepaths. And btw, the build is also failing.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
yyy1000: Done in dfff1b6 — `create(CoralCatalog, HiveMetastoreClient)` removed; new `create(CoralCatalog)` exposed; `create(HiveMetastoreClient)` marked `@Deprecated`. The new factory wires the inner `HiveToRelConverter` and `RelToAvroSchemaConverter` against the same `CoralCatalog` end-to-end, so the converter no longer mixes abstractions internally.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
yyy1000: Done in dfff1b6 — the two paths are now fully decoupled. `SchemaRelShuttle` branches on a single non-null field (CoralCatalog xor HiveMetastoreClient) with no fallback. A `CoralCatalog`-backed instance never resolves through `HiveMetastoreClient` and vice versa; if the configured catalog can't find the table it throws instead of silently degrading.

`ViewToAvroSchemaConverter` was also refactored along the same lines: `create(CoralCatalog)` constructs `HiveToRelConverter(coralCatalog)` + `RelToAvroSchemaConverter(coralCatalog)` end-to-end, and `inferAvroSchema` splits into two branches that share no state (`inferAvroSchemaCoralCatalog` and `inferAvroSchemaHiveMetastore`). A user can no longer build a `CoralRelNode` through `HiveMetastoreClient` and resolve base tables through `CoralCatalog` on the same converter instance.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/RelToAvroSchemaConverter.java
yyy1000: Done in dfff1b6 — `(CoralCatalog, HiveMetastoreClient)` constructor removed; new `(CoralCatalog)` constructor added; `(HiveMetastoreClient)` marked `@Deprecated` with a Javadoc pointer to the new constructor.

yyy1000:
Pushed dfff1b6 with the decoupling refactor — replies posted on each of the three threads above.

Also folded the `coral-hive:spotlessJava` → `generateTestGrammarSource` Gradle ordering fix into this push, so CI should now go green modulo the unrelated `ParseTreeBuilderTest.testSql11KeywordAsExplicitAlias` flakes that I confirmed reproduce on master without any of my changes (likely fallout from #602).

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -128,6 +131,46 @@ static Schema getAvroSchemaForTable(@Nonnull final Table table, boolean strictMo
...
+   * @param coralTable the table to resolve, must not be null
+   * @param strictMode if true, throw when no partner Avro schema is available
+   * @return the resolved Avro schema
+   */
+  static Schema getAvroSchemaForTable(@Nonnull final CoralTable coralTable, boolean strictMode) {
+    Preconditions.checkNotNull(coralTable);
+
+    // Hive-backed CoralTable: delegate to the existing Hive-table API to preserve current behavior.
+    if (coralTable instanceof HiveTable) {
+      return getAvroSchemaForTable(((HiveTable) coralTable).getHiveTable(), strictMode);
```
aastha25, line 153: `getHiveTable()`  method has documentation:
```
/**
   * INTERNAL API
   * @deprecated This method is for internal use only and will be removed in a future release.
   * Do not depend on this API.
   *
   * @return Hive metastore Table object
   */
```

please do not leverage this API. if the implementation requires a conversion from `(HiveTable) CoralTable` -> Hive's table object, implement the converter. Feel free to do that in coral-common so that other usages of that API can migrate too

yyy1000: Agreed on dropping the deprecated `getHiveTable()`. Worth flagging two things before sketching a fix:

1. PR #604 isn't the first caller. `HiveCalciteTableAdapter.java:122` and `ToRelConverter.java:209` already use `getHiveTable()` today — both inside `coral-common` itself. Whatever we land here is the natural migration target for those too.
2. A real converter needs more of `HiveTable`'s state than the current public API exposes — `serdeInfo.serializationLib` (AvroSerDe escape), serde `parameters` (partner-Avro extraction), and the cols/partition-keys split (today collapsed into one `StructType` by `getSchema()`). Reconstructing a Hive `Table` from `name()` + `properties()` + `getSchema()` alone would silently break the Hive arm.

Two ways forward:

1. Expand `HiveTable`'s public API in a separate prep PR, then add a real `HiveCoralTableConverter` here.
2. Add `HiveCoralTableConverter` here as a centralized facade that internally uses `getHiveTable()` (with `@SuppressWarnings`), so the deprecated call lives in exactly one place and the existing in-tree usages migrate behind the same converter.

1 is the right end state, 2 is a pragmatic interim. Which would you prefer?

aastha25: can you think about more options from a design standpoint? how about we expand `hiveTable.properties()` method [here](https://github.com/linkedin/coral/blob/master/coral-common/src/main/java/com/linkedin/coral/common/catalog/HiveTable.java#L64) to also include the serde properties in addition to the table properties - essentially that would truly capture the hive table's properties. 
would that resolve the issue? what all aspects of 'hiveTable' are required across coral-common & coral-schema? we can limit scope to just coral-schema for now. what are the implications here? or should we implement a new getAvroSchemaForTable() without knowledge of the base hive table? is that possible?

please spend some time thinking about these options and more and share what you think should be adopted and why.

yyy1000: Spent some time on this. My pick: targeted expansion of `HiveTable`'s public API plus a shared private helper between the two `getAvroSchemaForTable` overloads.

Four new methods on `HiveTable`, scoped to what coral-schema's Hive path actually consumes:

```java
public Map<String, String> serdeProperties()      // sd.serdeInfo.parameters
public String serializationLib()                  // sd.serdeInfo.serializationLib
public List<FieldSchema> regularColumns()         // sd.cols
public List<FieldSchema> partitionColumns()       // partitionKeys
```

Then in `SchemaUtilities`, extract the existing `getAvroSchemaForTable(Table)` body into a private helper that takes those fields plus `tableProperties`, `fullyQualifiedName`, `strictMode`. Both overloads (`Table` and `CoralTable`) become thin wrappers that route through it.

Result: bytecode-equivalent Hive behavior (same helper, same inputs), no `getHiveTable()` in coral-schema. Scope stays at "what coral-schema needs from HiveTable" per your framing — the other two in-tree usages (`HiveCalciteTableAdapter`, `ToRelConverter`) can migrate to these accessors in a follow-up without further API additions.

What do you think?

aastha25: i looked at Apache-Iceberg's `Table` API and serde properties are not retrievable using a top class API, it i shoved under properties. We should also look to not expand the user facing APIs extensively. 

So maybe create a constant key for `SERDE_SERIALIZATION` and put the serde in it in the same properties key value map. But agree that this is going beyond the scope of coral-schema. 
So for now lets log an issue for this follow up item, link this thread in the issue as well and link the issue in code as a TODO.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
```diff
@@ -128,6 +131,46 @@ static Schema getAvroSchemaForTable(@Nonnull final Table table, boolean strictMo
...
+    }
+
+    Preconditions.checkState(coralTable.getSchema() instanceof StructType,
+        "CoralTable schema must be a StructType but was " + coralTable.getSchema().getClass().getSimpleName());
+    StructType coralSchema = (StructType) coralTable.getSchema();
+
+    String fullName = coralTable.name();
+    int lastDot = fullName.lastIndexOf('.');
+    String recordNamespace = lastDot >= 0 ? fullName.substring(0, lastDot) : "";
+    String recordName = lastDot >= 0 ? fullName.substring(lastDot + 1) : fullName;
```
aastha25: is this is how name & namespace is extracted for hive tables? shouldnt we prefer to extract is from the Avro schema partner, if that is present? lets keep parity with old behaviour

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
```diff
@@ -184,6 +223,46 @@ private Schema inferAvroSchema(String dbName, String tableOrViewName, boolean st
     Preconditions.checkNotNull(dbName);
     Preconditions.checkNotNull(tableOrViewName);
 
+    if (coralCatalog != null) {
+      return inferAvroSchemaCoralCatalog(dbName, tableOrViewName, strictMode, forceLowercase);
```
aastha25: in your regression testing, please make sure that you test for permutations of strictMode + forceLowercase as well.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
```diff
@@ -184,6 +223,46 @@ private Schema inferAvroSchema(String dbName, String tableOrViewName, boolean st
...
+    }
+    return inferAvroSchemaHiveMetastore(dbName, tableOrViewName, strictMode, forceLowercase);
+  }
+
+  /**
+   * CoralCatalog-backed resolution. Tables and views are both surfaced through
+   * {@link CoralCatalog#getTable}; views use the CoralCatalog-aware
+   * {@link HiveToRelConverter} and {@link RelToAvroSchemaConverter}.
+   */
+  private Schema inferAvroSchemaCoralCatalog(String dbName, String tableOrViewName, boolean strictMode,
```
aastha25: nit: `inferAvroSchemaUsingCoralCatalog`

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
yyy1000: Confirmed offline — deferring this. Filed #606 to track the migration; linked the discussion threads (and your `SERDE_SERIALIZATION` constant-key suggestion) there. Added a TODO at the call site in this PR pointing at the issue (b8e6c1e).

coral-schema/src/main/java/com/linkedin/coral/schema/avro/SchemaUtilities.java
yyy1000: Good catch — fixed in b8e6c1e. When partner Avro is present, `MergeCoralSchemaWithAvro.mergeTopLevelStruct` already takes name/namespace from the partner record (via `copyRecord`), so legacy parity is already there for that case. For the no-partner fallback, I now match `convertHiveSchemaToAvro`: `recordName` = the bare table name, `recordNamespace` = the fully qualified name (e.g. `default.my_table`). Updated the affected `SchemaUtilitiesTests` assertions too.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
yyy1000: Tracking on internal testing side where the regression harness lives. Will extend `CoralSparkBaselineAnalysisForTrino` to emit one `avro_schema` column per `(strictMode, forceLowercase)` permutation so the diff query covers all four.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverter.java
yyy1000: Renamed in b8e6c1e. Also renamed the sibling `inferAvroSchemaHiveMetastore` → `inferAvroSchemaUsingHiveMetastore` for symmetry.

aastha25 (approved):

PR #585, opened 2026-03-17
Upgrade Hive 1.2.2 → 2.3.9

YogeshKothari26:
### What changes are proposed in this pull request, and why are they necessary?

This PR upgrades Coral's Hive dependency from **1.2.2 to 2.3.9**, and bumps the version from **2.3.x to 2.4.x** to reflect the scope of this change.

Part 2 of 3: [#580 Gradle](https://github.com/linkedin/coral/pull/580) → **this PR** (Hive) → [#596 Java 17](https://github.com/linkedin/coral/pull/596). Stays on Java 8.

#### Dependencies

- Hive 1.2.2 → 2.3.9
- DataNucleus/Derby/javax.jdo test deps consolidated in root `build.gradle`
- Pentaho exclusion (not in Maven Central)
- Hive CBO disabled via `systemProperty` (CalcitePlanner incompatible with Calcite 1.21.0.265)
- Excluded problematic transitives from published POM to protect downstream consumers

#### Source fixes

- `MetastoreProvider`: `getProxy(conf)` → `getProxy(conf, true)` (single-arg overload removed in 2.3.9). The new `boolean allowEmbedded` parameter controls whether an in-process HMS is permitted when `hive.metastore.uris` is unset. Passing `true` preserves 1.2.2 behavior, which `coral-service` relies on for local/test flows.

#### coral-spark-catalog compatibility

After merging master (which added `coral-spark-catalog` via [#584](https://github.com/linkedin/coral/pull/584)), the new module's tests needed Hive 2.3.9 compatibility fixes:
- Added DataNucleus/Derby/hive-serde test deps (same as other modules)
- Excluded Jackson from `hive-metastore` and `hive-exec-core` (Hive 2.3.9 brings Jackson 2.6.x which conflicts with Spark 3.5's Jackson 2.15)
- Disabled Hive CBO (same Calcite conflict as other modules)
- **Test-only changes** — no impact on published artifacts

#### Version bump

- `version.properties`: 2.3.* → 2.4.* (minor bump for Hive dependency upgrade)

#### Tests disabled

- `testEnumUnionString` — Hive 2.3.9 `SemanticAnalyzer` throws `AssertionError` in `UnparseTranslator.addTranslation` during CREATE VIEW with UNION ALL between Avro enum and string columns

<details>
<summary><strong>HMS API Changelog: 1.2.2 → 2.3.9</strong></summary>

**Core read APIs — UNCHANGED:**

[...]

PR #605, opened 2026-04-28
Expand testEnumUnionString comment with upstream-bug context

YogeshKothari26:

Documents context for the disabled `testEnumUnionString` test (disabled in [#585](https://github.com/linkedin/coral/pull/585)):

- The failure is an upstream Hive 2.3.9 parser bug — `SemanticAnalyzer` throws `AssertionError` in `UnparseTranslator.addTranslation` during `CREATE VIEW` parse for a UNION ALL between an Avro enum column and a string column.
- Coral's `mergeUnionSchema` (added in [#282](https://github.com/linkedin/coral/pull/282)) is unaffected and still produces a correct STRING-typed schema for already-created views at translation time.
- The bug fires at CREATE VIEW parse time on Hive 2.3.9. The case stays unsupported in Coral CI/i-tests (which embed Hive 2.3.9).
- Test will be re-enabled once the upstream Hive parser issue is resolved.

Comment-only change.

`./gradlew :coral-schema:spotlessCheck` — PASS. No code change.

### DIFF
--- a/coral-schema/src/test/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverterTests.java
+++ b/coral-schema/src/test/java/com/linkedin/coral/schema/avro/ViewToAvroSchemaConverterTests.java
@@ -1024,8 +1024,15 @@ public class ViewToAvroSchemaConverterTests {
     Assert.assertEquals(actualSchema.toString(true), TestUtils.loadSchema("testEnumUnionEnum-expected.avsc"));
   }
 
-  // Disabled: Hive 2.3.9 SemanticAnalyzer throws AssertionError in UnparseTranslator.addTranslation
-  // during CREATE VIEW with UNION ALL between Avro enum and string columns (HIVE-specific bug)
+  // Disabled: upstream Hive 2.3.9 parser bug. SemanticAnalyzer throws AssertionError
+  // in UnparseTranslator.addTranslation during CREATE VIEW parse when a UNION ALL
+  // mixes an Avro enum column with a string column. Coral's mergeUnionSchema (added
+  // in #282) is unaffected — it still produces a correct STRING-typed schema for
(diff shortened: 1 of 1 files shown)


aastha25:
if its a parser bug, does is mean its a hive 2.3 client side bug? we have internal view creation pipelines which use spark (and spark is internally using hive 2.3 client libs). is that job impacted or not?

YogeshKothari26:
@aastha25 — verified Spark CREATE VIEW path is not impacted by this bug.

**Test:** Created a Hive-backed Avro table with an enum column, ran the equivalent `CREATE VIEW v AS SELECT enum_field FROM t UNION ALL SELECT 'literal_string' FROM t` on Spark 3.5 with `spark.sql.hive.metastore.version = 2.3.9.1`. View created successfully, no `AssertionError`.

**Why Spark is safe:** Spark parses CREATE VIEW DDL with its own `SqlBaseParser` and stores the view body verbatim via `HiveExternalCatalog.createTable` → `Hive.createTable` (metadata API only). Hive's `SemanticAnalyzer` / `UnparseTranslator` — where this bug lives — are never invoked in Spark's flow. `DESCRIBE EXTENDED` confirms this empirically: `View Text` and `View Original Text` are byte-identical, meaning Spark stored the SQL as-is without re-canonicalizing. If `UnparseTranslator` had run, `View Text` would be the re-serialized form.

The bug fires only when Hive's parser is directly invoked (e.g. Hive CLI, embedded Hive in Coral's `HiveTester`). Spark-based view creation pipelines remain unaffected.

YogeshKothari26:
Same test passes on Spark 3.1 (built-in Hive 2.3.8.29): `CREATE VIEW` succeeded, no `AssertionError`.

aastha25 (approved):

PR #607, opened 2026-05-13
Warn when HiveViewExpander's expanded body row type disagrees with caller's

simbadzina:
## Summary

`HiveViewExpander.expandView()` previously discarded its `rowType` argument and returned whatever the re-parsed view body produced. The [`ViewExpander` contract](https://calcite.apache.org/javadocAggregate/org/apache/calcite/plan/RelOptTable.ViewExpander.html) is that the returned `RelRoot`'s row type should match the caller-provided `rowType`; when those two disagree in field count or field order, downstream consumers that rely on positional access can silently read values from the wrong source column.

Earlier revisions of this PR fixed the symptom inside Coral by wrapping the returned `RelRoot` in a `LogicalProject` that re-aligned fields to the caller-provided row type by name. The view producers that triggered the original mismatch have since been corrected upstream, so Coral no longer needs to silently rewrite the expanded view.

## What this PR does

Adds a warn-only diagnostic for the `ViewExpander` contract violation:

- `HiveViewExpander.expandView()` invokes `warnIfRowTypeMisaligned(root, rowType)` and returns the original `RelRoot` unchanged.
- `warnIfRowTypeMisaligned` logs a single `WARN` when the expanded body's row type disagrees with the caller-provided row type. The check is recursive: it compares field count, then case-insensitive field names by order at every nesting level — STRUCT-within-STRUCT and ARRAY/MULTISET-of-STRUCT included — so a nested reorder one or more levels deep still surfaces. No-op fast path when the types already agree.
- A `TODO` marks the intent to harden this from a warning into an `IllegalStateException` once we are confident no live Hive view trips it. The current `WARN` is a transitional safeguard, not the long-term design.

```mermaid
flowchart LR
  E[warnIfRowTypeMisaligned<br/>expected rowType vs<br/>expanded body rowType] --> Q{Field counts match<br/>and names aligned by order<br/>at every nesting level?}
  Q -->|Yes| K[Return root unchanged]
  Q -->|No| W[LOG.warn expected vs actual<br/>return root unchanged]
```

## Why warn rather than auto-realign

The previous revision's `LogicalProject` rewrite was a correct response to the contract violation but was also a load-bearing rewrite of the user's query plan: it changed which physical columns produce each output position. With the upstream view producers corrected, that rewrite is no longer necessary, and an explicit warning is a less invasive signal — it surfaces the violation without silently mutating the plan. The path remains open to tighten this into a thrown exception (`TODO` in the source) once we have signal that no live view trips it.

## Test plan

- [x] Unit tests in `HiveViewExpanderTest` covering top-level cases: aligned (no warn), case-only differences (no warn), reordered fields (one warn, root unchanged), arity mismatch (one warn, root unchanged), and a contract check that the warning path leaves the `RelRoot` un-rewritten.
- [x] Unit tests for nested-row recursion: aligned `STRUCT<...>` (no warn), reordered fields inside a top-level `STRUCT`, reordered fields inside `ARRAY<STRUCT<...>>`, and reordered fields in a three-level `STRUCT<STRUCT<STRUCT<...>>>` nesting.
- [x] Tests inject a Mockito-mocked `org.slf4j.Logger` via reflection in `@BeforeMethod` and restore the original in `@AfterMethod`, so they assert warn emission without depending on the test classpath's slf4j binding.
- [x] Sanity-check, top-level: removing the `LOG.warn` line inside `warnIfRowTypeMisaligned` causes exactly the two top-level warn-asserting tests to fail; restoring it turns the suite green.
- [x] Sanity-check, recursion: stubbing `nestedRowTypesAlignedByOrder` to `return true` causes exactly the three nested-aware warn-asserting tests to fail; restoring it turns the suite green.
- [x] Full `:coral-hive:test` and `:coral-trino:test` suites pass.

### DIFF
--- a/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveViewExpander.java
+++ b/coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveViewExpander.java
@@ -1,30 +1,44 @@
 /**
- * Copyright 2017-2022 LinkedIn Corporation. All rights reserved.
+ * Copyright 2017-2026 LinkedIn Corporation. All rights reserved.
  * Licensed under the BSD-2 Clause license.
  * See LICENSE in the project root for license information.
  */
 package com.linkedin.coral.hive.hive2rel;
 
+import java.util.ArrayList;
 import java.util.List;
 
 import javax.annotation.Nonnull;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 
 import org.apache.calcite.plan.RelOptTable;
+import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.RelRoot;
+import org.apache.calcite.rel.logical.LogicalProject;
 import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.util.Util;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.linkedin.coral.common.FuzzyUnionSqlRewriter;
 
 
 /**
  * Class that implements {@link org.apache.calcite.plan.RelOptTable.ViewExpander}
- * interface to support expansion of Hive Views to relational algebra.
+ * interface to support expansion of Hive Views to relational algebra. The
+ * returned {@link RelRoot} is realigned to the caller-provided
+ * {@link RelDataType} by name (case-insensitive) to preserve the
+ * {@link RelOptTable.ViewExpander} contract.
  */
 public class HiveViewExpander implements RelOptTable.ViewExpander {
+
+  private static final Logger LOG = LoggerFactory.getLogger(HiveViewExpander.class);
 
   private final HiveToRelConverter hiveToRelConverter;
   /**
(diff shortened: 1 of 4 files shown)


coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveViewExpander.java
```diff
@@ -46,6 +60,58 @@ public RelRoot expandView(RelDataType rowType, String queryString, List<String>
...
+    RelRoot root = hiveToRelConverter.getSqlToRelConverter().convertQuery(sqlNode, true, true);
+    return alignToRowType(root, rowType);
+  }
+
+  /**
+   * Wrap {@code root} in a {@link LogicalProject} that re-orders its output
+   * fields to match {@code expected} by name (case-insensitive). No-op when the
+   * orderings already agree. If a name is missing from {@code root} or arities
+   * differ, returns {@code root} unchanged and logs a warning.
+   */
```
wmoustafa, line 72: Is not the right behavior to throw an exception? There should not be a disagreement between the caller expectation and the view structure in the first place.

wmoustafa (dismissed):

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveViewExpander.java
simbadzina: Do you mean for when there is a mismatch like an missing field, or even when there is an ordering issue?

For the ordering, I think we can self-heal drift with the alignment code since there is a complete info.

I'm assuming a change like a new column could show up a mismatch, missing field.
Automatically taking the root would expose the column to consumers, whereas an error would break the view. I think for `select *` we'd want the new column to seamless appear.

simbadzina: The source of the divergence I observed was in another system, outside of Coral. During view creation, the storage columns were being explicitly set in the storageDescriptor of the view in a different order than in the view text. This particular instance of divergence has been fixed in that system.

To guard against similar broken views Coral can throw an exception as you suggested, or at the least log a warning.

coral-hive/src/main/java/com/linkedin/coral/hive/hive2rel/HiveViewExpander.java
```diff
@@ -46,6 +55,68 @@ public RelRoot expandView(RelDataType rowType, String queryString, List<String>
...
+    }
+    final RelDataType ac = a.getComponentType();
+    final RelDataType bc = b.getComponentType();
+    if ((ac == null) != (bc == null)) {
+      return false;
+    }
+    if (ac != null) {
+      return nestedRowTypesAlignedByOrder(ac, bc);
+    }
+    return true;
```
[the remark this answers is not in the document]
simbadzina: Can you create a TODO comment for this.

PR #610, opened 2026-06-15
[Coral-Schema] Reconstruct multi-branch unions from Iceberg union-structs

yyy1000:
## What
Adds a multi-branch union reconstruction path to `MergeCoralSchemaWithAvro` (the Iceberg-first Avro merge engine introduced in #600/#604).

## Why
Iceberg has no union type. A Hive `uniontype<A,B,C>` that has been persisted into an Iceberg table surfaces as the struct `{tag:INT, field0:A, field1:B, field2:C}` — the Trino union representation ([trinodb/trino#3483](https://github.com/trinodb/trino/pull/3483)), the same encoding `HiveToCoralTypeConverter.convertUnion` produces — while the partner Avro still describes the column as a real union `[null,A,B,C]`.

Before this change, the engine could not match the union partner, so it synthesized a record: multi-branch unions collapsed, and union-typed field defaults failed outright. This diverged from the legacy Hive Avro path (`MergeHiveSchemaWithAvro.union`).

## How
- **`unionPartnerOrNull`** detects the case, requiring all of: (1) the Coral struct is union-shaped (`tag:INT` then `field0..fieldN-1` in order), (2) the partner is an Avro union, and (3) the partner's non-null branch count equals the member count. A `memberCount >= 2` guard makes detection collision-free — a genuine nullable struct yields `[null, record]` (one non-null branch), which can never equal a member count of two or more. Single-member union-structs fall back to the struct path.
- **`mergeUnionStruct`** merges each `fieldN` member against the partner union branch by ordinal, strips per-member null wrappers (the union's own NULL branch carries nullability), and emits the NULL branch first when the partner union has one — matching `MergeHiveSchemaWithAvro.union`. Emitting a real union also lets the partner's field default apply exactly as on the Hive path.

## Testing
Six new tests (full `coral-schema` suite green, spotless clean): multi-branch reconstruction, the no-null-branch case, arrays of unions, the union-default regression, and a negative case proving a genuine record partner is not mistaken for a union.

### DIFF
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
@@ -20,6 +20,7 @@ import org.apache.avro.Schema;
 import com.linkedin.coral.common.types.ArrayType;
 import com.linkedin.coral.common.types.BinaryType;
 import com.linkedin.coral.common.types.CoralDataType;
+import com.linkedin.coral.common.types.CoralTypeKind;
 import com.linkedin.coral.common.types.DecimalType;
 import com.linkedin.coral.common.types.MapType;
 import com.linkedin.coral.common.types.StructField;
@@ -88,7 +89,12 @@ class MergeCoralSchemaWithAvro {
   private Schema mergeType(CoralDataType coralType, @Nullable Schema partner) {
     switch (coralType.getKind()) {
       case STRUCT:
-        return mergeStruct((StructType) coralType, partner);
+        StructType structType = (StructType) coralType;
+        Schema unionPartner = unionPartnerOrNull(structType, partner);
+        if (unionPartner != null) {
+          return mergeUnionStruct(structType, unionPartner);
+        }
+        return mergeStruct(structType, partner);
       case ARRAY:
         return mergeArray((ArrayType) coralType, partner);
       case MAP:
@@ -117,6 +123,81 @@ class MergeCoralSchemaWithAvro {
       result.setFields(fields);
     }
     return applyCoralNullability(result, structType.isNullable(), partner);
+  }
+
+  /**
+   * Decides whether {@code structType} is a Trino-style union-struct whose partner Avro is a genuine
+   * multi-branch union, and if so returns that partner union (null branch intact). Returns null to fall
+   * back to regular struct handling.
+   *
+   * <p>Iceberg's type system has no union type, so a Hive {@code uniontype<A,B,C>} that has been persisted
+   * into Iceberg surfaces as the struct {@code {tag:INT, field0:A, field1:B, field2:C}} — the same encoding
+   * {@link com.linkedin.coral.common.HiveToCoralTypeConverter#convertUnion} produces (Trino union
+   * representation, see <a href="https://github.com/trinodb/trino/pull/3483">trinodb/trino#3483</a>). To
+   * stay faithful to the legacy Hive Avro path ({@link MergeHiveSchemaWithAvro#union}), such a column must be
+   * emitted as an Avro union rather than a record.
+   *
+   * <p>The struct shape alone is ambiguous — a genuine record could be named {@code {tag, field0, ...}} — so
+   * two further conditions are required: the partner must be an Avro union, and its non-null branch count
+   * must equal the number of {@code fieldN} members. A genuine nullable struct yields {@code [null, record]}
+   * (one non-null branch), which fails the count check for any multi-branch union. The only residual
+   * ambiguity is a single-member {@code uniontype<record>}, which is vanishingly rare; it is treated as a
+   * struct (the pre-existing behavior).
+   */
+  @Nullable
+  private Schema unionPartnerOrNull(StructType structType, @Nullable Schema partner) {
+    if (partner == null || partner.getType() != Schema.Type.UNION || !isUnionStruct(structType)) {
+      return null;
+    }
+    int memberCount = structType.getFields().size() - 1; // exclude the leading "tag" field
+    int nonNullBranchCount = SchemaUtilities.discardNullFromUnionIfExist(partner).getTypes().size();
+    // Require at least two members so the count check is collision-free: a genuine nullable struct yields
+    // [null, record] (one non-null branch), which can never equal a member count of two or more. A
+    // single-member union-struct is left to the struct path (see Javadoc).
+    return memberCount >= 2 && memberCount == nonNullBranchCount ? partner : null;
+  }
+
+  /**
+   * Recognizes the union-struct shape: a leading {@code tag} field of type INT followed by
+   * {@code field0, field1, ..., fieldN-1} in order. See {@link #unionPartnerOrNull}.
+   */
+  private boolean isUnionStruct(StructType structType) {
+    List<StructField> fields = structType.getFields();
+    if (fields.size() < 2) {
+      return false;
+    }
+    StructField tag = fields.get(0);
+    if (!"tag".equals(tag.getName()) || tag.getType().getKind() != CoralTypeKind.INT) {
+      return false;
+    }
+    for (int i = 1; i < fields.size(); i++) {
+      if (!("field" + (i - 1)).equals(fields.get(i).getName())) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  /**
+   * Reconstructs an Avro union from a union-struct, merging each {@code fieldN} member against the
+   * corresponding partner union branch (by ordinal). Null placement follows {@link MergeHiveSchemaWithAvro#union}:
+   * the NULL branch is emitted first when the partner union carries one. Caller ({@link #unionPartnerOrNull})
+   * guarantees the member count matches the partner's non-null branch count.
+   */
+  private Schema mergeUnionStruct(StructType unionStruct, Schema partnerUnion) {
+    List<Schema> partnerBranches = SchemaUtilities.discardNullFromUnionIfExist(partnerUnion).getTypes();
+    List<StructField> members = unionStruct.getFields(); // [tag, field0, field1, ...]
+    List<Schema> unionTypes = new ArrayList<>();
+    if (SchemaUtilities.nullExistInUnion(partnerUnion)) {
+      unionTypes.add(Schema.create(Schema.Type.NULL));
+    }
+    for (int i = 1; i < members.size(); i++) {
+      Schema branch = mergeType(members.get(i).getType(), partnerBranches.get(i - 1));
+      // A union member is never itself nullable in Avro — the union's own NULL branch carries nullability,
+      // so strip any option wrapper the recursive merge may have added around the member.
+      unionTypes.add(SchemaUtilities.extractIfOption(branch));
+    }
+    return Schema.createUnion(unionTypes);
   }
 
   private Schema mergeArray(ArrayType arrayType, @Nullable Schema partner) {
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
@@ -457,20 +457,42 @@ class MergeCoralSchemaWithAvro {
   }
 
   /**
-   * Find a partner field by case-insensitive name match, returning the first match. Matches the
-   * behavior of {@link MergeHiveSchemaWithAvro}'s partner accessor. Iceberg's schema spec disallows
-   * sibling fields that differ only in case, so this is unambiguous for Iceberg-sourced Coral schemas.
+   * Resolves the partner field for a Coral field name, following the documented matching rules:
+   *
+   * <ol>
+   *   <li>an exact, case-sensitive match wins outright;</li>
+   *   <li>otherwise a <em>unique</em> case-insensitive match;</li>
+   *   <li>otherwise no match.</li>
+   * </ol>
+   *
+   * <p>Iceberg forbids sibling fields differing only in case, so the Coral side is always unambiguous —
+   * but the <em>partner</em> is Avro, which permits both {@code fA} and {@code fa} in one record. A
+   * first-match-wins scan could therefore attach the wrong field's doc, default, props or enum/fixed
+   * promotion even when an exact match exists. Preferring the exact match removes that; refusing to
+   * guess between two inexact candidates keeps a rare, genuinely ambiguous schema from silently
+   * inheriting arbitrary metadata.
+   *
+   * <p>The output field name is unaffected — it is always the Coral name (see {@link #mergeField}). Only
+   * the source of the partner's metadata is decided here.
    */
   @Nullable
   private Schema.Field findPartnerField(@Nullable Schema partnerRecord, String coralFieldName) {
     if (partnerRecord == null) {
       return null;
     }
+    Schema.Field exactMatch = partnerRecord.getField(coralFieldName);
+    if (exactMatch != null) {
+      return exactMatch;
+    }
+    Schema.Field caseInsensitiveMatch = null;
     for (Schema.Field field : partnerRecord.getFields()) {
       if (field.name().equalsIgnoreCase(coralFieldName)) {
-        return field;
+        if (caseInsensitiveMatch != null) {
+          return null; // ambiguous: two partner fields differ only in case and neither matches exactly
+        }
+        caseInsensitiveMatch = field;
       }
     }
-    return null;
+    return caseInsensitiveMatch;
   }
 }
--- a/coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
+++ b/coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
@@ -34,7 +34,12 @@ import static com.linkedin.coral.schema.avro.AvroSerdeUtils.*;
  * Merges a Coral schema (from Iceberg) with a partner Avro schema, using Iceberg-first semantics:
  *
  * <ul>
- *   <li>CoralDataType is the source of truth for field existence, name, nullability, and type.</li>
+ *   <li>CoralDataType is the source of truth for field existence, name, and type.</li>
+ *   <li><b>Nullability is always relaxed:</b> where CoralDataType and the partner disagree, the output
+ *       is nullable. The merge never narrows. Iceberg's {@code required} flag is frequently inherited
+ *       from a Hive migration rather than from an enforced constraint, so treating it as authoritative
+ *       would drop a {@code null} branch the partner declared and produce a schema stricter than the
+ *       data. This supersedes the earlier Iceberg-first nullability rule.</li>
  *   <li>Partner Avro contributes whatever metadata it carries for a matched field — defaults, docs,
  *       field props, aliases, union envelope shape and null placement, enum/fixed/uuid materialization.
  *       Nothing is fabricated: attributes the partner does not carry are absent from the output. When
@@ -121,7 +126,7 @@ class MergeCoralSchemaWithAvro {
       result = Schema.createRecord("record" + recordNum, null, "namespace" + recordNum, false);
       result.setFields(fields);
     }
-    return applyCoralNullability(result, structType.isNullable(), partner);
+    return applyRelaxedNullability(result, structType.isNullable(), partner);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
```diff
@@ -119,6 +125,81 @@ private Schema mergeStruct(StructType structType, @Nullable Schema partner) {
...
+   *
+   * <p>The struct shape alone is ambiguous — a genuine record could be named {@code {tag, field0, ...}} — so
+   * two further conditions are required: the partner must be an Avro union, and its non-null branch count
+   * must equal the number of {@code fieldN} members. A genuine nullable struct yields {@code [null, record]}
+   * (one non-null branch), which fails the count check for any multi-branch union. The only residual
+   * ambiguity is a single-member {@code uniontype<record>}, which is vanishingly rare; it is treated as a
+   * struct (the pre-existing behavior).
+   */
+  @Nullable
+  private Schema unionPartnerOrNull(StructType structType, @Nullable Schema partner) {
```
simbadzina: **Suggestion: return a `boolean` and rename to `isMultiBranchUnionStruct`**

This method only ever returns `partner` unchanged or `null` — it never constructs or transforms the schema it hands back. That makes the `Schema`-or-null return effectively a boolean in disguise, and the caller has to re-derive the meaning from the null check:

```java
Schema unionPartner = unionPartnerOrNull(structType, partner);
if (unionPartner != null) {
  return mergeUnionStruct(structType, unionPartner);
}
```

I'd return a `boolean` named for what's being asserted, and let the caller use `partner` directly:

```java
if (isMultiBranchUnionStruct(structType, partner)) {
  return mergeUnionStruct(structType, partner);
}
return mergeStruct(structType, partner);
```

```java
private boolean isMultiBranchUnionStruct(StructType structType, @Nullable Schema partner) {
  if (partner == null || partner.getType() != Schema.Type.UNION || !isUnionStruct(structType)) {
    return false;
  }
  int memberCount = structType.getFields().size() - 1; // exclude the leading "tag" field
  int nonNullBranchCount = SchemaUtilities.discardNullFromUnionIfExist(partner).getTypes().size();
  return memberCount >= 2 && memberCount == nonNullBranchCount;
}
```

A predicate name documents the decision at the call site, drops the `@Nullable` return, and removes the `unionPartner` local that's just an alias for an input. (The `@link` references in `isUnionStruct` and `mergeUnionStruct` would need updating to the new name.)

coral-schema/src/test/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvroTests.java
```diff
@@ -327,8 +327,59 @@ public void shouldHandleNestedStructsWithPartner() {
...
+    Schema u = result.getField("u").schema();
+    assertEquals(u.getType(), Schema.Type.UNION);
+    assertEquals(u.getTypes().get(0).getType(), Schema.Type.NULL);
+    assertEquals(u.getTypes().get(1).getType(), Schema.Type.INT);
+    assertEquals(u.getTypes().get(2).getType(), Schema.Type.STRING);
+    assertEquals(u.getTypes().get(3).getType(), Schema.Type.BOOLEAN);
+  }
+
+  @Test
+  public void shouldReconstructUnionWithoutNullBranch() {
```
simbadzina, line 360: **Suggestion: add coverage for non-primitive union members**

Both union tests here use primitive members (`int`/`string`/`boolean`). The reconstruction path does more than copy primitives across — each member is recursively merged against its partner branch and then unwrapped via `extractIfOption`, so I'd add cases where a `fieldN` member is itself:

- a **record** (`uniontype<int, struct<x:int>>`) — verify the struct member becomes a record branch with its fields merged from the partner, and is *not* double-wrapped in its own `[null, record]` option (the union's own NULL branch already carries nullability)
- an **array** (`uniontype<int, array<string>>`) — verify the member stays an array branch rather than being collapsed
- a member whose partner branch is an **enum** (`uniontype<string, int>` with partner `[null, enum, int]`) — verify the STRING member promotes to the enum, confirming branches go through the normal promotion path rather than emitting a bare string

These are the cases most likely to regress if the per-branch merge or the `extractIfOption` unwrap is ever changed, and they aren't exercised today.

yyy1000: done

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
yyy1000: done

simbadzina (dismissed):
LGTM.

Beyond the tests in the PR, Claude did some additional corner-case checking locally:

- Probed ~25 edge cases around the union-struct path: unions nested in arrays/maps and inside regular structs, non-contiguous `fieldN`, case variations on `tag`/`fieldN`, member/branch count mismatches, null placement in the partner union, and partner branch order divergence. The false-positive guards hold up — a genuine nullable struct that happens to reuse the `tag`/`field0` names still comes out a record.
- Ran each of those against the legacy `MergeHiveSchemaWithAvro` path for comparison. No regressions: everywhere the new code throws, legacy throws the same way, and on arity drift the new code actually degrades more gracefully than legacy (which throws `IndexOutOfBoundsException`).
- Ran the full multi-module build locally against d07c8ecb — all modules green, coral-schema 152 tests passing.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
```diff
@@ -88,7 +89,11 @@ private Schema mergeTopLevelStruct(StructType structType, @Nullable Schema partn
   private Schema mergeType(CoralDataType coralType, @Nullable Schema partner) {
     switch (coralType.getKind()) {
       case STRUCT:
-        return mergeStruct((StructType) coralType, partner);
+        StructType structType = (StructType) coralType;
+        if (isMultiBranchUnionStruct(structType, partner)) {
+          return mergeUnionStruct(structType, partner);
+        }
+        return mergeStruct(structType, partner);
```
aastha25: where are we taking care of nullable single uniontype ? there are 4 cases in total - 
(1) nullable complex union - 
(2) non-nullable complex union
(3) nullable single union
(4) non-nullable single union

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
```diff
@@ -119,6 +124,80 @@ private Schema mergeStruct(StructType structType, @Nullable Schema partner) {
...
+    }
+    StructField tag = fields.get(0);
+    if (!"tag".equals(tag.getName()) || tag.getType().getKind() != CoralTypeKind.INT) {
+      return false;
+    }
+    for (int i = 1; i < fields.size(); i++) {
+      if (!("field" + (i - 1)).equals(fields.get(i).getName())) {
+        return false;
+      }
+    }
```
aastha25, line 239: is the login assuming that we dont see a nullable compelx union -  {null, tag, field0, field1...} or is it expected that `structType` will not be nullable? is the latter, we should document it in the method java docs.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
```diff
@@ -119,6 +124,80 @@ private Schema mergeStruct(StructType structType, @Nullable Schema partner) {
...
+    List<StructField> fields = structType.getFields();
+    if (fields.size() < 2) {
+      return false;
+    }
+    StructField tag = fields.get(0);
+    if (!"tag".equals(tag.getName()) || tag.getType().getKind() != CoralTypeKind.INT) {
+      return false;
+    }
+    for (int i = 1; i < fields.size(); i++) {
+      if (!("field" + (i - 1)).equals(fields.get(i).getName())) {
```
aastha25: lets do case insensitive checks

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
```diff
@@ -116,7 +126,124 @@ private Schema mergeStruct(StructType structType, @Nullable Schema partner) {
...
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Recognizes the union-struct shape: a leading {@code tag} field of type INT followed by
+   * {@code field0, field1, ..., fieldN-1} in order. See {@link #isReconstructableUnionStruct}.
+   */
+  private boolean isUnionStruct(StructType structType) {
```
aastha25, line 224: this only detects complex union. 
Imagine a single union of struct , partner schema (avro type) is [struct] and Coral schema (StructType) - struct. 
then checking member of this struct is not relevant and needs special handling

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
```diff
@@ -116,7 +126,124 @@ private Schema mergeStruct(StructType structType, @Nullable Schema partner) {
...
+    }
+    int memberCount = structType.getFields().size() - 1; // exclude the leading "tag" field
+    List<Schema> nonNullBranches = SchemaUtilities.discardNullFromUnionIfExist(partner).getTypes();
+    if (memberCount != nonNullBranches.size()) {
+      return false;
+    }
+    if (memberCount >= 2) {
+      return true;
+    }
+    return !describesStructItself(nonNullBranches.get(0));
```
aastha25, line 199: i do not follow this check. At this point in the code, we have established that partner avro schema is (1) valid (2) it is of type union `uniontype<..>` and (3) the coral struct type is a struct with tag and at least 2 members - `{tag, f0, f1}`. 
why are we checking if the `f0` is a record or a uniontype itself?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
```diff
@@ -116,7 +126,124 @@ private Schema mergeStruct(StructType structType, @Nullable Schema partner) {
...
+   * {@code HiveTable} to the Hive path), and {@code IcebergToCoralTypeConverter} passes Iceberg's
+   * {@code isOptional()} straight through, so the flag reflects the Iceberg column rather than a
+   * constant. The NULL branch is emitted first, matching {@link MergeHiveSchemaWithAvro#union}.
+   *
+   * <p>Member nullability is deliberately <em>not</em> relaxed into the envelope. Avro forbids a union
+   * directly containing a union, and in the union-struct encoding only one member is live at a time, so
+   * members are optional regardless of whether the union itself is nullable. Relaxing from them would
+   * make every union nullable; each member therefore keeps its option wrapper stripped.
+   */
+  private Schema mergeUnionStruct(StructType unionStruct, Schema partnerUnion) {
```
aastha25, line 260: would prefer to call this method `mergeComplexUnionStruct` just to be clear that this only handles compelx unions.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
```diff
@@ -88,7 +94,11 @@ private Schema mergeTopLevelStruct(StructType structType, @Nullable Schema partn
   private Schema mergeType(CoralDataType coralType, @Nullable Schema partner) {
     switch (coralType.getKind()) {
       case STRUCT:
-        return mergeStruct((StructType) coralType, partner);
+        StructType structType = (StructType) coralType;
+        if (isReconstructableUnionStruct(structType, partner)) {
+          return mergeUnionStruct(structType, partner);
+        }
+        return mergeStruct(structType, partner);
```
aastha25: please run a single union of struct example through this case. 
coralType = struct = {} , 
corresponding partner schema = 
partner = uniontype<struct>

coral-schema/src/test/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvroTests.java
```diff
@@ -327,8 +338,414 @@ public void shouldHandleNestedStructsWithPartner() {
...
+    assertEquals(enumBranch.getType(), Schema.Type.ENUM);
+    assertEquals(enumBranch.getName(), "Color");
+    assertEquals(enumBranch.getEnumSymbols(), Arrays.asList("RED", "GREEN"));
+    assertEquals(u.getTypes().get(2).getType(), Schema.Type.INT);
+  }
+
+  @Test
+  public void shouldReconstructNullableSingleUnion() {
+    // uniontype<string> persisted into Iceberg is {tag, field0} — HiveToCoralTypeConverter.convertUnion
+    // emits the union-struct encoding for EVERY arity, not just multi-branch. A nullable single union must
```
aastha25: is this assumption right? isnt this peristsed into iceberg as type `string` ?

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
yyy1000: You're right, and it was a real gap — thanks for catching it. `isUnionStruct` only recognises the `{tag, field0, …}` encoding, so the case you describe — Iceberg flattened the uniontype away, Coral sees a plain `struct`, partner still says `[struct]` — never reached it and the envelope was dropped.

Production confirmed it independently: `organization_mp.product_dbchanges_hourly` `$.value.productCategoryUrns` is `uniontype<array<string>>` stored by Iceberg as a bare `array<string>`, with the partner still declaring `[{"type":"array","items":"string"}]`. The parity harness had been reporting it as a regression.

Fixed in aed4e5f0: the unwrap/rewrap now sits in `mergeType` above the kind switch, so it covers struct, array and map — not just leaves, where it previously lived. Union-struct reconstruction is checked first and keeps precedence, since it handles the opposite situation (Iceberg *preserved* the encoding).

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
yyy1000: The premise is slightly off, which tells me the code needs to say this out loud: this line is only reachable when `memberCount == 1`, because the check above short-circuits:

```java
if (memberCount >= 2) {
  return true;
}
return !describesStructItself(nonNullBranches.get(0));
```

For two or more members the count is conclusive, exactly as you say. Arity 1 is the ambiguous one: both `uniontype<X>` and a genuine nullable struct named `{tag, field0}` present a single non-null branch, so the count proves nothing. The partner's sole branch breaks the tie — for a genuine struct it is the record describing that struct and so carries its own `tag` field, whereas for `uniontype<X>` it is the member type `X`.

We can't simply treat arity 1 as a struct: `HiveToCoralTypeConverter.convertUnion` emits `{tag, field0}` for *every* arity, so that would leak the internal encoding into the output — which is what 80c14fde fixed.

I'll add a comment at the call site making the "only reachable at arity 1" precondition explicit; it shouldn't take reading two methods to see it.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
yyy1000: Gentle pushback: it handles single-member unions too, so `mergeComplexUnionStruct` would be inaccurate. `isReconstructableUnionStruct` admits arity 1 whenever the partner's sole branch isn't the struct's own record, and `shouldReconstructNullableSingleUnion` / `shouldReconstructNonNullableSingleUnion` cover it — that was the point of 80c14fde.

The naming pain is real though: "union-struct" is doing double duty for the Trino encoding and for the merge direction. Would `mergeUnionFromUnionStruct` work for you? It says what it produces and from what, without implying arity. Happy to take your preference.

coral-schema/src/main/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvro.java
yyy1000: Ran it. Before aed4e5f0 it was broken. Here's the trace now, with `coralType = struct{x}` (Iceberg flattened) and `partner = ["Inner"]`:

1. `mergeType` — `isReconstructableUnionStruct` is false (the Coral type is a plain struct, not `{tag, field0}`), so union-struct reconstruction correctly declines.
2. `nullFreeSingleBranchOrNull(["Inner"])` returns `Inner`.
3. `mergeByKind` → `mergeStruct`, merging against the partner record, so the name `Inner` plus its docs, props and defaults all survive.
4. The result is a RECORD, so the envelope is restored: `["Inner"]`.

If the Coral struct is nullable, step 3 yields `["null", Inner]` — already a union, so no second envelope is added, since `["null", X]` is simultaneously the plain nullable form and the only Avro representation of a nullable single union.

Previously step 2 didn't exist: `mergeStruct` called `extractIfOption(["Inner"])`, which unwraps only the two-branch `["null", X]`, so the one-branch union came back untouched, `extractPartnerRecord` returned null, and the output was a generated record with **no partner metadata and no envelope**. Now covered by `shouldPreserveSingleElementUnionEnvelopeAroundAStructAndStillMergeIt`, which asserts both the envelope and that the partner's name and docs survive.

coral-schema/src/test/java/com/linkedin/coral/schema/avro/MergeCoralSchemaWithAvroTests.java
yyy1000: Good question, and you're right that flattening is what happens in practice.

Both shapes occur. The one this test covers — `uniontype<string>` surfacing as `{tag, field0}` — is the Trino union representation, which we *do* observe for multi-branch unions in production (the four `facetClauses` `array<union>` tables). For arity 1 I have no production example of the encoding being preserved, so this test is defensive rather than evidence-based, and the comment should say so.

The flattened shape you're describing is real and was being mishandled: `organization_mp.product_dbchanges_hourly` `$.value.productCategoryUrns` is `uniontype<array<string>>` stored as a bare `array<string>` with partner `[{"type":"array","items":"string"}]`. That's fixed in aed4e5f0 — the same gap you flagged in your other comment.

So both directions are handled now: encoding preserved → reconstruct the union; encoding flattened → merge against the sole branch and restore the envelope. I'll relabel this test's comment to make clear it covers the preserved-encoding case and isn't asserting that Iceberg always preserves it.

aastha25:
reviewer notes for single element union is handled in code. 
preferred path: single union in avro has been extracted as underlying type in input iceberg (and hence the Coral IR). Hence, merge happens between `'X'` & `['X']`
defensive fallback path: single union in avro has been converted to `struct{tag:int, field0:X}`. Hence, merge happens between `struct{tag:int, field0:X}`. & `['X']`

```
┌─────────────────────────────┬─────────────────────────────────────────────────────────────────────────────────┬────────────────────────────────────────────────────────────┬───────────────────────────────────────────────────────────────┐
│ Scenario                    │ Coral type from Iceberg → Avro representation                                   │ Handled by                                                 │ Test                                                          │
├─────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────┤
│ Flattened — the common case │ Coral: X (string, …) Avro partner: ["X"] Avro output: ["X"] — or            │ mergeType → nullFreeSingleBranchOrNull → merge branch →    │ shouldPreserveSingleElementUnionEnvelopeWhenIcebergFlattensIt │
│                             │ ["null","X"] when optional                                                      │ restore envelope                                           │                                                               │
├─────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────┤
│ Struct survived — rare /    │ Coral: struct{tag:int, field0:X} Avro partner: ["null","X"]Avro output:          │ isReconstructableUnionStruct (memberCount==1 →             │ shouldReconstructNullableSingleUnion                          │
│ defensive                   │ ["null","X"]                                                                    │ describesStructItself)                                     │                                                               │
└─────────────────────────────┴─────────────────────────────────────────────────────────────────────────────────┴────────────────────────────────────────────────────────────┴───────────────────────────────────────────────────────────────┘

aastha25 (approved):

